PR #9, opened 2020-09-22
modules: make image configurable

hunchback:


roee88:
Since modules are installed by the control plane, this is meaningless. At least until we add the capability to override default values using the module CRD, and this is undecided yet. 

Instead, the image should be fixed and set to the real mover image and an issue should be opened for "Enable setting values for modules at registration time".

roee88:
Leaving this open to change the value of the mover image to be fixed to the latest mover image (not dummy)

roee88:
Fixed images handled by  https://github.com/IBM/the-mesh-for-data/pull/18

hunchback:
@roee88 - why was this closed ? this enables to use helm charts residing on a designated registry (rather than the default one)

roee88:
Please see https://github.com/IBM/the-mesh-for-data/issues/12 and target it in a separate PR.

PR #10, opened 2020-09-22
build: Build movement-controller image in CI

froesef:
This PR will always build the movement-controller image that is used in some demos.

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -10,9 +10,7 @@ IMG_MC ?= ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/movement-controller:${DOCKER_TA
 IMG_DM ?= ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/dummy-mover:${DOCKER_TAGNAME}
 
 .PHONY: all
-all: docker-movement-controller docker-build-dummy-mover
-	$(MAKE) docker-push IMG=$(IMG_MC)
-	$(MAKE) docker-push IMG=$(IMG_DM)
+all: docker-build-all docker-push-all
 
 .PHONY: docker-build-all
 docker-build-all: docker-movement-controller docker-build-dummy-mover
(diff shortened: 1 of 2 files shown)


build/Makefile
```diff
@@ -10,7 +10,15 @@ IMG_MC ?= ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/movement-controller:${DOCKER_TA
 IMG_DM ?= ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/dummy-mover:${DOCKER_TAGNAME}
 
 .PHONY: all
-all: docker-movement-controller docker-dummy-mover
+all: docker-movement-controller docker-build-dummy-mover
```
hunchback, line 13: why not just append `docker-push-all` at the end of this line?

build/Makefile
```diff
@@ -10,7 +10,15 @@ IMG_MC ?= ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/movement-controller:${DOCKER_TA
...
+	$(MAKE) docker-push IMG=$(IMG_MC)
+	$(MAKE) docker-push IMG=$(IMG_DM)
```
hunchback, line 15: As an alternative implement `all` as:
```
all: docker-build-all docker-push-all
```

build/Makefile
```diff
@@ -10,7 +10,15 @@ IMG_MC ?= ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/movement-controller:${DOCKER_TA
...
+
+.PHONY: docker-build-all
+docker-build-all: docker-movement-controller docker-build-dummy-mover
```
hunchback, line 18: renaming to `docker-build` may be better

build/Makefile
```diff
@@ -10,7 +10,15 @@ IMG_MC ?= ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/movement-controller:${DOCKER_TA
...
+
+.PHONY: docker-push-all
+docker-push-all:
```
hunchback, line 21: renaming to `docker-push` may be better

hunchback (approved):
added some comments which are optional

build/Makefile
froesef: `docker-build` and `docker-push` are already inherited from docker.mk. If I specify the targets again they will be overwritten.

roee88:
Can you explain why? I thought that you only use it for local testing and otherwise it is bundled in the manager

froesef:
We're using the image quite a bit for internal demos. And like this it's easy to access it. Also it can be very useful for the mover development as it helps to deploy a minimal setup to quickly test changes to the motion controller in a K8s integrated scenario.

PR #11, opened 2020-09-22
travis: fix failures on master

hunchback:


roee88 (commented):
We need to revisit what we do in Travis.

roee88 (approved):

PR #13, opened 2020-09-23
deploy: Dynamic namespaces

froesef:
This PR gets rid of the default m4d-system namespace. It's created twice currently. Once by kubectl create ns and once by from its definition in the yaml. 
The proposal is to only create it using kubectl create ns so that it can be changed dynamically to the value of `DOCKER_NAMESPACE`. That way e.g. when building and deploying to openshift the image will be in the correct openshift project. 
CAUTION: This will change the default namespace from `m4d-system` to `the-mesh-for-data` as this is the namespace that we're publishing the images to.

### DIFF
--- a/manager/Makefile
+++ b/manager/Makefile
@@ -26,14 +26,14 @@ docker-build: generate
 
 # Deploy only movement-controller in the configured Kubernetes cluster in ~/.kube/config
 deploy_mc: manifests $(TOOLBIN)/kustomize $(TOOLBIN)/kubectl
-	$(TOOLBIN)/kubectl create namespace m4d-system || true
+	$(TOOLBIN)/kubectl create namespace ${DOCKER_NAMESPACE} || true
 	cd config/manager && $(ABSTOOLBIN)/kustomize edit set image controller=${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/movement-controller:${DOCKER_TAGNAME}
-	$(TOOLBIN)/kustomize build --load_restrictor none config/movement-controller | $(TOOLBIN)/kubectl apply -f -
+	cd config/movement-controller && $(ABSTOOLBIN)/kustomize edit set namespace ${DOCKER_NAMESPACE}
+	$(TOOLBIN)/kustomize build --load_restrictor none config/movement-controller | $(TOOLBIN)/kubectl apply --namespace=${DOCKER_NAMESPACE} -f -
 
 undeploy_mc: manifests $(TOOLBIN)/kustomize $(TOOLBIN)/kubectl
-	$(TOOLBIN)/kubectl create namespace m4d-system || true
 	cd config/manager && $(ABSTOOLBIN)/kustomize edit set image controller=${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/movement-controller:${DOCKER_TAGNAME}
-	$(TOOLBIN)/kustomize build --load_restrictor none config/movement-controller | $(TOOLBIN)/kubectl delete -f -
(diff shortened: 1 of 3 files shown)


manager/Makefile
```diff
@@ -26,14 +26,14 @@ docker-build: generate
 
 # Deploy only movement-controller in the configured Kubernetes cluster in ~/.kube/config
 deploy_mc: manifests $(TOOLBIN)/kustomize $(TOOLBIN)/kubectl
-	$(TOOLBIN)/kubectl create namespace m4d-system || true
+	$(TOOLBIN)/kubectl create namespace ${DOCKER_NAMESPACE} || true
```
roee88, line 29: Tying this to the DOCKER_NAMESPACE is limiting and confusing. Just because it works without extra work in the internal openshift registry doesn't mean that we should force it to be like that. Instead, **if** you use the OpenShift internal registry then set DOCKER_NAMESPACE to the same value as `KUBE_NAMESPACE` (or an equivalent).

manager/Makefile
```diff
@@ -26,14 +26,14 @@ docker-build: generate
...
 	cd config/manager && $(ABSTOOLBIN)/kustomize edit set image controller=${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/movement-controller:${DOCKER_TAGNAME}
-	$(TOOLBIN)/kustomize build --load_restrictor none config/movement-controller | $(TOOLBIN)/kubectl apply -f -
+	cd config/movement-controller && $(ABSTOOLBIN)/kustomize edit set namespace ${DOCKER_NAMESPACE}
```
roee88, line 31: This modifies the YAMLs in the source tree

manager/config/manager/manager.yaml
```diff
@@ -35,7 +29,7 @@ spec:
           - name: ENABLE_WEBHOOKS
             value: "true"
           - name: MOVER_IMAGE
-            value: "image-registry.openshift-image-registry.svc:5000/mover/mover:1.0-SNAPSHOT"
+            value: "image-registry.openshift-image-registry.svc:5000/the-mesh-for-data/mover:latest"
```
roee88, line 32: This should by default point to the mover image from ghcr.io. For local testing Aidan has implemented something to override the image repo, we should make it cover this case as well if it doesn't.

manager/config/movement-controller/kustomization.yaml
```diff
@@ -2,36 +2,27 @@
...
-namespace: m4d-system
+namespace: the-mesh-for-data
 
 # Value of this field is prepended to the
 # names of all resources, e.g. a deployment named
 # "wordpress" becomes "alices-wordpress".
 # Note that it should also match with the prefix (text before '-') of the namespace
 # field above.
-namePrefix: m4d
+namePrefix: m4d-
```
roee88, line 12: Not used as indented (see the comment in the YAML)

roee88 (changes requested):
Hi @froesef 

It might be better to open an issue before submitting a PR for such changes (and in general).

As for the changes this PR introduces, I'm not sure that this is the right approach. Beyond the comments in the review I would recommend considering the following approach instead:
- Resources don't have `metadata.namespace` set 
- Don't reuse DOCKER_NAMESPACE
- Keep m4d-system as the default (because it's hardcoded elsewhere)
- Either modify all `kubectl apply` to have `-n $KUBE_NAMESPACE`, or just deploy to the current namespace (in that case the install instructions should include changing to the desired namespace before deploying). 

Also consider that m4d-system is hardcoded in different places across the codebase. It will take much more effort to fully allow setting it to something arbitrary.

roee88:
@froesef  Can I close this? We can re-open when we're ready to make that change.

roee88:
Closing, please re-open when relevant again

PR #15, opened 2020-09-23
create a new namespace for each blueprint

shlomitk1:
This PR takes care of the following:
    -Creating a dedicated namespace for a blueprint upon m4d app creation.
    -Deleting the namespace upon m4d app deletion.

This PR does not use hierarchical namespaces and does not handle roles in a new namespace.


manager/controllers/app/blueprint_mgr.go
```diff
@@ -27,10 +26,16 @@ func containsTemplate(templateList []app.ComponentTemplate, moduleName string) b
 
 // GetBlueprintSignature returns the signature (name and namespace) of the blueprint owned by the given M4DApplication instance
 func (r *M4DApplicationReconciler) GetBlueprintSignature(applicationContext *app.M4DApplication) *app.Blueprint {
+	//sanity - finalizer should exist, since it represents the blueprint namespace name
```
roee88, line 29: The use of finalizer name as the namespace indicator seems odd here isn't it? I always thought that the finalizer name should typically represent the controller itself, so the controller would have a chance to run all of its cleaning logic. Reusing it for storing the namespace seems a bit like an abuse.

The code related to finalizer is already not so clean so I suggest to consider the following:
1. Revert the changes in this PR and use a Status field instead to indicate the generated namespace name
2. Keep the code around finalizers independent of the cleanup logic and place the cleanup logic in a separate function (like `deleteExternalResources` in https://book.kubebuilder.io/reference/using-finalizers.html) that will handle all cleaning, including the namespace if indicated in the status.

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -272,3 +268,22 @@ func (r *M4DApplicationReconciler) GetAllModules() (map[string]*app.M4DModule, e
 	}
 	return moduleMap, nil
 }
+
+// CreateAppNamespace creates a namespace in which the blueprint and the relevant resources will be running
+// It returns the generated namespace name
+func (r *M4DApplicationReconciler) CreateAppNamespace(app *app.M4DApplication) (string, error) {
+	genNamespace := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{GenerateName: "m4d-"}}
```
roee88, line 275: Would be useful (for debugging) to add a label with the value of the application name

manager/controllers/app/m4dapplication_controller_test.go
```diff
@@ -388,11 +388,14 @@ var _ = Describe("M4DApplication Controller", func() {
 
 			// Create M4DApplication
 			Expect(k8sClient.Create(context.Background(), resource)).Should(Succeed())
+			time.Sleep(2 * time.Second)
```
roee88, line 391: Should probably use Eventually instead of relying on sleep

roee88 (changes requested):

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -91,24 +94,24 @@ func (r *M4DApplicationReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
 
 // reconcileFinalizers reconciles finalizers for M4DApplication
 func (r *M4DApplicationReconciler) reconcileFinalizers(applicationContext *app.M4DApplication) error {
-	// finalizer (Blueprint)
-	finalizerName := r.Name + ".finalizer"
-	hasFinalizer := ContainsFinalizer(applicationContext, finalizerName)
-
+	// finalizer (DataMeshBlueprint)
```
roee88, line 97: What is DataMesh btw? ;)

manager/controllers/app/m4dapplication_controller.go
simanadler: Agree with @roee88 that for debugging and logging purposes we need some way to see an association with the M4DApplication

shlomitk1: Added labels to the generated namespace.

manager/controllers/app/blueprint_mgr.go
shlomitk1: Done.

manager/controllers/app/m4dapplication_controller_test.go
shlomitk1: Done.

manager/controllers/app/m4dapplication_controller.go
shlomitk1: merge error... Re-applied merge.

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -119,6 +122,11 @@ func (r *M4DApplicationReconciler) reconcileFinalizers(applicationContext *app.M
 	}
 	// Make sure this CRD instance has a finalizer
 	if !hasFinalizer {
+		var finalizerName string
+		var errCreate error
+		if finalizerName, errCreate = r.CreateAppNamespace(applicationContext); errCreate != nil {
```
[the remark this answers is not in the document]
shlomitk1: Fixed. It is in the status field of m4dapplication.

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -272,3 +299,26 @@ func (r *M4DApplicationReconciler) GetAllModules() (map[string]*app.M4DModule, e
...
 	return moduleMap, nil
 }
+
+// CreateAppNamespace creates a namespace in which the blueprint and the relevant resources will be running
+// It stores the generated namespace name inside app status
+func (r *M4DApplicationReconciler) CreateAppNamespace(app *app.M4DApplication) error {
+	genNamespace := &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{GenerateName: "m4d-"}}
+	genNamespace.Labels = make(map[string]string)
+	genNamespace.Labels["ownerName"] = app.Name
+	genNamespace.Labels["ownerNamespace"] = app.Namespace
```
roee88, line 309: I think it can be one label (too many labels make cli output ugly). Also, should we prefix it with a domain to avoid collisions? How about something like:
```
genNamespace.Labels = map[string]string{
   "m4d.ibm.com.owner": app.Namespace + "." + app.Name
}
```

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -272,3 +299,26 @@ func (r *M4DApplicationReconciler) GetAllModules() (map[string]*app.M4DModule, e
...
+	if err := r.Create(context.Background(), genNamespace); err != nil {
+		return err
+	}
+	r.Log.V(0).Info("Created namespace " + genNamespace.Name + " for " + app.Namespace + "/" + app.Name)
+	app.Status.BlueprintNamespace = genNamespace.Name
+	return nil
+}
+
+// DeleteAppNamespace deletes the blueprint namespace upon blueprint deletion
+func (r *M4DApplicationReconciler) DeleteAppNamespace(name string) error {
```
roee88, line 319: This is a generic `DeleteNamespace`, not tried to App. Do we have somewhere to move this to?

manager/controllers/app/blueprint_mgr.go
```diff
@@ -27,10 +26,15 @@ func containsTemplate(templateList []app.ComponentTemplate, moduleName string) b
 
 // GetBlueprintSignature returns the signature (name and namespace) of the blueprint owned by the given M4DApplication instance
 func (r *M4DApplicationReconciler) GetBlueprintSignature(applicationContext *app.M4DApplication) *app.Blueprint {
+	//sanity - app.Status.BlueprintNamespace should exist, since it represents the blueprint namespace name
+	if len(applicationContext.Status.BlueprintNamespace) == 0 {
+		return nil
+	}
```
roee88, line 32: I had some issues understanding the flow and I suspect it has to do with this sanity check

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -139,6 +140,31 @@ func ContainsFinalizer(obj *app.M4DApplication, finalizer string) bool {
 	return false
 }
 
+// DeleteExternalResources deletes all resources allocated by the controller: blueprint, storage, namespace
+func (r *M4DApplicationReconciler) DeleteExternalResources(applicationContext *app.M4DApplication) error {
```
roee88, line 144: I think that this function don't have to be exported

manager/apis/app/v1alpha1/m4dapplication_types.go
```diff
@@ -111,6 +111,9 @@ type M4DApplicationStatus struct {
 	// whether reconcile was called because the desired state changed, or whether the Blueprint status changed.
 	// +optional
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// BlueprintNamespace represents the namespace where the blueprint (and the relevant resources) will be allocated.
+	BlueprintNamespace string `json:"blueprintNamespace,omitempty"`
```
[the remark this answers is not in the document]
shlomitk1: Added.

manager/controllers/app/blueprint_mgr.go
```diff
@@ -27,10 +26,15 @@ func containsTemplate(templateList []app.ComponentTemplate, moduleName string) b
 
 // GetBlueprintSignature returns the signature (name and namespace) of the blueprint owned by the given M4DApplication instance
 func (r *M4DApplicationReconciler) GetBlueprintSignature(applicationContext *app.M4DApplication) *app.Blueprint {
+	//sanity - app.Status.BlueprintNamespace should exist, since it represents the blueprint namespace name
+	if len(applicationContext.Status.BlueprintNamespace) == 0 {
```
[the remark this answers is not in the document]
shlomitk1: OK.

manager/controllers/app/m4dapplication_controller.go
shlomitk1: Done.

manager/controllers/app/m4dapplication_controller.go
shlomitk1: Changed the name. I would leave this function here because it is going to be changed, together with the create function, when we will manage subNamespaceAnchors instead of namespaces (HNC related)

manager/controllers/app/m4dapplication_controller.go
shlomitk1: OK

roee88 (approved):

simanadler (approved):

PR #16, opened 2020-09-23
scripts: docker password is not always available in Travis

roee88:
DOCKER_PASSWORD was set in plaintext in the Travis settings. I had to change it to an encrypted variable and reset it to a new token. Since then, Travis builds fail.

The reason is that pull requests from forks are not allowed to use secure variables. See https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions.

This pull request changes the code as follows:

1. Checks the existence of DOCKER_PASSWORD rather than DOCKER_USERNAME (NOTE: I believe this is preferable to hiding the username as well, since it was useful for me in the past especially since Travis replaces that value with `[secure]` even if it's taken from something other than the DOCKER_USERNAME)
2. Removes duplication of  the variables defined in ibmcloud.mk
3. Removes cloud-registry secret from deploy-secret-provider.sh (the image of the secret provider are public)
4. **Temporarily disables the tests for `pkg/helm`**. This is done because these tests assume being able to push to a registry. However, the Travis script doesn't set it to run against a local registry (runs against ghcr.io instead).

### DIFF
--- a/.mk/docker.mk
+++ b/.mk/docker.mk
@@ -1,20 +1,17 @@
-DOCKER_PASSWORD ?= $(shell echo "${DOCKER_PASSWORD_ENCODED}" | base64 -d)
-DOCKER_HOSTNAME ?= ghcr.io
-DOCKER_NAMESPACE ?= the-mesh-for-data
+#export DOCKER_USERNAME ?= ""
+#export DOCKER_PASSWORD ?= ""
+export DOCKER_HOSTNAME ?= ghcr.io
+export DOCKER_NAMESPACE ?= the-mesh-for-data
+export DOCKER_TAGNAME ?= latest
+
 DOCKER_NAME ?= m4d
-DOCKER_TAGNAME ?= latest
+DOCKER_FILE ?= Dockerfile
 DOCKER_CONTEXT ?= .
-GO_INPUT_FILE ?= main.go
-GO_OUTPUT_FILE ?= manager
+
 IMG ?= ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/${DOCKER_NAME}:${DOCKER_TAGNAME}
 
-export DOCKER_USERNAME
-export DOCKER_PASSWORD
-export DOCKER_HOSTNAME
-export DOCKER_NAMESPACE
-export DOCKER_TAGNAME
-
-DOCKER_FILE ?= Dockerfile
+GO_INPUT_FILE ?= main.go
+GO_OUTPUT_FILE ?= manager
 
 .PHONY: docker-all
 docker-all: docker-build docker-push
@@ -29,7 +26,7 @@ docker-build-local:
 
 .PHONY: docker-push
 docker-push:
-ifdef DOCKER_USERNAME
+ifdef DOCKER_PASSWORD
 	@docker login \
 		--username ${DOCKER_USERNAME} \
 		--password ${DOCKER_PASSWORD} ${DOCKER_HOSTNAME}
--- a/.mk/docker.mk
+++ b/.mk/docker.mk
@@ -26,7 +26,7 @@ docker-build-local:
 
 .PHONY: docker-push
 docker-push:
-ifdef DOCKER_PASSWORD
+ifneq (${DOCKER_PASSWORD},)
 	@docker login \
 		--username ${DOCKER_USERNAME} \
 		--password ${DOCKER_PASSWORD} ${DOCKER_HOSTNAME}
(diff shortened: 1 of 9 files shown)


.mk/docker.mk
```diff
@@ -1,20 +1,17 @@
-DOCKER_PASSWORD ?= $(shell echo "${DOCKER_PASSWORD_ENCODED}" | base64 -d)
-DOCKER_HOSTNAME ?= ghcr.io
-DOCKER_NAMESPACE ?= the-mesh-for-data
+#export DOCKER_USERNAME ?= ""
+#export DOCKER_PASSWORD ?= ""
```
hunchback, line 2: delete

hunchback, line 2:
```
export DOCKER_USERNAME ?=
export DOCKER_PASSWORD ?=
```

else it will stop supporting private repos (which may be used for tests)

.mk/docker.mk
```diff
@@ -29,7 +26,7 @@ docker-build-local:
 
 .PHONY: docker-push
 docker-push:
-ifdef DOCKER_USERNAME
+ifdef DOCKER_PASSWORD
```
hunchback, line 29:
```
ifeq (${DOCKER_PASSWORD},)
```

.mk/helm.mk
```diff
@@ -9,7 +9,7 @@ export HELM_EXPERIMENTAL_OCI=1
 
 .PHONY: helm-login
 helm-login: $(TOOLBIN)/helm
-ifdef DOCKER_USERNAME
+ifdef DOCKER_PASSWORD
```
hunchback, line 12:
```
ifeq (${DOCKER_PASSWORD},)
```

.mk/docker.mk
roee88: Travis unsets DOCKER_PASSWORD

.mk/docker.mk
roee88: will change, thanks. 

Regarding your comment "which may be used for tests", note the issue that this PR tries to resolve. We will only be able to do such tests in master builds, not in PRs.

secret-provider/deployment/deploy-secret-provider.sh
```diff
@@ -5,21 +5,8 @@
...
-                --docker-server="$DOCKER_HOSTNAME" \
-                --docker-username="$DOCKER_USERNAME" \
-                --docker-password="$DOCKER_PASSWORD" \
-                -n $KUBE_NAMESPACE
-    fi
-
-    kubectl patch serviceaccount secret-provider -p \
-    '{"imagePullSecrets": [{"name": "cloud-registry"}]}' \
-    -n $KUBE_NAMESPACE
-
```
hunchback, line 22: we should add an extra variable `DOCKER_SECRET` and if set we should create the k8s secret, thus supporting working against a private password protected registry

.mk/docker.mk
hunchback: yes (if you revert the export statement above)

roee88: changed

secret-provider/deployment/deploy-secret-provider.sh
roee88: This section was only required when the secret provider itself used a private image. Since it doesn't, I don't see the point.

hunchback:
i see ... correct

pkg/helm/helm.go
```diff
@@ -185,17 +185,14 @@ func (r *Impl) RegistryLogin(hostname string, username string, password string,
...
-	if username != "" {
-		cfg, err := getConfig("")
-		if err != nil {
-			return err
-		}
-		logout := action.NewRegistryLogout(cfg)
-		var buf bytes.Buffer
-		return logout.Run(&buf, hostname)
+func (r *Impl) RegistryLogout(hostname string) error {
+	cfg, err := getConfig("")
```
froesef, line 189: If the method gets called but DOCKER_USERNAME and DOCKER_PASSWORD are not specified (e.g. when using local registry that does not need any credentials) does this function fail in `getConfig("")` ? Or can the `logoug.Run` deal with this situation?

roee88: It isn't directly related to not having username/password. It throws an error "not logged in"  if logout is called but no login was done before.  I think is the expected behavior. 

Caller code should only call logout if login returned successfully. I think that changing RegistryLogin to not silently ignore being called without username/password would be the right approach. What you agree?

roee88: I updated the code according to the logic described in the message above

froesef (approved):

PR #18, opened 2020-09-24
manager: Change the default MOVER_IMAGE to the public ghcr.io image

froesef:


roee88 (approved):
There are uses of dummy mover as default in the https://github.com/IBM/the-mesh-for-data/blob/8b795d0138dbfabb696197d0009d80c07c63ce1f/modules/m4d-db2wh/templates/batchtransfer.yaml and https://github.com/IBM/the-mesh-for-data/blob/8b795d0138dbfabb696197d0009d80c07c63ce1f/modules/m4d-kafka/templates/streamtransfer.yaml

Do you want to change these to the regular mover image instead? We need to re-write them to a single module without hard coding so not sure if a must, but would be nice.

PR #23, opened 2020-09-28
Add fields to partition by a certain column when writing to S3/COS

froesef:


roee88 (approved):
This is the first field that is only applicable for destination and not target. Do we expect more fields like that?

Issue #17, opened 2020-09-23
Helm chart resources are not created in an intended workspace


shlomitk1:
When helm is applied from command line with an option --namespace, the behavior is correct: all resources are allocated in the specified namespace. For example:
`/hack/tools/bin/helm upgrade --no-hooks --install test --namespace xxx --values values.yaml .`

When pkg/helm.go functions are invoked in `applyChartResource` of Blueprint reconciler, the resources are allocated in the current namespace, although the correct namespace name is passed to the install/upgrade methods of the helm package.
A hack is to specify namespace for each resource using` namespace: {{ .Values.metadata.namespace }}`

roee88:
I ran the `TestHelmRelease` test against a different namespace than the current one in context (changed `kubeNamespace`) and it did create the chart in the right namespace (the one set in `kubeNamespace` and passed to Install and not the current namespace). So this functions seems to be working properly. 

Have you debugged applyChartResource and observed that the right namespace is passed to `Helmer.Install`? I need more details.

shlomitk1:
Yes, this is exactly the problem. Chart is created in the correct namespace, the actual resources are not.
When current namespace is m4d-system and the specified namespace is m4d-cwd6t, I get the resources in m4d-system and the chart in m4d-cwd6t. By resources I mean service + deployment + configMap + serviceAccount yamls.
```
helm list --namespace=m4d-cwd6t
NAME                 	NAMESPACE	REVISION	UPDATED                                	STATUS  	CHART                    	APP VERSION
r417b5910533125d0e4f1	m4d-cwd6t	4       	2020-09-23 10:50:03.958106485 +0000 UTC	deployed	arrow-flight-module-0.0.1	1.0     
```

I've checked that **install.Namespace** is correct (see the function below).
```
func (r *Impl) Install(chart *chart.Chart, kubeNamespace string, releaseName string, vals map[string]interface{}) (*release.Release, error) {
	cfg, err := getConfig(kubeNamespace)
	if err != nil {
		return nil, err
	}
	install := action.NewInstall(cfg)
	install.ReleaseName = releaseName
	install.Namespace = kubeNamespace
	return install.Run(chart, vals)
}
```

roee88:
I see. The problem is in pkg/helm. It uses https://github.com/helm/helm/blob/b0fdb5461fbc7e1b52f61613d3e7a22d85c7a794/pkg/cli/environment.go#L178 which targets the current namespace.

[...]

PR #24, opened 2020-09-29
environment object was created with a wrong namespace

shlomitk1:
Fixes https://github.com/IBM/the-mesh-for-data/issues/17

While creating cli environment for helm install/upgrade operation, the namespace is taken from HELM_NAMESPACE env variable.
This PR modifies HELM_NAMESPACE env variable s.t. the resources will be created in the right namespace.

### DIFF
--- a/pkg/helm/helm.go
+++ b/pkg/helm/helm.go
@@ -15,9 +15,7 @@ import (
 	"helm.sh/helm/v3/pkg/release"
 )
 
-var (
-	settings = cli.New()
-)
+var settings *cli.EnvSettings = nil
 
 func getConfig(kubeNamespace string) (*action.Configuration, error) {
 	actionConfig := new(action.Configuration)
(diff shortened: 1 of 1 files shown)


pkg/helm/helm.go
```diff
@@ -15,17 +15,16 @@ import (
...
+var settings *cli.EnvSettings = nil
 
 func getConfig(kubeNamespace string) (*action.Configuration, error) {
 	actionConfig := new(action.Configuration)
 
 	if kubeNamespace == "" {
 		kubeNamespace = "default"
 	}
-
+	os.Setenv("HELM_NAMESPACE", kubeNamespace)
```
roee88, line 26: race condition if this is called by multiple threads at once

roee88 (approved):
Please ask to be assigned on the issue before submitting a PR.
I will keep the issue open and do a follow-up to fix this without concurrency issues.

Issue #25, opened 2020-09-29
manager: allocate Helm resources in the intended namespace


shlomitk1:
The current solution is to set HELM_NAMESPACE before creating an appropriate action for install/upgrade.
Enhancement: create Config without using cli package to make it thread-safe.

PR #29, opened 2020-09-29
helm: allocate resources in the specified namespace

shlomitk1:
This PR fixes https://github.com/IBM/the-mesh-for-data/issues/25

### DIFF
--- a/pkg/helm/helm.go
+++ b/pkg/helm/helm.go
@@ -11,11 +11,13 @@ import (
 
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart"
-	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/release"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
 )
 
-var settings *cli.EnvSettings = nil
+var (
+	debugOption = (os.Getenv("HELM_DEBUG") == "true")
+)
 
 func getConfig(kubeNamespace string) (*action.Configuration, error) {
 	actionConfig := new(action.Configuration)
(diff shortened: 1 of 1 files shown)


pkg/helm/helm.go
```diff
@@ -11,21 +11,25 @@ import (
...
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart"
-	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/release"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
 )
 
-var settings *cli.EnvSettings = nil
+var (
+	debugOption = (os.Getenv("HELM_DEBUG") == "true")
```
roee88, line 19: We should probably use a real logger instead of these copies from the Helm package, but OK for now.

pkg/helm/helm.go
```diff
@@ -145,6 +149,7 @@ func (r *Impl) Install(chart *chart.Chart, kubeNamespace string, releaseName str
 	install := action.NewInstall(cfg)
 	install.ReleaseName = releaseName
 	install.Namespace = kubeNamespace
+	install.Wait = true
```
roee88, line 152: How does this work when used from a reconcile loop? Since it's a long running process is it more likely to have overlapping installs now? Are we dealing with it in the blueprint reconciler?

roee88 (commented):
Can you please separate it to two PRs?

shlomitk1:
> 
> 
> Can you please separate it to two PRs?

Done. Adding  "wait" will be a part of another PR including changes in blueprint reconciler as well.

pkg/helm/helm.go
shlomitk1: Yes, but not in this PR, thus I'm removing the wait argument for now.

roee88 (approved):

PR #34, opened 2020-09-30
travis: speedup

hunchback:


.travis.yml
```diff
@@ -125,44 +173,66 @@ jobs:
...
     - REGISTRY=ibmcloud
     install:
     - make install-tools
-    script:
-    - make $CLUSTER || travis_terminate 1;
-    - make cluster-prepare || travis_terminate 1;
-    - make -C $COMPONENT deploy
-  - stage: verify-deploy
-    name: "Verify deploy"
+    script: true
```
roee88, line 184: Why a `deploy` and not directly a script? I also don't think that `deploy.script` is a list IIRC

hunchback: as this will be changed to deployment on a OpenShift or shared cluster in future

hunchback: correct. will fix

roee88 (approved):

PR #38, opened 2020-09-30
Speedup docker builds by cross-compiling go binaries local

froesef:
Building the connector docker containers currently takes 5 minutes as it loads the whole directory into the docker context (which is then almost 1GB) and needs to download a golang environment and compile it in a docker container. Then the compiled file is just added to the image resulting in an image that is 588MB.

This new method builds the binaries directly and copies them into an empty docker container. Resulting in an image that is around 17MB and a build time that is multiples faster.

### DIFF
--- a/connectors/opa/Makefile
+++ b/connectors/opa/Makefile
@@ -8,6 +8,11 @@ include $(ROOT_DIR)/.mk/docker.mk
 OPA_PATH=${ABSTOOLBIN}/opa
 OPA_DEFAULT_POLICY_FILE=${ROOT_DIR}/third_party/opa/opa-policy.rego
 
+docker-build:
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o opa-connector opa_connector.go evaluate_opa.go
+	docker build . -t $(IMG) -f Dockerfile
+	(rm opa-connector || true)
+
 .PHONY: opaserver
 opaserver:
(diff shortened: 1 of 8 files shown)


hunchback (approved):

connectors/opa/Makefile
```diff
@@ -8,6 +8,11 @@ include $(ROOT_DIR)/.mk/docker.mk
 OPA_PATH=${ABSTOOLBIN}/opa
 OPA_DEFAULT_POLICY_FILE=${ROOT_DIR}/third_party/opa/opa-policy.rego
 
+docker-build:
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o opa-connector opa_connector.go evaluate_opa.go
+	docker build . -t $(IMG) -f Dockerfile
+	(rm opa-connector || true)
```
roee88, line 14: why not just `rm opa-connector`?

manager/Makefile
```diff
@@ -21,7 +21,7 @@ build-linux: generate vet
 
 # Overwrite docker-build from docker.mk
 docker-build: generate
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o manager main.go
 	docker build . -t ${IMG} -f Dockerfile
```
roee88, line 26: No cleanup of `manager` binary here? Alternatively we can avoid those cleanups and always build to ROOTDIR/bin that will be added to .gitignore.

PR #41, opened 2020-10-01
travis: Remove fmt and vet from `make test`

froesef:
Removes the fmt and vet commands from `make test`. This is already covered by travis in `make verify`


hunchback (approved):

roee88:
make build also runs fmt and vet (in the `manager` make target)

PR #48, opened 2020-10-02
connectors: vault connector unit tests

rohithdv:
This PR does the following changes to vault connector:
(1) Minor redesign of vault connector to make it suitable for unit tests
(2) Preliminary unit test for vault connector

To run the unit test, execute the following:
(1) cd to connectors/vault
(2) go test -v ./...

Note regarding the changes in this PR:
For vault connector unit tests we need to mock vault server. We are using using vault's public testing api for this purpose. The following refs explain how to use hashicorp vault's public testing api.
(1) https://stackoverflow.com/questions/57771228/mocking-hashicorp-vault-in-go
(2) https://www.youtube.com/watch?v=yszygk1cpEc&feature=youtu.be&t=3238
(3) go mod related changes for using hashicorp's public testing api
(i) https://github.com/hashicorp/vault/issues/9575
(ii) https://github.com/kubernetes/client-go/issues/628


connectors/vault/Dockerfile
```diff
@@ -1,7 +1,19 @@
...
+COPY pkg/connectors/protobuf protobuf
+
+WORKDIR /m4d/connectors/vault
+COPY connectors/vault .
+
+RUN go get -v ./...
+
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o vault-connector *.go
 
 EXPOSE 50083
```
hunchback, line 18: please align build of binaries according to what @froesef  has done elsewhere (building outside of Docker for better performance)

connectors/vault/go.mod
```diff
@@ -7,7 +7,17 @@ replace github.com/ibm/the-mesh-for-data v0.0.0 => ../..
 go 1.13
 
 require (
+	//https://github.com/kubernetes/client-go/issues/628
```
hunchback, line 10: remove this file (not needed once building outside docker)

connectors/vault/user_vault_init.sh
```diff
@@ -25,7 +25,8 @@ export VAULT_BIN=$DIR_ABSTOOLBIN/vault
 export USER_VAULT_KAFKA_ASSET_KEY="{\"catalog_id\":\"87ffdca3-8b5d-4f77-99f9-0cb1fba1f73f\",\"asset_id\":\"01c6f0f0-9ffe-4ccc-ac07-409523755e72\"}"
 export USER_VAULT_KAFKA_ASSET_CRED="my_kafka_credentials"
 
-export USER_VAULT_EGERIA_ASSET_KEY="{\"ServerName\":\"cocoMDS3\",\"AssetGuid\":\"4098e18e-bd53-4fd0-8ff8-e1c8e9fc42da\"}"
+#export USER_VAULT_EGERIA_ASSET_KEY="{\"ServerName\":\"cocoMDS3\",\"AssetGuid\":\"4098e18e-bd53-4fd0-8ff8-e1c8e9fc42da\"}"
+export USER_VAULT_EGERIA_ASSET_KEY="{\"ServerName\":\"cocoMDS3\",\"AssetGuid\":\"f6d9bf8c-dd37-4747-bca9-8ca9d1a5bb8f\"}"
```
hunchback, line 29: how is this change related to this PR?

connectors/vault/testutil/helper.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func GetEnvironment() string {
+	userVaultPath := EnvValues["USER_VAULT_PATH"]
+	log.Printf("EnvVariables = %v\n", EnvValues)
+
+	return userVaultPath
+}
+
+func createDatasetRead(datasetID string) *pb.DatasetContext {
```
etykha, line 30: is this function used?

connectors/vault/testutil/helper.go
```diff
@@ -0,0 +1,51 @@
...
+	dataset := &pb.DatasetIdentifier{DatasetId: datasetID}
+	operation := &pb.AccessOperation{Type: pb.AccessOperation_READ}
+	datasetContext := &pb.DatasetContext{Dataset: dataset, Operation: operation}
+	return datasetContext
+}
+
+func createDatasetTransferFirst(datasetID string) *pb.DatasetContext {
```
etykha, line 37: don't see this function used

connectors/vault/testutil/helper.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2020 IBM Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+package testutil
```
etykha, line 4: why we need helper file if it has only 2 short functions used in only 1 other file? such helper files usually serve several other files. Do you expect tests to grow to several other files?

connectors/vault/vaultconn-bl/vault_connector_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/hashicorp/vault/api"
+	"github.com/hashicorp/vault/http"
+	"github.com/hashicorp/vault/vault"
+
+	tu "github.com/ibm/the-mesh-for-data/connectors/vault/testutil"
+	vltutils "github.com/ibm/the-mesh-for-data/connectors/vault/vault_utils"
+	pb "github.com/ibm/the-mesh-for-data/pkg/connectors/protobuf"
+)
+
+func TestVaultConnectorNormalRun(t *testing.T) {
```
etykha, line 19: each test function should have comment explaining what exactly is tested here

connectors/vault/vaultconn-bl/vault_connector_test.go
```diff
@@ -0,0 +1,98 @@
...
+	conf.Address = addr
+
+	client, err := api.NewClient(conf)
+	if err != nil {
+		t.Fatal(err)
+	}
+	client.SetToken(rootToken)
+
+	// Setup required secrets, policies, etc.
+	_, err = client.Logical().Write("secret/foo", map[string]interface{}{
```
etykha, line 81: Do we use this secret/foo ?

etykha: same for "bar". should be deleted if not used

connectors/vault/Dockerfile
rohithdv: fixed. Thanks

connectors/vault/go.mod
rohithdv: fixed. Thanks.

connectors/vault/user_vault_init.sh
rohithdv: This is needed to test the PR for the entire flow (as the assets which were created earlier now no longer exist). Although this change is not relevant to this PR, we need to change this to validate there is no change in the normal flow due to changes done in this PR.

connectors/vault/vaultconn-bl/vault_connector.go
```diff
@@ -0,0 +1,33 @@
+// Copyright 2020 IBM Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+package vaultconnbl
```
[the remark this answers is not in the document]
rohithdv: renamed this file to vault_reader.go for better readability.

connectors/vault/testutil/helper.go
rohithdv: removed this function. Thanks.

connectors/vault/testutil/helper.go
rohithdv: I think since we have followed this design in all connectors we can conform to this design for now. It is more for uniformity of design that this is being followed. Also it makes the connectors independent of each other and also self contained in my opinion.

connectors/vault/vaultconn-bl/vault_connector_test.go
rohithdv: fixed. Thanks.

connectors/vault/go.mod
rohithdv: In this PR we need to edit the private go.mod to have some packages exclusively for running unit tests. In we do not have a private go.mod then we have to move the contents of this file to the root level go.mod. If we do that, when we run make verify, the command automatically removes all packages not used in go files causing the unit tests to fail. The make verify command does not seem to check if the package is being used in test files. Hence for this purpose, one option is to keep using the private go.mod file with specific packages needed for test files listed in this private go.mod file. Please suggest if there are any other alternatives to this issue. Thanks.

connectors/vault/vault_connector.go
```diff
@@ -5,52 +5,52 @@ package main
...
-	log.Printf("Vault connector: GetCredentialsInfo, in = %v\n", in)
+	log.Println("*********************************Received ApplicationContext *****************************")
+	log.Println(in)
+	log.Println("******************************************************************************************")
 
-	// secretaddr := url.QueryEscape(in.DatasetId)
-	secretaddr := strings.ReplaceAll(in.DatasetId, " ", "")
+	vaultPathKey := GetEnv(VaultPathKey)
+	eval, err := s.vaultConnector.GetCredentialsInfo(in, vault, vaultPathKey)
+	jsonOutput, _ := json.MarshalIndent(eval, "", "\t")
```
shlomitk1, line 36: Check err before obtaining json

connectors/vault/vault_utils/vault_utils.go
```diff
@@ -135,3 +137,24 @@ func (vlt *VaultConnection) GetFromVault(innerVaultPath string) (string, error)
 
 	return string(b), nil
 }
+
+// GetFromVault returns the credentials from vault as json
+func (vlt *VaultConnection) AddToVault(innerVaultPath string, credentialsMap map[string]interface{}) (string, error) {
```
shlomitk1, line 142: Comment is not related to the function.

connectors/vault/Makefile
```diff
@@ -10,7 +10,7 @@ VAULT_PATH=${ABSTOOLBIN}/vault
 include $(ROOT_DIR)/pkg/policy-compiler/policy-compiler.env
 
 docker-build:
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o vault-connector vault_connector.go vault_utils.go
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -o vault-connector vault_connector.go
```
froesef, line 13: Please add the ` go test -v ./...` also in the test target in line 50 in this Makefile. Otherwise the CI won't run the test that you just added.

rohithdv: fixed. Thanks.

connectors/vault/vault_utils/vault_utils.go
rohithdv: fixed. Thanks

connectors/vault/vault_connector.go
rohithdv: fixed. Thanks

connectors/vault/vaultconn-bl/vault_reader.go
```diff
@@ -0,0 +1,33 @@
...
+
+	// secretaddr := url.QueryEscape(in.DatasetId)
+	secretaddr := strings.ReplaceAll(in.DatasetId, " ", "")
+
+	readCredentials, err := vault.GetFromVault(vaultPathKey, secretaddr)
+	if err != nil {
+		log.Printf("Error in vaultConnector, got error from the vault: %v\n", err.Error())
+		return nil, fmt.Errorf("error in retrieving the secret from vault(key = %s): %v", secretaddr, err)
+	}
+	log.Println("Read credentials from vault: " + readCredentials)
```
froesef, line 30: Please do not log credentials

connectors/vault/vaultconn-bl/vault_reader.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func (s *Server) GetCredentialsInfo(in *pb.DatasetCredentialsRequest, vault vaultutils.VaultConnection, vaultPathKey string) (*pb.DatasetCredentials, error) {
+	log.Println("Vault connector: GetCredentialsInfo")
+	log.Printf("Vault connector: GetCredentialsInfo, in = %v\n", in)
+
+	// secretaddr := url.QueryEscape(in.DatasetId)
+	secretaddr := strings.ReplaceAll(in.DatasetId, " ", "")
+
+	readCredentials, err := vault.GetFromVault(vaultPathKey, secretaddr)
```
froesef, line 25: As far as I can see only `Client` from `VaultConnection` is used. The method `GetFromVault` can be rewritten to only use `api.Client` as input instead the whole `VaultConnection` that it does not use. That would also make the test easier as `VaultConfig` and `VaultConnection` are not needed.

connectors/vault/vaultconn-bl/vault_reader_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+func createTestVault(t *testing.T) (net.Listener, *api.Client) {
+	t.Helper()
+
+	// Create an in-memory, unsealed core (the "backend", if you will).
+	core, keyShares, rootToken := vault.TestCoreUnsealed(t)
+	_ = keyShares
+
+	// Start an HTTP server for the core.
+	ln, addr := http.TestServer(t, core)
```
froesef, line 69: Why do we need an http server if we run the vault tests against an in-memory vault client?

connectors/vault/vaultconn-bl/vault_reader.go
rohithdv: sometimes we may need to check for debugging purposes if the correct credentials are being retrieved. so can we have a flag based approach here. if the flag is set log the credentials otherwise dont log it. is this ok? @froesef

connectors/vault/vaultconn-bl/vault_reader_test.go
rohithdv: This is starting in memory vault server per unit test to avoid writing brittle unit tests. It is a public api exposed by vault  which is useful for making the unit test very self contained. You can check the explanation here in this video from 53.58 to  57.51 seconds (https://www.youtube.com/watch?v=yszygk1cpEc&feature=youtu.be&t=3238).

connectors/vault/vaultconn-bl/vault_reader.go
rohithdv: Though only `Client ` from `VaultConnection` is being used in the `GetFromVault` function, the `vault` object in Line 25 is being reused after [it was created](https://github.com/rohithdv/the-mesh-for-data/blob/255e89d8e98483b9261bbd3f54186f2259d74088/connectors/vault/vault_connector.go#L60) as part of initialisation when the vault connector started. During this initialisation, the `VaultConfig` object is being used. We are trying to reuse a single object for different purposes so that the entire functionality can be linked to a single `VaultConnection` object. Please let me know if this is ok. Thanks. @froesef

froesef: If you move the initialization into the `CreateVaultConnection` you don't even need a `VaultConnection` struct because you're only working on `api.Client` afterwards. Please correct me if I'm getting this wrong. But then the `VaultConnection` struct is only used to organize and group the GetFromVault and AddToVault methods under one structure. 
Instead of `func (vlt *VaultConnection) GetFromVault(vaultPathKey string, innerVaultPath string) (string, error)` you could have `func GetFromVault(client *api.Client, vaultPathKey string, innerVaultPath string) (string, error)`

connectors/vault/vaultconn-bl/vault_reader.go
froesef: A flag also introduces potential risks. Credentials should never be logged. If you need something for debugging or testing there are probably better ways to test this. 

I think in this setup you have a vault server that is mocked and a GetCredentials method that returns the credentials that you can then check using assert checks in the test. If you're worried about different things that you can put into vault you can write tests about what happens when e.g. wrong input get's put into vault.

But any logging of credentials is a potential security issue.

connectors/vault/vaultconn-bl/vault_reader.go
rohithdv: Actually at a high level `VaultConnection` is abstracting out the vault related functions (which are defined in vault_utils.go). If we dont have `VaultConnection` struct, then 
(1) there will be changes to the `InitVault() ` . we will need to pass Config object to it as a parameter.
(2) Also there will be changes to` GetFromVault() `and `AddToVault()`  as you have mentioned. 
(3) Also we need to import vault related libraries in vault_connector.go (`"github.com/hashicorp/vault/api"`) which was earlier imported in vault_utils.go.
(4)` CreateVaultConnection()` should return` *api.Client` instead of `VaultConnection`. 
 Should we go ahead in this direction ? Please advise. Thanks. @froesef

connectors/vault/vaultconn-bl/vault_reader.go
rohithdv: ok sure. So writing tests is the only approach as you have mentioned. So there is no way we can at runtime verify if the correct credentials have been got. I will remove the logging. Thanks.

connectors/vault/vaultconn-bl/vault_reader.go
froesef: I think in the end it's a question of programming style. I would do it the way you described earlier. But if you want to keep `VaultConnection` in order to keep everything encapsulated behind it that's also fine. I think we're all quite new to Go.

roee88:
replaced by https://github.com/IBM/the-mesh-for-data/pull/164

connectors/vault/vaultconn-bl/vault_reader.go
rohithdv: I agree with your view. Since it is mainly a matter of programming style I think it would be ok to keep the design as it is now if it is fine with you. This design had emerged due to discussion with other governance team members. Unless there is a real need to change the design I would prefer to keep the design unchanged. Please let me know if it is ok. Thanks.

PR #53, opened 2020-10-11
enhancing helm package functionality

shlomitk1:
This PR adds the following functionality to the helm package:
- A wait option (Helm will report that the status is deployed after the resources (pods,services,...) are ready.
- GetResources method that returns a list of deployed resources and their current status
Added the following tests:
- Check for Deployed release status
- Check for the correct number of deployed resources
- Check for the resource readiness using kstatus package.


go.mod
```diff
@@ -26,9 +26,11 @@ require (
 	istio.io/gogo-genproto v0.0.0-20191009201739-17d570f95998 // indirect
 	k8s.io/api v0.18.6
 	k8s.io/apimachinery v0.18.6
+	k8s.io/cli-runtime v0.18.4
 	k8s.io/client-go v0.18.6
 	rsc.io/letsencrypt v0.0.3 // indirect
 	sigs.k8s.io/controller-runtime v0.6.2
+	sigs.k8s.io/kustomize/kstatus v0.0.2
```
roee88, line 33: I thought sigs.k8s.io/kustomize/kstatus has moved to sigs.k8s.io/cli-utils/pkg/kstatus

manager/controllers/app/blueprint_controller.go
```diff
@@ -116,12 +116,12 @@ func (r *BlueprintReconciler) applyChartResource(ctx context.Context, log logr.L
 
 	rel, err := r.Helmer.Status(kubeNamespace, releaseName)
 	if err == nil && rel != nil {
-		rel, err = r.Helmer.Upgrade(chart, kubeNamespace, releaseName, vals)
+		rel, err = r.Helmer.Upgrade(chart, kubeNamespace, releaseName, vals, false)
```
roee88, line 119: Why are we adding Wait flag if we're not using it? Is there a plan to use it? Isn't it too risky to do a synchronic wait as part of reconcile (rather than polling)?

pkg/helm/helm.go
```diff
@@ -248,3 +259,34 @@ func (r *Impl) ChartPull(ref string) error {
...
+	config, err = getConfig(kubeNamespace)
+	if err != nil || config == nil {
+		return resources, err
+	}
+	status := action.NewStatus(config)
+	rel, err = status.Run(releaseName)
+	if err != nil {
+		return resources, err
+	}
+	resourceList, _ = config.KubeClient.Build(bytes.NewBufferString(rel.Manifest), false)
```
roee88, line 279: Is `_` an ignored error?

pkg/helm/helm.go
```diff
@@ -248,3 +259,34 @@ func (r *Impl) ChartPull(ref string) error {
...
+	for _, res := range resourceList {
+		if err := res.Get(); err != nil {
+			return resources, err
+		}
+		obj := res.Object
+		if unstr, ok := obj.(*unstructured.Unstructured); ok {
```
roee88, line 285: Why are we sure that this is always convertible?

pkg/helm/helm_test.go
```diff
@@ -115,18 +118,29 @@ func TestHelmRelease(t *testing.T) {
...
 	Log(t, "install", err)
 
-	_, err = impl.Upgrade(origChart, kubeNamespace, releaseName, vals)
+	_, err = impl.Upgrade(origChart, kubeNamespace, releaseName, vals, true)
 	assert.Nil(t, err)
 	Log(t, "upgrade", err)
 
-	_, err = impl.Status(kubeNamespace, releaseName)
-	assert.Nil(t, err)
+	assert.Eventually(t, func() bool {
```
roee88, line 129: Seems to check status update rather than get resources, is it necessary?

go.mod
shlomitk1: I copied from their README that was not updated... Will change.

manager/controllers/app/blueprint_controller.go
shlomitk1: Wait will be used in the next PR. 
My understanding is that "wait" does not block the function, but rather means that the release status becomes Deployed only after the resources are ready.

pkg/helm/helm.go
shlomitk1: yes

pkg/helm/helm.go
shlomitk1: runtime object is always convertible. In any case we check for an error

shlomitk1: We report an error if it is an invalid run-time object, if it is not.

pkg/helm/helm_test.go
shlomitk1: Yes. "wait" is true, and we wait for the release to be deployed before checking the resources.

pkg/helm/helm.go
roee88: Don't ignore errors :)

pkg/helm/helm.go
roee88: I understand that, but getting an error is not desired here. My question is about the fact that the interface in `res` returns an Object and you assume it's always `unstructured.Unstructured`. It might be true, but that's an implementation detail that can change. Unless there is some magic allowing Object to be convertible to unstructured.Unstructured via plain cast and I'm not aware of it.

shlomitk1: I rely on the KubeClient.Build interface that returns unstructured objects:
`// Build validates for Kubernetes objects and returns unstructured infos.`

pkg/helm/helm.go
shlomitk1: Committed an error check.

roee88 (approved):
If we end up not using waitOption or not using getresources then we will need to drop them. In the future. I didn't see a concrete description that explains where how and why we would use waitOption.

PR #59, opened 2020-10-12
Quick-start Installation script

tomersolomon1:
This script is for installing the m4d core components (manager, secret-provider) as well as installing the 3rd party open-source components, and also installs the connectors for these 3rd party components.

For installing on Kind, run the following from the top-level directory:
```
kubectl config set-context --current --namespace=m4d-system
./hack/tools/install.sh
```

For installing on OpenShift add the `WITHOUT_OPENSHIFT` environment variable:
```
kubectl config set-context --current --namespace=m4d-system
WITHOUT_OPENSHIFT=false ./hack/tools/install.sh

```

##### EDIT
Following @roee88's request, I've moved the script to be in:
```
hack/install.sh
```


hack/tools/install.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
```
roee88, line 1: Move hack/tools/install.sh to hack/install.sh

hack/tools/install.sh
```diff
@@ -0,0 +1,43 @@
...
+
+set -e
+
+: ${KUBE_NAMESPACE:=m4d-system}
+
+source secret-provider/deploy/vault-util.sh
+
+kubectl create ns $KUBE_NAMESPACE || true
+
+kubectl config set-context --current --namespace=$KUBE_NAMESPACE
```
roee88, line 11: What relies on m4d-system to be the current namespace?

hack/tools/install.sh
```diff
@@ -0,0 +1,43 @@
...
+make -C third_party/opa deploy
+
+# Waiting for the vault deployment to become ready
+# We're using old-school while b/c we can't wait on object that haven't been created, and we can't know for sure that the statefulset had been created so far
+# See https://github.com/kubernetes/kubernetes/issues/75227
+while [[ $(kubectl get -n $KUBE_NAMESPACE pods -l statefulset.kubernetes.io/pod-name=vault-0 -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]];
+do
+    echo "waiting for vault pod to become ready" 
+    sleep 5
+done
```
roee88, line 29: should this move to third_party/vault and be added as a `wait` target in third_party/vault/Makefile?

hack/tools/install.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# Perform a port-forward to communicate with Vault
+port_forward
+
+# Install the manager, the connectors and the secret-provider
+# This also configures the "external" endpoint for mimicking an external Vault installation
+WITHOUT_PORT_FORWARD=true WITHOUT_VAULT=false make deploy
+
+# Configure the secret-end point in vault
```
roee88, line 38: what do you mean "the secret-end point"?

hack/tools/install.sh
```diff
@@ -0,0 +1,43 @@
...
+export AUTH_METHOD=K8S
+WITHOUT_PORT_FORWARD=true make -C secret-provider configure-vault
```
roee88, line 40: Shouldn't this be part of the secret provider deployment?

hack/tools/install.sh
tomersolomon1: The comment refers to the 2 next lines:
```
export AUTH_METHOD=K8S
WITHOUT_PORT_FORWARD=true make -C secret-provider configure-vault
```

hack/tools/install.sh
tomersolomon1: Ok will move `AUTH_METHOD=K8S` overthere

hack/tools/install.sh
tomersolomon1: Sure

hack/tools/install.sh
tomersolomon1: Ok

hack/tools/install.sh
tomersolomon1: You could in theory install to a different ns if you want. The only thing the currently "hard-coded" (but I guess could be modified using helm) is as far as I know in the vault YAML:

roee88: I didn't mean to ask if we can install to a different namespace (we can't yet). I asked why does it need to be the current namespace. Asking because @hunchback took the approach of always specifying the target namespace (`--namespace`/`-n` and the KUBE_NAMESPACE variable)

tomersolomon1: It seems the installation of some of the components assume that we've set `m4d-system` to be the current ns. As we discussed, I'll remove that line from the script and put it in the README.

hack/install.sh
```diff
@@ -0,0 +1,35 @@
...
+
+kubectl apply -f manager/config/prod/deployment_configmap.yaml -n $KUBE_NAMESPACE
+
+make cluster-prepare
+
+# Install third party components
+make -C third_party/vault deploy
+make -C third_party/egeria deploy
+make -C third_party/opa deploy
+
```
ronenkat, line 20: @tomersolomon1 can we add a flag to enable the deployment of these components? this would allow easier deployment of different connectors. And have the option to override the flags with settings that we define before running the script?

tomersolomon1: I can add a flags like `WITHOUT_X`, so this code snippet would become:

ronenkat: @tomersolomon1 any way to make it a positive statement? i.e., WITHOUT ==> With ?
if so, I am good.

froesef (approved):
I'd propose that we also include building the containers in the install script as an optional step. But we can do that in a later PR as well. Otherwise it works fine on my local kind.

secret-provider/deploy/deploy.sh
```diff
@@ -33,10 +34,17 @@ deploy() {
 }
 
 case "$1" in
-undeploy)
+  deploy)
+    deploy
```
froesef, line 38: I wasn't sure where to put this comment. But while working on some other local testing I noticed that the secret-provider has its own service account. When using your own images this service account does not have the imagePullSecret added. 
So we should either add that here or in `registry/deploy.sh`

```
kubectl patch serviceaccount secret-provider -p \
            "{\"imagePullSecrets\": [{\"name\": \"$NAME\"}]}" \
            -n $namespace
```

hack/install.sh
tomersolomon1: Currently in the code-base the convention is to use WITHOUT_X. We could use here WITH_X, but then the script won't be consistent with the rest of the script.

secret-provider/deploy/deploy.sh
tomersolomon1: I'm not sure I follow, why does it need `imagePullSecret`? I thought now the image comes from a public registry

hack/install.sh
roee88: +1 for the suggestion by @tomersolomon1

roee88:
> I'd propose that we also include building the containers in the install script as an optional step. But we can do that in a later PR as well. Otherwise it works fine on my local kind.

The followup should be installing without scripts (like, just apply YAMLs). Let's not mix install with the development cycle :)

roee88 (approved):

hack/install.sh
ronenkat: I would like it changed, the positive way is much better. please handle through an independent issue.

PR #60, opened 2020-10-12
Add a module to copy from S3 to S3

froesef:


roee88:
I propose to hold this off and work together on a single module for implicit copy instead.

ronenkat:
What is not S3 generic in this module? can we make it S3? and then support multiple providers?

modules/m4d-s3-to-s3/templates/batchtransfer.yaml
```diff
@@ -26,13 +26,28 @@ spec:
...
+  {{ if eq .id "removed-ID" }}
+  - action: "RemoveColumns"
+    description: "redacting column: {{ .args.column_name }}"
+    columns: [ "{{ .args.column_name }}" ]
+  {{ end }}
+  {{ end }}
+  {{ if .Values.secretProviderURL }}
+  secretProviderURL: {{ .Values.secretProviderURL | quote }}
+  {{ end }}
   secretProviderRole: demo
```
roee88: We have made changed such that `credentialLocation` (e.g., `.source.credentialLocation`) is used instead to point to the full URL to obtain the credentials from. It includes the role as well.

roee88 (approved):

PR #61, opened 2020-10-13
Extended integration tests

froesef:
This PR introduces an approach of integration testing that does the following:
1. Deploy a local kind cluster
2. Build docker images and push to local registry
3. Deploy manager and test services to local kind cluster
4. Deploy helm charts to local registry (and test them with sample values against deployed cluster (thus also tests webhooks of CRDs etc))
5. Run tests that are able to run against an existing cluster (thus tests should just create CRDs and observe if the resulting change in CRD status and possible creation of other K8s objects satisfies the given test)

In order to serve the helm charts in K8s from a local registry the registry has to support TLS. Thus the deployment has been moved to a TLS registry with dummy credentials. Due to a regression bug in Docker this might currently not work for current docker versions(I have issues on OS X) but the bug is fixed in current versions of the compose-cli library and will be fixed in future Docker releases.

The proposal would also be to remove the publish stages afterwards and basically have one stage that builds the images, tests them and if all the integration tests work retag the images to the public URLs and push them (master only action)

I did not remove any other travis stages yet.


roee88:
Big PR, it will take me time to review it. 

Reading your description, did you mean dummy certificates? We don't need password to login right?

Also, can you share reference to the docker bug you referred to? I didn't understand what doesn't work exactly and which environments it affects.

.mk/cluster.mk
```diff
@@ -35,3 +35,8 @@ kind: kind-cleanup kind-setup istio-setup istio-status
 
 .PHONY: minikube
 minikube: minikube-cleanup minikube-setup istio-setup istio-status
+
+.PHONY: install-cert-local
+install-cert-local:
+	sudo cp hack/tools/registry/ca.crt /usr/local/share/ca-certificates
+	sudo update-ca-certificates
```
shlomitk1, line 42: Is there a way to update certificates without having sudo permissions on a machine?

manager/config/integration-tests/deployment_configmap.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+  #internal vault related variables
+  VAULT_ADDRESS: "http://vault.m4d-system.svc.cluster.local:8200/"
+  VAULT_DATASET_HOME: "m4d/dataset-creds/"
+  VAULT_USER_HOME: "m4d/user-creds/"
+  VAULT_DATASET_MOUNT: "v1/sys/mounts/m4d/dataset-creds"
+  VAULT_USER_MOUNT: "v1/sys/mounts/m4d/user-creds"
+  VAULT_TTL: "24h"
+  VAULT_AUTH: "kubernetes"
+  RUN_WITHOUT_VAULT: "1"
```
shlomitk1, line 34: RUN_WITHOUT_VAULT is not needed if the vault is installed as well. Otherwise, the credentials won't be stored, and the copy will fail.

manager/testdata/e2e/module-implicit-copy-kafka-to-s3-stream.yaml
```diff
@@ -82,5 +82,5 @@ spec:
     - id: "encrypt-ID"
       name: "encrypt"
       level: 2 # column
-  chart: ghcr.io/the-mesh-for-data/m4d-kafka:0.1.0
+  chart: kind-registry:5000/m4d-system/m4d-kafka:0.1.0
```
shlomitk1, line 85: A read module chart is not pushed to kind-registry. Will it be a problem to access ghcr.io for this chart?

modules/m4d-kafka/Makefile
```diff
@@ -4,6 +4,7 @@ all: helm-all
 ROOT_DIR := ../..
 DOCKER_NAME := m4d-kafka
 DOCKER_TAGNAME := 0.1.0
+HELM_VALUES := --values values.yaml.sample
```
shlomitk1, line 7: Why is this needed? How can we test the values used by the blueprint controller to install this chart?

froesef:
> Reading your description, did you mean dummy certificates? We don't need password to login right?

Yes we're creating self signed certs for testing. The username and password are `user` and `pwd` as those are dummy values just used for a local registry. 

> Also, can you share reference to the docker bug you referred to? I didn't understand what doesn't work exactly and which environments it affects.

This is the error message I'm getting with docker. The issue references another PR that is already merged.
https://github.com/docker/compose-cli/issues/712

.mk/cluster.mk
froesef: I did not find any on ubuntu. OS X seems to have such a utility.

manager/testdata/e2e/module-implicit-copy-kafka-to-s3-stream.yaml
froesef: I think for the read module it's fine to pull it from ghcr.io as it's coming from a different repo anyways. For the modules that are maintained in this repository it's important to test the local version here so we test possible changes.

modules/m4d-kafka/Makefile
froesef: How I understood this is that @hunchback uses these sample files for testing of the helm charts. Thus not in combination with blueprints.

roee88:
> Yes we're creating self signed certs for testing. The username and password are user and pwd as those are dummy values just used for a local registry.

user and pwd for what?

.mk/tools.mk
```diff
@@ -32,7 +32,7 @@ $(TOOLBIN)/hugo:
 
 INSTALL_TOOLS += $(TOOLBIN)/golangci-lint
 $(TOOLBIN)/golangci-lint:
-	GOBIN=$(ABSTOOLBIN) GO111MODULE=on go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.19.1
+	GOBIN=$(ABSTOOLBIN) GO111MODULE=on go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.23.0
```
roee88, line 35: to be handled by https://github.com/IBM/the-mesh-for-data/pull/72

.travis.yml
```diff
@@ -79,13 +82,28 @@ jobs:
...
   - stage: Test
-    name: "Integration tests for helm"
+    name: "Integration tests for manager in kind"
     env:
-    - DOCKER_HOSTNAME=localhost:5000
-    - DOCKER_NAMESPACE=m4d
+    - DOCKER_HOSTNAME=kind-registry:5000
+    - DOCKER_NAMESPACE=m4d-system
+    - DOCKER_USERNAME=user # Hardcoded for kind testing
+    - DOCKER_PASSWORD=pwd  # Hardcoded for kind testing
```
roee88, line 90: I thought that we omitted those

Makefile
```diff
@@ -21,12 +21,21 @@ e2e:
 	# $(MAKE) -C pkg/helm test
 	$(MAKE) -C manager e2e
 
+# Splitting cluster-prepare into deploy and deploy-wait
+# Allows to differentiate between YAMLs to be deployed
+# and waiting for the state to be ready
+
```
roee88, line 27: No need for the comment

.travis.yml
```diff
@@ -79,13 +82,28 @@ jobs:
...
+    - make docker-minimal-it
+    #- make -C test/services docker Already build in docker-minimal-it
+    - make cluster-prepare-wait || travis_terminate 1;
+    - make -C secret-provider configure-vault
+    - make -C secret-provider deploy || travis_terminate 1;
+    - make -C manager deploy_it || travis_terminate 1;
+    - make install-cert-local || travis_terminate 1;
+    - make -C manager wait_for_manager || travis_terminate 1;
+    - make helm || travis_terminate 1;
+    - make -C manager run-integration-tests || travis_terminate 1;
```
roee88, line 106: Add a comment to explain what it does

Makefile
```diff
@@ -79,6 +96,30 @@ docker-build-local:
...
+docker-retag-images:
+	docker tag ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/manager:latest ghcr.io/the-mesh-for-data/manager:latest
+	docker tag ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/secret-provider:latest ghcr.io/the-mesh-for-data/secret-provider:latest
+	docker tag ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/egr-connector:latest ghcr.io/the-mesh-for-data/egr-connector:latest
+	docker tag ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/movement-controller:latest ghcr.io/the-mesh-for-data/movement-controller:latest
+	docker tag ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/dummy-mover:latest ghcr.io/the-mesh-for-data/dummy-mover:latest
+	docker tag ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/opa-connector:latest ghcr.io/the-mesh-for-data/opa-connector:latest
+	docker tag ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/vault-connector:latest ghcr.io/the-mesh-for-data/vault-connector:latest
+	docker tag ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/serverpolicycompiler-mock:latest ghcr.io/the-mesh-for-data/serverpolicycompiler-mock:latest
+	docker tag ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/data-catalog-mock:latest ghcr.io/the-mesh-for-data/data-catalog-mock:latest
```
roee88, line 109: Is there any way we don't hard code things?

Makefile
```diff
@@ -79,6 +96,30 @@ docker-build-local:
...
+docker-push-public:
+	docker push ghcr.io/the-mesh-for-data/manager:latest
+	docker push ghcr.io/the-mesh-for-data/secret-provider:latest
+	docker push ghcr.io/the-mesh-for-data/egr-connector:latest
+	docker push ghcr.io/the-mesh-for-data/movement-controller:latest
+	docker push ghcr.io/the-mesh-for-data/dummy-mover:latest
+	docker push ghcr.io/the-mesh-for-data/opa-connector:latest
+	docker push ghcr.io/the-mesh-for-data/vault-connector:latest
+	docker push ghcr.io/the-mesh-for-data/serverpolicycompiler-mock:latest
+	docker push ghcr.io/the-mesh-for-data/data-catalog-mock:latest
```
roee88, line 121: Why not invoke `docker-push` target? something like `DOCKER_HOSTNAME=ghcr.io DOCKER_NAMESPACE=the-mesh-for-data DOCKER_TAG=latest make docker-push`

hack/tools/create_kind.sh
```diff
@@ -38,18 +56,22 @@ kind_create() {
...
 case "$op" in
     cleanup)
         header_text "Uninstalling kind cluster"
         registry_delete || true
+        certs_delete || true
         kind_delete || true
         ;;
     *)
         header_text "Installing kind cluster"
+        certs_create
```
roee88, line 74: Doesn't it make sense to move the certs install to here as well?

hack/tools/kind-config.yaml
```diff
@@ -9,4 +9,8 @@ nodes:
 containerdConfigPatches:
 - |-
   [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:5000"]
-    endpoint = ["http://kind-registry:5000"]
+    endpoint = ["https://kind-registry:5000"]
+  [plugin."io.containerd.grpc.v1.cri".registry.configs."kind-registry".tls]
+    insecure_skip_verify = true
+  [plugin."io.containerd.grpc.v1.cri".registry.configs."kind-registry:5000".tls]
+    insecure_skip_verify = true
```
roee88, line 16: when do we use insecure

manager/Makefile
```diff
@@ -27,20 +28,33 @@ docker-build: generate
...
 	$(TOOLBIN)/kustomize build --load_restrictor none config/movement-controller | $(TOOLBIN)/kubectl delete -f -
 
-.PHONY: motion-integration-tests
-motion-integration-tests: #deploy_mc
-	IMG=${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/movement-controller:${DOCKER_TAGNAME} $(MAKE) deploy_mc
-	$(TOOLBIN)/kubectl wait --for=condition=available -n m4d-system deployment/m4dcontroller-manager --timeout=120s
-	NO_SIMULATED_PROGRESS=true USE_EXISTING_CONTROLLER=true USE_EXISTING_CLUSTER=true go test ./... -v -run TestBatchTransferAPIs -count 1
+deploy_it: docker-secret
+	$(TOOLBIN)/kubectl create namespace ${KUBE_NAMESPACE} || true
+	cd config/manager && $(ABSTOOLBIN)/kustomize edit set image controller=${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/manager:${DOCKER_TAGNAME}
```
roee88, line 41: `config/manager` or `config/integration-tests`?

manager/config/integration-tests/datacatalog.yaml
```diff
@@ -0,0 +1,41 @@
+# Copyright 2020 IBM Corp.
```
roee88, line 1: Why are these in `config/integrations-tests` and not part of the mocks themselves?

manager/config/movement-controller/kustomization.yaml
```diff
@@ -9,7 +9,7 @@ namespace: m4d-system
 # "wordpress" becomes "alices-wordpress".
 # Note that it should also match with the prefix (text before '-') of the namespace
 # field above.
-namePrefix: m4d
+namePrefix: m4d-
```
roee88, line 12: why the change?

roee88 (commented):
The overall approach looks good. The main problem I have is that the makefiles are messy already and this PR adds on top of it. I guess it's fine to do a cleanup of the makefiles after this PR.

manager/config/movement-controller/kustomization.yaml
froesef: It was an old inconsistency that I fixed. This is the kustomize folder that I use for the motion only testing.

manager/config/integration-tests/datacatalog.yaml
froesef: I think we should keep the integration tests yamls together and not spread over so they don't add confusion. Otherwise the setup for integration tests gets even more complicated having to install 5 yaml files from 5 different directories.

.mk/cluster.mk
hunchback: having `root` access on the dev machine is a reasonable requirement, I suggest not spending time to bypass this, @shlomitk1  I would suggest using a local dev environment (on your laptop) rather than a shared account on an IT provisioned VM

manager/testdata/e2e/module-implicit-copy-kafka-to-s3-stream.yaml
hunchback: for development purposes, I really want to ability to run my development helm chart (from local repo), but I also understand the need to have the sample work w/o a local repo ... so the only way this can work is using templating, if we wrap our samples (all together) as a helm package then (replacing helm repo) would be easy.

modules/m4d-kafka/Makefile
hunchback: correct, these values are used for testing purposes

hunchback:
> > Reading your description, did you mean dummy certificates? We don't need password to login right?
> 
> Yes we're creating self signed certs for testing. The username and password are `user` and `pwd` as those are dummy values just used for a local registry.
> 
> > Also, can you share reference to the docker bug you referred to? I didn't understand what doesn't work exactly and which environments it affects.
> 
> This is the error message I'm getting with docker. The issue references another PR that is already merged.
> [docker/compose-cli#712](https://github.com/docker/compose-cli/issues/712)

I suggest we handle docker installation ourselves, this will enable us using an older version of docker which does not have this bug (https://github.com/docker/compose-cli/issues/712#issuecomment-700812166)

Makefile
hunchback: can u use a makefile macro to reduce duplication?

hunchback (changes requested):
several issues:

1. i would like to be able to test the change on my dev machine so for that i would like you to create a top-level makefile target and thus call this target from travis

2. a secondary comment is that i rather you make more use of macros and defines so as to use code duplication

roee88:
> > > Reading your description, did you mean dummy certificates? We don't need password to login right?
> > 
> > 
> > Yes we're creating self signed certs for testing. The username and password are `user` and `pwd` as those are dummy values just used for a local registry.
> > > Also, can you share reference to the docker bug you referred to? I didn't understand what doesn't work exactly and which environments it affects.
> > 
> > 
> > This is the error message I'm getting with docker. The issue references another PR that is already merged.
> > [docker/compose-cli#712](https://github.com/docker/compose-cli/issues/712)
> 
> I suggest we handle docker installation ourselves, this will enable us using an older version of docker which does not have this bug ([docker/compose-cli#712 (comment)](https://github.com/docker/compose-cli/issues/712#issuecomment-700812166))

I see bo reason to do that. The docker service in Travis does work. Managing docker installation for users is a headache and doesn't really worth the effort.

hunchback (approved):
I notice issues which will not enable the test to work as-is such as its' dependence on kind-registry hostname and possibly other stuff, as its' passing in the CI and this is a new capability I am approving but I would request a follow-up PRs handling:
1. working fully on a local VM
2. refactoring and simplification of code

PR #62, opened 2020-10-13
Gui documentation changes and fixes

shlomitk1:
Small changes and adjustments:
- not passing 'credentials' property name in the 'data' body
- ownerID -> ownerId
- deploying GUI in default namespace
- modified instructions to deploy GUI


samples/gui/Deployment.yaml
```diff
@@ -56,12 +56,12 @@ spec:
     spec:
       containers:
       - name: datauserclient
-        image: image-registry.openshift-image-registry.svc:5000/datauser1/datauserclient:latest
+        image: image-registry.openshift-image-registry.svc:5000/default/datauserclient:latest
         imagePullPolicy: Always
```
simanadler, line 60: For integration with Watson Studio it appears that we will need to deploy to namespace cp4d, which is where the user's compute runs.  If we have a deployment script, this should be a parameter of the deployment.

samples/gui/Deployment.yaml
```diff
@@ -56,12 +56,12 @@ spec:
...
         ports:
         - containerPort: 3000
       - name: datauserserver
-        image: image-registry.openshift-image-registry.svc:5000/datauser1/datauserserver:latest
+        image: image-registry.openshift-image-registry.svc:5000/default/datauserserver:latest
```
simanadler, line 64: Same as previous comment.

samples/gui/deploy.sh
```diff
@@ -6,12 +6,12 @@ export VAULT_TOKEN=$(kubectl get secrets vault-unseal-keys -n m4d-system -o json
...
-kubectl create secret generic vault-unseal-keys --from-file=vault-root=./token.txt || true
+kubectl delete secret vault-unseal-keys -n default
+kubectl create secret generic vault-unseal-keys --from-file=vault-root=./token.txt -n default || true
 kubectl delete service datauserserver || true
 kubectl delete service datauserclient || true
-kubectl delete deployment gui
-kubectl apply -f Deployment.yaml
+kubectl delete deployment gui -n default
+kubectl apply -f Deployment.yaml -n default
 rm ./token.txt
```
simanadler, line 15: See previous comments.  Script should receive as a parameter the namespace in which to deploy the GUI.

simanadler (changes requested):

samples/gui/deploy.sh
shlomitk1: Done. Relies on $KUBE_NAMESPACE that should be set prior to running deploy.sh

samples/gui/README.md
```diff
@@ -76,7 +76,8 @@ docker build . -t <registry>/<namespace>/datauserclient:latest
 ## Deployment
   ```
 cd <root>>/samples/gui
-  
+
+export KUBE_NAMESPACE=<namespace>
```
simanadler, line 80: Please add some documentation describing what this is.  I.e. The namespace in which the workload is running and point out that the GUI must run in that namespace.  Also, I would suggest perhaps calling the env variable WORKLOAD_NAMESPACE.

samples/gui/Deployment.yaml
shlomitk1: Done.

samples/gui/README.md
shlomitk1: Changed the name, added documentation.

simanadler (approved):

Issue #63, opened 2020-10-14
Blueprints do not have finalizers


shlomitk1:
Problem: blueprints do not have finalizers. As a result, it is possible that deletion of a blueprint will not result in the deletion of its owned resources (Helm charts).
Proposed solution: add a finalizer to the blueprint and delete the allocated helm releases in the process of finalizer reconciliation.
Affected components: manager

PR #73, opened 2020-10-19
Add finalizers to blueprint

shlomitk1:
This PR adds a finalizer to Blueprint. Deletion of the blueprint results in deleting the associated helm releases.
Fixes https://github.com/IBM/the-mesh-for-data/issues/63


manager/controllers/app/blueprint_controller.go
```diff
@@ -51,41 +52,87 @@ func (r *BlueprintReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 
-func (r *BlueprintReconciler) deleteChartResource(ctx context.Context, kubeNamespace string, releaseName string) (ctrl.Result, error) {
-	rel, err := r.Helmer.Status(kubeNamespace, releaseName)
-	if err == nil && rel != nil {
-		_, _ = r.Helmer.Uninstall(kubeNamespace, releaseName)
+// reconcileFinalizers reconciles finalizers for Blueprint
+func (r *BlueprintReconciler) reconcileFinalizers(blueprint *app.Blueprint) error {
+	// finalizer
+	finalizerName := r.Name + ".finalizer"
+	hasFinalizer := utils.ContainsFinalizer(blueprint.GetFinalizers(), finalizerName)
```
roee88, line 89: I believe that we can use `ctrlutil.ContainsFinalizer` now (we use v0.6.2 of controller-runtime)

https://github.com/kubernetes-sigs/controller-runtime/blob/v0.6.2/pkg/controller/controllerutil/controllerutil.go#L290

manager/controllers/app/blueprint_controller.go
```diff
@@ -51,41 +52,87 @@ func (r *BlueprintReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-func (r *BlueprintReconciler) applyChartResource(ctx context.Context, log logr.Logger, ref string, vals map[string]interface{}, kubeNamespace string, releaseName string) (ctrl.Result, error) {
+func (r *BlueprintReconciler) deleteExternalResources(blueprint *app.Blueprint) error {
+	for _, step := range blueprint.Spec.Flow.Steps {
+		releaseName := getReleaseName(step)
+		rel, err := r.Helmer.Status(blueprint.Namespace, releaseName)
+		if err == nil && rel != nil {
+			_, _ = r.Helmer.Uninstall(blueprint.Namespace, releaseName)
+		}
+	}
+	return nil
```
roee88, line 132: This seems like a best effort to delete, ignoring errors. Either:
1. Declare that this is a best effort and then you can avoid the call to Status (counting on the removal of the namespace to do the trick in most cases)
2. Track errors and if you encountered one then on the end of the loop don't return nil but rather an error (better IMHO)

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -108,7 +108,7 @@ func checkBlueprintStatus(applicationContext *app.M4DApplication, blueprint *app
 func (r *M4DApplicationReconciler) reconcileFinalizers(applicationContext *app.M4DApplication) error {
 	// finalizer (Blueprint)
 	finalizerName := r.Name + ".finalizer"
-	hasFinalizer := ContainsFinalizer(applicationContext, finalizerName)
+	hasFinalizer := utils.ContainsFinalizer(applicationContext.GetFinalizers(), finalizerName)
```
roee88, line 111: same comment as above about `ContainsFinalizer`

manager/controllers/utils/utils.go
```diff
@@ -216,3 +216,14 @@ func SupportsFlow(array []app.ModuleFlow, element app.ModuleFlow) bool {
...
+// ContainsFinalizer returns true if the given finalizer string appears in the list of finalizers of the object
+// Should be implemented in controller-runtime package - TO FIX
+func ContainsFinalizer(finalizers []string, finalizer string) bool {
+	for _, e := range finalizers {
+		if e == finalizer {
+			return true
+		}
+	}
+	return false
+}
```
roee88, line 229: Same comment as above about `ContainsFinalizer`

manager/controllers/app/blueprint_controller.go
```diff
@@ -51,41 +52,87 @@ func (r *BlueprintReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+			return err
+		}
+	}
+	return nil
+}
 
-	return ctrl.Result{}, nil
+func getReleaseName(step app.FlowStep) string {
+	// we add the "r" character at the beginning of the release name, since it must begin with an alphabetic character
+	return "r" + utils.Hash(step.Name, 20)
```
roee88, line 121: unrelated to this PR but I think that we should start moving common code like `Hash` to appropriate folders in `pkg`

roee88 (commented):
Looks good, minor comments.

manager/controllers/app/blueprint_controller.go
shlomitk1: Done.

manager/controllers/app/blueprint_controller.go
shlomitk1: Done.

manager/controllers/app/m4dapplication_controller.go
shlomitk1: Done.

manager/controllers/utils/utils.go
shlomitk1: Done.

manager/controllers/app/blueprint_controller.go
```diff
@@ -51,41 +52,88 @@ func (r *BlueprintReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+func (r *BlueprintReconciler) deleteExternalResources(blueprint *app.Blueprint) error {
+	var err error
+	for _, step := range blueprint.Spec.Flow.Steps {
+		releaseName := getReleaseName(step)
+		if rel, errStatus := r.Helmer.Status(blueprint.Namespace, releaseName); errStatus != nil || rel == nil {
+			continue
+		}
+		_, err = r.Helmer.Uninstall(blueprint.Namespace, releaseName)
+	}
+	return err
```
roee88, line 133: 1. Should we log errStatus for debugging?
2. This implementation simply returns what the last call to uninstall returns, which might be nil even if a previous uninstall call returned an error. You can either fix this directly (e.g., to only assign to the err var on the first actual error), or consider to group all returned errors (see implementation in https://github.com/helm/helm/blob/master/pkg/action/uninstall.go or the https://github.com/hashicorp/go-multierror library).

shlomitk1: Fixed.

roee88 (approved):

Issue #74, opened 2020-10-21
generating API docs fails on OSX


roee88:
The generation of api docs fails because of differences between the GNU and BSD versions of sed.

reported by @ronenkat

PR #75, opened 2020-10-21
fix: generation of API docs fails in osx

roee88:
Fixes https://github.com/IBM/the-mesh-for-data/issues/74

Replaced the use of `sed` with a bash script


ronenkat (approved):
Fixed the issue. Verified on Mac

PR #77, opened 2020-10-21
Define configmap for GUI

shlomitk1:
This PR defines a new configmap for GUI which includes vault definitions and a geography to be easily configured during GUI deployment.


roee88:
Just as a note, there is a good chance that none of these configurations will be needed in the near future.

Geography should be removed from `M4DApplication` and from here, and be defined in the control plane configuration instead. The geography indicates where the workload is and should be taken from a reliable source (c.f. indication of where we want to run - M4DApplication don't deploy the workload). This should eventually be discovered by the platform, but for now we can set it as a control plane parameter set by the administrator, and we will extend this when we move to multi cluster.

An issue about system credentials provided by the users will be opened soon.

PR #78, opened 2020-10-22
fixing the installation script

tomersolomon1:
Fixed the install.sh script to use correctly the `WITHOUT_X` environment variables. Also fixed Vault's Makefile (it's a miracle it worked before, probably b/c we didn't need to wait for Vault b/c Egeria's deployment takes very long time).

Also fixed the number of Egeria connector instances to be 1 (was 3 before for no good reason).

Also fixed Egeria's installation script - didn't work properly on OpenShift.
Previously, due to a bug, the script was relying on the existence of the `egeria-catalog` in order to work properly.

### DIFF
--- a/hack/install.sh
+++ b/hack/install.sh
@@ -4,9 +4,9 @@ set -e
 
 : ${KUBE_NAMESPACE:=m4d-system}
 : ${PORT_TO_FORWARD:=8200}
-: ${WITHOUT_VAULT=false}
-: ${WITHOUT_EGERIA=false}
-: ${WITHOUT_OPA=false}
+: ${WITHOUT_VAULT:=false}
+: ${WITHOUT_EGERIA:=false}
+: ${WITHOUT_OPA:=false}
 
 source third_party/vault/vault-util.sh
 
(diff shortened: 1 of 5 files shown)


hack/install.sh
```diff
@@ -17,9 +17,9 @@ kubectl apply -f manager/config/prod/deployment_configmap.yaml -n $KUBE_NAMESPAC
 make cluster-prepare
 
 # Install third party components
-WITHOUT_VAULT || make -C third_party/vault deploy
-WITHOUT_EGERIA || make -C third_party/egeria deploy
-WITHOUT_OPA || make -C third_party/opa deploy
+$WITHOUT_VAULT || make -C third_party/vault deploy
```
ronenkat, line 20: @tomersolomon1 you are running "make cluster-prepare" which runs "$(MAKE) -C third_party/vault deploy"
See line 28 at Makefile - https://github.com/IBM/the-mesh-for-data/blob/master/Makefile#L28
there is a duplication, one that always install Vault, the other on condition.

tomersolomon1: Suggestion - remove Vault installation from the `cluster-prepare` phase, so it's installed (only) by the installation script - and keep it optional.
Anyhow, if we're talking about the `cluster-prepare` phase, why we need it to create `registry` secret?
https://github.com/IBM/the-mesh-for-data/blob/master/Makefile#L27
https://github.com/IBM/the-mesh-for-data/blob/master/third_party/registry/deploy.sh#L18

Is it possible this is a leftover from the time before moving to public docker registry?

roee88 (approved):

ronenkat (approved):

hack/install.sh
roee88: Let's open a new issue for these

tomersolomon1: See https://github.com/IBM/the-mesh-for-data/issues/83

PR #81, opened 2020-10-25
add AssetID to read arguments inside a blueprint

shlomitk1:
This PR adds a new field **assetID**  inside read module arguments in a blueprint.
This argument identifies an asset that can be used  by a read module for creating data access instructions 
as **"{{ .Values.assetID}}"**

```
    - arguments:
        flow:  read
        read:
          - assetID: {"asset_id": "xxx", "catalog_id": "s3"}
            source:
              connection:
                name:  cos
                s3:
                  bucket:           m4d-test-bucket
                  endpoint:         s3.eu-gb.cloud-object-storage.appdomain.cloud
                  object_key:       small.parq
               ...
      name:            notebook-arrow-flight-module-{"asset_id": "xxx", "catalog_id": "s3"}
      template:        arrow-flight-module

```

### DIFF
--- a/manager/apis/app/v1alpha1/blueprint_types.go
+++ b/manager/apis/app/v1alpha1/blueprint_types.go
@@ -31,6 +31,11 @@ type ReadModuleArgs struct {
 	// +required
 	Source DataStore `json:"source"`
 
+	// AssetName represents the asset name to be used for accessing the data when it is ready
+	// It is copied from the m4dapplication
+	// +required
+	AssetName string `json:"assetName"`
+
 	// Transformations are different types of processing that may be done to the data
 	// +optional
 	Transformations []pb.EnforcementAction `json:"transformations,omitempty"`
--- a/manager/apis/app/v1alpha1/blueprint_types.go
+++ b/manager/apis/app/v1alpha1/blueprint_types.go
@@ -32,7 +32,7 @@ type ReadModuleArgs struct {
 	Source DataStore `json:"source"`
 
 	// AssetName represents the asset name to be used for accessing the data when it is ready
-	// It is copied from the m4dapplication resource
+	// It is copied from the M4DApplication resource
 	// +required
 	AssetName string `json:"assetName"`
 
--- a/manager/apis/app/v1alpha1/blueprint_types.go
+++ b/manager/apis/app/v1alpha1/blueprint_types.go
@@ -32,7 +32,7 @@ type ReadModuleArgs struct {
 	Source DataStore `json:"source"`
 
 	// AssetName represents the asset name to be used for accessing the data when it is ready
-	// It is copied from the m4dapplication
+	// It is copied from the m4dapplication resource
 	// +required
 	AssetName string `json:"assetName"`
 
--- a/manager/apis/app/v1alpha1/blueprint_types.go
+++ b/manager/apis/app/v1alpha1/blueprint_types.go
@@ -31,10 +31,10 @@ type ReadModuleArgs struct {
 	// +required
 	Source DataStore `json:"source"`
 
-	// AssetName represents the asset name to be used for accessing the data when it is ready
+	// AssetID identifies the asset to be used for accessing the data when it is ready
 	// It is copied from the M4DApplication resource
 	// +required
-	AssetName string `json:"assetName"`
+	AssetID string `json:"assetID"`
 
 	// Transformations are different types of processing that may be done to the data
 	// +optional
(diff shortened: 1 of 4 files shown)


roee88:
The only thing that confuses me is the text in the PR description. What is asset_id/xxx/catalog_id/s3?

manager/apis/app/v1alpha1/blueprint_types.go
```diff
@@ -31,6 +31,11 @@ type ReadModuleArgs struct {
 	// +required
 	Source DataStore `json:"source"`
 
+	// AssetName represents the asset name to be used for accessing the data when it is ready
+	// It is copied from the m4dapplication
```
roee88, line 35:
```suggestion
	// It is copied from the M4DApplication resource
```

manager/controllers/app/moduleinstance.go
```diff
@@ -183,6 +183,7 @@ func (r *M4DApplicationReconciler) SelectModuleInstancesPerDataset(item modules.
 	readInstructions := make([]app.ReadModuleArgs, 0)
 	readInstructions = append(readInstructions, app.ReadModuleArgs{
 		Source:          readSource,
+		AssetName:       item.AssetID,
```
roee88, line 186: You created AssetID for internal stuff like logs (I still differ in opinion but nevermind). This is external. Can we use the original instead?

manager/apis/app/v1alpha1/blueprint_types.go
shlomitk1: Done.

manager/controllers/app/moduleinstance.go
shlomitk1: Changed for all cases to be copied as-is from the original resource. Does not affect validity of kubernetes resource names or the path to secret-provider.

roee88 (approved):

manager/apis/app/v1alpha1/blueprint_types.go
```diff
@@ -31,6 +31,11 @@ type ReadModuleArgs struct {
 	// +required
 	Source DataStore `json:"source"`
 
+	// AssetName represents the asset name to be used for accessing the data when it is ready
+	// It is copied from the M4DApplication resource
+	// +required
+	AssetName string `json:"assetName"`
```
roee88, line 37: Already approved, but we might want to use `AssetID` to be consistent?

shlomitk1: Changed the name.

manager/apis/app/v1alpha1/blueprint_types.go
```diff
@@ -31,6 +31,11 @@ type ReadModuleArgs struct {
 	// +required
 	Source DataStore `json:"source"`
 
+	// AssetID identifies the asset to be used for accessing the data when it is ready
+	// It is copied from the M4DApplication resource
+	// +required
+	AssetID string `json:"assetID"`
```
simanadler, line 37: Shouldn't this be part of the DataStore struct?

shlomitk1: This is relevant only for the read module that includes this information in data access code. 
Conceptually, DataStore represents the "physical" data endpoint. If read module had a Destination as well as a Source, AssetID  would have been used inside Destination.

simanadler: In the (future) ingest scenario we will have situations where we need to provide the asset ID of both the source and the destination, so I would suggest we put it in the DataStore - at least as an optional field.  Lineage will need this information as well

roee88: @simanadler I'm not sure if this is the right way to specify this for the ingest flow target. Would it be okay to keep this PR as is and revisit after finalizing the ingest flow?

shlomitk1: @simanadler If I understand you correctly, you suggest to connect the connection details with a cataloged asset. If a copy module catalogs the implicit copy, and a read is applied after a copy, it will actually read a newly cataloged asset. However, from the perspective of a person who writes the code that interacts with the read module, it will be more work to copy a new access instructions each time a cataloged version changes. 
I agree that we'll need a new field to connect the data store with the cataloged version but not for the purpose of this PR.

simanadler: OK

PR #86, opened 2020-10-26
Don't set image pull policy by default

froesef:
This PR does not set the imagePullPolicy by default. 
Like this the manager can set the imagePullPolicy when defaulting BatchTransfer and StreamTransfer.

The current policy is IfNotPresent which will reuse images that are already downloaded in a cluster. When updating the image behind the tag (Currently we're still using latest) this will result in still using the old image.

Like this the installation of the-mesh-for-data decides the policy.

### DIFF
--- a/modules/m4d-db2wh/templates/batchtransfer.yaml
+++ b/modules/m4d-db2wh/templates/batchtransfer.yaml
@@ -27,8 +27,10 @@ spec:
   - name: "Remove columns"
     action: "RemoveColumns"
     columns: ["JOB", "SSN"]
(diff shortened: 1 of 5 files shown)


roee88 (approved):

shlomitk1 (approved):
I understand that this PR provides a ground for the further PR that will override the values by manager, but leaves the arguments as-is (default null value for imagePullPolicy is equivalent to the previous default IfNotPresent).

roee88:
From https://kubernetes.io/docs/concepts/containers/images/#updating-images it seems to not be equivalent when latest tag is used.

PR #99, opened 2020-10-29
travis: Reduce CI process

froesef:
In order to test if the retagging and pushing to public images works fine as expected I'd like to remove the condition `condition: $TRAVIS_PULL_REQUEST = false` once it's ready and before it would get merged anyways. This is will overwrite the public images from a branch but I think it's better to work out a fix on master branch if the code does not work. Once it's verified to work I'll put the condition back into place.


hunchback (approved):

roee88:
`deploy` can't run in PRs, regardless of the (redundant) condition.

roee88:
@froesef Please resolve conflicts

PR #113, opened 2020-11-03
os: automate setup of a ubuntu host from scrach

hunchback:
provide a means for setting up an ubuntu host from scrach (works on VM and on WSL), addressing part of https://github.com/IBM/the-mesh-for-data/issues/127

### DIFF
--- a/.mk/verify.mk
+++ b/.mk/verify.mk
@@ -18,13 +18,6 @@ define license_python
 	($(ABSTOOLBIN)/license_finder --python-version=3 || true)
 endef
 
-GO_VERSION:=1.13
-CODE_MAINT += go-version
-.PHONY: go-version
-go-version:
-	@(go version | grep -q 'go$(GO_VERSION)\(\.[0-9]*\)\? ') || \
-	echo 'WARNING: bad go version to fix run: eval "$$(gimme $(GO_VERSION))"'
-
 CODE_MAINT += fmt
 .PHONY: fmt
 fmt:
--- a/hack/tools/install_os.sh
+++ b/hack/tools/install_os.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+set -x
+set -e
+
+ID=$(lsb_release -i)
+ID=${ID##Distributor ID:[[:space:]]}
+RELEASE=$(lsb_release -r)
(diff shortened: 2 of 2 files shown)


.mk/verify.mk
```diff
@@ -18,13 +18,6 @@ define license_python
 	($(ABSTOOLBIN)/license_finder --python-version=3 || true)
 endef
 
-GO_VERSION:=1.13
```
roee88, line 21: Why is this removed?

go.mod
```diff
@@ -3,19 +3,22 @@ module github.com/ibm/the-mesh-for-data
...
 
 require (
+	cloud.google.com/go v0.66.0 // indirect
 	emperror.dev/errors v0.7.0
 	github.com/go-chi/chi v4.1.2+incompatible
 	github.com/go-chi/render v1.0.1
 	github.com/go-logr/logr v0.1.0
 	github.com/gogo/protobuf v1.3.1
-	github.com/golang/protobuf v1.3.5
+	github.com/golang/protobuf v1.4.2
```
roee88, line 12: Incompatible with out project

roee88 (changes requested):
Please open an issue before submitting a PR, especially for things that were not previously discussed in an issue or elsewhere. Also, please follow the contribution instructions for signing off commits.

.mk/verify.mk
hunchback: there are new version of go 1.14 and 1.15, thus no need to limit or bind our project to any specific version (and leave go version and IDE choice to developer).

go.mod
hunchback: reverting ... but sooner or later it will have to be addressed as currently go.mod automatically suggests this change when building the project

roee88:
See my comment on the issue.

Technical comments:
- Have you actually tested the project with go 1.14/1.15? 
- Missing go install
- Missing jdk install
- Missing docker post-install for linux instructions (e.g., creating group)
- Missing other tools such as make and jq
- Why are conntrack, ethtool and apache2-utils needed?

Rather than trying to solve this for "any ubuntu VM" I'm closing this PR and we can discuss in the issue to see if we really want to vendor such files (I don't think so).

Issue #88, opened 2020-10-27
Tavis builds fail because of cert-manager errors


roee88:
Some builds such as https://travis-ci.com/github/IBM/the-mesh-for-data/jobs/408525032 fail:
```
error: unable to read URL "https://github.com/jetstack/cert-manager/releases/download/v0.13.1/cert-manager.yaml", server reported 429 too many requests, status code=429
```

Should either vendor this YAML or add caching to it in Travis.

hunchback:
Other common fails are:

```
Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "webhook.cert-manager.io": Post https://cert-manager-webhook.cert-manager.svc:443/mutate?timeout=30s: dial tcp 10.102.184.81:443: connect: connection refused
Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "webhook.cert-manager.io": Post https://cert-manager-webhook.cert-manager.svc:443/mutate?timeout=30s: dial tcp 10.102.184.81:443: connect: connection refused
../.mk/operator.mk:33: recipe for target 'deploy' failed
```

as can be seen in https://travis-ci.com/github/IBM/the-mesh-for-data/jobs/429420848

PR #114, opened 2020-11-03
travis: wait for slow startup of cert-manager

hunchback:
attempt to fix sporadic failure due to cert-manager not starting up intime

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,9 +26,6 @@ cluster-prepare:
 	$(MAKE) -C third_party/cert-manager deploy
 	$(MAKE) -C third_party/registry deploy
 	$(MAKE) -C third_party/vault deploy
-
-.PHONY: cluster-prepare-wait
-cluster-prepare-wait:
 	$(MAKE) -C third_party/cert-manager deploy-wait
 	$(MAKE) -C third_party/vault deploy-wait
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -26,9 +26,6 @@ cluster-prepare:
 	$(MAKE) -C third_party/cert-manager deploy
 	$(MAKE) -C third_party/registry deploy
 	$(MAKE) -C third_party/vault deploy
-
-.PHONY: cluster-prepare-wait
-cluster-prepare-wait:
```
roee88, line 31: The split (cluster-prepare and cluster-prepare-wait) was done in https://github.com/IBM/the-mesh-for-data/pull/61 by @froesef. It allows installing the integration test stuff while the third parties are being installed.

roee88 (commented):
My understanding is that the problem is not that cert-manager isn't ready but rather that we fail to download cert-manager.yaml. See https://github.com/IBM/the-mesh-for-data/issues/88. I don't understand how the changes in this PR help solve the issue. Can you explain?

roee88:
I see now that this is an attempt to solve another problem that isn't described in an issue. So for completeness this is what it tries to solve:
```
Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "webhook.cert-manager.io": Post https://cert-manager-webhook.cert-manager.svc:443/mutate?timeout=30s: dial tcp 10.98.47.35:443: connect: connection refused
```

hunchback:
addressing part of https://github.com/IBM/the-mesh-for-data/issues/88

hunchback:
this is going to be redundant due to elimination of stages from travis

Issue #82, opened 2020-10-25
ci: add support for releases


roee88:
Problem: Currently we don't do releases at all. This includes: no GitHub releases, no GitHub tags, no tags on containers

Requirements:
- track releases using tags (e.g., release-v0.0.1)
- tag images in CI when running from a tag branch
- optionally create GitHub releases with release notes
- optionally support vertioned docs in website

Personal suggestion for now is to only do a minimal change to the CI to tag images the branch name for tag branches. Then when ready @ronenkat will create a GitHub tag for the first release that will trigger the CI.

roee88:
@hunchback want to take this one?

hunchback:
I  plan to follow a an example from a small sized project such as razee: https://github.com/razee-io/Razee/releases/tag/0.0.9

roee88:
All non optional requirements are implemented (some more PRs might be opened for fixes). @hunchback please open concrete issues for future improvements when needed.

PR #42, opened 2020-10-01
Modules: make image/chart registry configurable

hunchback:

PR #115, opened 2020-11-03
travis: add release support

hunchback:
I first go at auto creating a github release based on the sequence proposed here: https://prof.bagneris.net/posts/2017/github-releases-travis-ci-en.html

for now on creation of a tag I expect github to generate the release which would include:
- README.md
- the-mesh-for-data.src.tag.gz

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -136,3 +136,12 @@ jobs:
 
 notifications:
   email: false
+deploy:
+  provider: releases
+  api_key:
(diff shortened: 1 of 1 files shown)


hunchback:
addressing https://github.com/IBM/the-mesh-for-data/issues/120

.travis.yml
```diff
@@ -136,3 +136,18 @@ jobs:
...
   email: false
+env:
+  global:
+    - RELEASE_NAME=the-mesh-for-data
+before_deploy:
+- git archive --format=tag.gz -o /tmp/$RELEASE_NAME.src.tar.gz --prefix=$RELEASE_NAME/
+deploy:
+  provider: releases
+  api_key:
+    secure: Es3z5bIx3N2en+vn6ogPyftwMoyUYE60s29gfBmThBd0UoQKKltGn5bCo1UnR5NpmNAyDlJtz2yN239IImp2RYwLGVXMrT/lqsTtsrEm8rKEgMWhteBdxWeZhBWMEip0PjZznO0gY9ALAH0IALvYqtVmo93ZtvXXbyYgzXleqNSGE1ChK3u7gFrNeEHKH1ivQ0xKozg3W4kaKAxwnbsi7KSxyb0o/aK8QGPya8Ldfn0l7DhxMuSe9a3l30dnoh0EOTEsH9yosYmhqG+e4erKBSdjKbzQpxzh9xpKHbw998p84OsVrSR8UyRNmzG6IhfISVRUsn73CtpG7imeMhTaezcXQ0Hy1r6WvumoSl7oOFLUz1b3R59bfR1XCYKLcWgK4kek/unzd0x2OJJWdfy/RgCRwnr60O0hD3bwm5w9I2DywyF7ilObGp0M/oo8LiKryawe9lahkrIhRNVdgNEO62q/UmNTOfTH3pm2wXDMAaciphXkbK5ZFfGo5Xzdpx6Cm72ctMc3hcfLw4W5btNOKTXlfLHNdicTSRdFVlosvZMBgYpuoeloj232Kx4HYvbOals7vRM2FhZGWRZ4g2QMB8b6piVD/TRYUUnZeUOP4q+Hi6E2hb902M7LJZz968clMGeC6nLhsOJBHh+4lAfL1RBex85Y5ma60OV1YpvdUnk=
```
ronenkat, line 147: Should we have an api_key her? what is it for?

roee88 (changes requested):
I don't understand what you are releasing. As far as I know, the source code in both zip and tar.gz format is attached to *any* github release already. What we need for the main repo is described in https://github.com/IBM/the-mesh-for-data/issues/82, including the first step which is to create an empty release but to tag docker images appropriately.

.travis.yml
hunchback: this key was created by the `travis` command line its' encrypted so that only travis service (which has the private key of travis) can decrypt and then use the api_key for creating the release ... so there is no security risk here.

hunchback:
> I don't understand what you are releasing. As far as I know, the source code in both zip and tar.gz format is attached to _any_ github release already. 

I removed the tag.gz (although was not able to find that the tag.gz and zip are created _also_ in the case of a customized release (once we run this change we will know for sure). 

I added README.md just as an example / place holder.

> What we need for the main repo is described in #82, including the first step which is to create an empty release but to tag docker images appropriately.

my change causes the creation of a release automatically on tagging ... as opposed to manually invoking github UI, as our release include non-default files (such as change-log or other files) its' important the process is automatic.

As to tagging docker images, this currently has no meaning and can't be tested as we don't yet support redirecting all images in the project to use a specified tag other than `latest`, this will be handled after changes by @froesef and #42 and as these images (and helm charts) will be tagged in the public repos this change will not effect the format of the github release (which is the focus of this change).

roee88 (approved):

.travis.yml
roee88: @ronenkat are we good here?

Issue #103, opened 2020-11-01
Support modules with multiple flows/data stores


shlomitk1:
Problem statement: a module that can support multiple flows, e.g. read and write, or multiple datasets should be instantiated once. However, the current version will instantiate multiple instances of the same module.

Use-case 1: two data sets A and B, a single read module is selected for both. Currently, the blueprint contains two instances of a module: one for A, another for B. We want a single instance of the module for both A and B.

Use-case 2: A single module supports both read and write. Currently, the blueprint contains two instances of a module: for read and for write. We want a single instance of the module with both read and write sections inside its arguments.

Required:
- to remove oneOf selector in ModuleArguments
- to unite instances of the same module into one inside the blueprint

Affected components: manager (blueprint generation)

simanadler:
> Problem statement: a module that can support multiple flows, e.g. read and write, or multiple datasets should be instantiated once. However, the current version will instantiate multiple instances of the same module.
> 
> Use-case 1: two data sets A and B, a single read module is selected for both. Currently, the blueprint contains two instances of a module: one for A, another for B. We want a single instance of the module for both A and B.

@shlomitk1 @roee88 @ronenkat 
1) What about transformations?  Will the transformations be per data set, or do they have to match in order to be merged into a single instance?

2) As we move to multi-cloud I think the logic will be more complex than this.  Since we will begin multi-cloud work shortly, I suggest to wait before making this change.

> 
> Use-case 2: A single module supports both read and write. Currently, the blueprint contains two instances of a module: for read and for write. We want a single instance of the module with both read and write sections inside its arguments.
> 
> Required:
> 
>     * to remove oneOf selector in ModuleArguments
> 
>     * to unite instances of the same module into one inside the blueprint
> 
@shlomitk1 
Same comments as the previous scenario 
> 
> Affected components: manager (blueprint generation)

@shlomitk1 @roee88 @ronenkat  As we move to multi-cluster we need to think whether this optimization will be done by the blueprint generator or by the blueprint controller.

shlomitk1:
@simanadler Transformations are a part of the Read, together with source and destination.
Suppose we have a dataset that needs to be read from S1 while performing action1 and action2, in addition to a dataset that needs to be read from S2 while performing action3.
Instead of a read module for reading from S1 doing action1,action2 AND another read module reading from S2 doing action3 we'll get a single read module that will (1) read from S1 doing action1,action2 AND (2) read from S2 doing action3.

[...]

PR #116, opened 2020-11-03
Support multiple data stores

shlomitk1:
This PR fixes https://github.com/IBM/the-mesh-for-data/issues/103
Goal: to avoid multiple instances of any given read/write module 

Introduced changes:
- ModuleArguments structure does not contain Flow. Arguments may contain any number of Read and Write elements.
- Prior to blueprint generation, module instances are revisited. If there are several instances of the same read/write module, they are combined into a single instance.
- Added a test that checks that a single read module is constructed in a blueprint in the case of multiple data assets.


simanadler:
@shlomitk1 @roee88  Please see my comments in the issue this addresses https://github.com/IBM/the-mesh-for-data/issues/103.

roee88 (approved):

PR #123, opened 2020-11-05
policy-compiler: add data_location field in egeria input asset json 

### LINKED ISSUE #122: Kubeflow sample: M4DApplication returns a failure status
Problem: the kubeflow sample is no longer working. The reason is that Egeria connector returns "geo not supported yet" instead of the data location. Since transformations need to be made, manager requires a copy to the processing geography. No copy modules are installed in the sample (no buckets exist either), thus no blueprint is created.
Solution: add “geo” in fullPath field in https://github.com/IBM/the-mesh-for-data/blob/master/third_party/egeria/usage/example_transactions.csv.json#L5 and it will “automatically” appear in qualifiedName  in https://github.com/IBM/the-mesh-for-data/blob/master/connectors/egeria/src/main/resources/json/asset/asset_metadata.json#L42. Then the string will be parsed by egeria-connector, and the "geo" field will be passed to manager. 
Requires changes to egeria-connector, kubeflow sample documentation.

rohithdv:
This PR fixes #122. We have added a "data_location" field (which is mandatory) in the asset json which is input to Egeria during configuration time. This field needs to be populated with the correct data_location of the asset and will be in turned parsed by egeria connector and set in the geo field of the data catalog response https://github.com/IBM/the-mesh-for-data/blob/89a5359b4f4c911eb366bfaa8612f25b919b0354/pkg/connectors/protos/data_catalog_response.proto#L54 which is sent to the manager. The manager further will process the geo field accordingly.

### DIFF
--- a/connectors/egeria/src/main/java/com/ibm/egeria/EgeriaClient.java
+++ b/connectors/egeria/src/main/java/com/ibm/egeria/EgeriaClient.java
@@ -411,6 +411,10 @@ public final class EgeriaClient {
 
         //it can contain a direct link to the file or a json with remote object
         String qualifiedName = assetMetaDataHelper.getQualifiedName();
+        // fix for https://github.com/IBM/the-mesh-for-data/issues/122 - start 
+        JsonObject storeJson = JsonParser.parseString(qualifiedName).getAsJsonObject();
+        String geo = storeJson.get("data_location").getAsString();
+        // fix for https://github.com/IBM/the-mesh-for-data/issues/122 - end 
         DataStore dataStore = getAssetStore(qualifiedName);
 
         datasetDetails = DatasetDetails.newBuilder()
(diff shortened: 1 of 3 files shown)


shlomitk1 (changes requested):
Please modify this line as well and set geography to US: https://github.com/IBM/the-mesh-for-data/blob/89a5359b4f4c911eb366bfaa8612f25b919b0354/samples/kubeflow/example_transactions.csv.json#L5

rohithdv:
> Please modify this line as well and set geography to US:
> 
> https://github.com/IBM/the-mesh-for-data/blob/89a5359b4f4c911eb366bfaa8612f25b919b0354/samples/kubeflow/example_transactions.csv.json#L5

Fixed. Please verify. Thanks.

shlomitk1 (approved):

PR #126, opened 2020-11-08
cert-manager: download resources into travis cache

hunchback:
fixing one of the issues in https://github.com/IBM/the-mesh-for-data/issues/88 by using a local cache (backed up by travis)

```
error: unable to read URL "https://github.com/jetstack/cert-manager/releases/download/v0.13.1/cert-manager.yaml", server reported 429 too many requests, status code=429
```

### DIFF
--- a/third_party/cert-manager/Makefile
+++ b/third_party/cert-manager/Makefile
@@ -2,16 +2,21 @@ ROOT_DIR:=../..
 include $(ROOT_DIR)/Makefile.env
 include $(ROOT_DIR)/.mk/tools.mk
 
-YAML:=https://github.com/jetstack/cert-manager/releases/download/v0.13.1/cert-manager.yaml
+REMOTE:=https://github.com/jetstack/cert-manager/releases/download/v0.13.1/cert-manager.yaml
+LOCAL:=$(ABSTOOLBIN)/cert-manager.yaml
+
+$(LOCAL):
+	mkdir -p $(shell dirname $(LOCAL))
+	wget -x -O $(LOCAL) $(REMOTE)
 
 .PHONY: deploy
-deploy: $(TOOLBIN)/kubectl
-	$(TOOLBIN)/kubectl apply --validate=false -f $(YAML)
+deploy: $(TOOLBIN)/kubectl $(LOCAL)
+	$(TOOLBIN)/kubectl apply --validate=false -f $(LOCAL)
 
 .PHONY: deploy-wait
 deploy-wait: $(TOOLBIN)/kubectl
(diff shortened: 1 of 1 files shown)


froesef:
I would say let's not do this via some caching mechanism but just download the file locally and push it into git.

third_party/cert-manager/Makefile
```diff
@@ -2,16 +2,21 @@ ROOT_DIR:=../..
 include $(ROOT_DIR)/Makefile.env
 include $(ROOT_DIR)/.mk/tools.mk
 
-YAML:=https://github.com/jetstack/cert-manager/releases/download/v0.13.1/cert-manager.yaml
+REMOTE:=https://github.com/jetstack/cert-manager/releases/download/v0.13.1/cert-manager.yaml
+LOCAL:=${HOME}/.download/cert-manager.yaml
```
roee88, line 6: Do we really want to keep this in the user's machine (when not running from Travis)? 

Maybe instead we should make the default `LOCAL` empty and only if it's defined then add the `$(LOCAL):` target and use it in `deploy`, otherwise use $REMOTE directly.

hunchback: @roee88 so what you are saying is that you want to create two different flows:
- travis host: use cached version of yaml
- development host: always download from internet

this is a change which degrades the code as it creates (needlely) two different flows rather than having a simple flow (the `travis host` one) which works well for all cases.

splitting into two flows may make sense if there is a _huge_ benefit such as saving development time of similar ... but in this case I see no benefit.

roee88: All I'm saying is that I don't like getting things downloaded to a `.download` dir on my laptop. That's why I'm asking if you can please change it to something else, and I just gave an example of what else can be done. You can alternatively just change the default LOCAL to be the current folder (and add to .gitignore) or take the approach described by @froesef to just vendor the YAML in the code.

hunchback: @roee88 / @froesef - I don't want to vendor the yaml file as its' a large 6.5K line file, thus it is anyway don't going to be maintained by us - also considering as going forward as we will want to upgrade cert manager from the current version to newer (so that it will be compatible with newer environments) its' best that it remains outside of the project.

regards `.download` location - will will change it so that it is placed within the project (same as hack/tools/bin) for both development and travis (@roee88 - please confirm that this is acceptable and I will do the modification)

roee88: Sure, just make sure gitignore captures it.

roee88 (approved):

Issue #96, opened 2020-10-28
Taxonomy for dataset credentials and connection properties. 


roee88:
# Problem description

There are different connection types (s3, kafka, etc.). Each connection type has a different structure for credentials (e.g., s3 credentials includes access key and secret key) and connection properties (e.g., endpoint).

However, the exact structure for each type is undefined. We have inconsistencies between different components in the project (connectors and modules). For example, `mover` expects `{'data': {'access_key': '<value>', 'secret_key': '<a3026f48a13ff1246e26b09f99237eb457bec35e4321ab12>'}, ...}` while `flight` expects  `{'access_key': '<value>', 'secret_key': '<value>'}`

# Proposed solution

Document a fixed mapping that must be followed by all connectors and modules. 
The preferred structure for credentials is one that can easily be used when adding a secret to vault through its UI.

froesef:
Actually the mover can deal with data being in `data` or outside like flight. But we were running into the problem that the mover was expecting `accessKey` and `secretKey`. So I completely agree that there is an issue and that we should unify this!

One way is that we could fix the credentials in the protobuf return type of the connector (and use the same structure for the secretProvider)

Thus in the example below a connector would set the fields where it has a value for. As protobuf structs in go get generated with `json:"field,omitempty"` when converting a struct to JSON it would result in an object that only has the fields that are not empty. This would allow for a standard mapping of credentials from the connectors.
```
message Credentials {
    string access_key = 1; 
    string secret_key = 2; 
    string user = 3; 
    string password = 4; 
    string api_key = 5; 
}
```

roee88:
@froesef why the preference to put all credential forms in a single message? Don't we also want to attach it to the connection type to associate the expected credentials for each connection type? 

I wonder if we need to have more structure here, e.g., to differentiate between S3 and COS because the latter can have an API key without HMAC credentials so a module with code that requires access_key and secret_key won't fit. Does the mover module support both?

froesef:
No the mover is currently based on the open-source stocator plugin that is working with HMAC credentials. As I understand we'd have to switch to the `ibm-sdk` version of it to support API key. I chose the HMAC version as these can also be used with different S3 implementations.

rohithdv:
[...]

PR #137, opened 2020-11-15
Taxonomy for dataset credentials and connection properties 

rohithdv:
Fix for #96. 
This PR expands [DatasetCredentials Response Proto](https://github.com/IBM/the-mesh-for-data/blob/973a84f7bc1b97175c1c6c24366175cf67d5e4ca/pkg/connectors/protos/data_credential_response.proto#L11) with fields specific to COS, DB2 scenarios as discussed in https://github.com/IBM/the-mesh-for-data/issues/96#issuecomment-718147708 and  https://github.com/IBM/the-mesh-for-data/issues/96#issuecomment-721667258. This PR also has the associated changes to  vault connector module which processes the DatasetCredentials GRPC request.

Some minor changes (like updating a new asset in egeria, etc) are also committed in this PR for local testing purpose only.


pkg/connectors/protos/data_credential_response.proto
```diff
@@ -6,7 +6,27 @@ syntax = "proto3";
...
+    string bucket = 8;
+    string database = 9;
+    string host = 10;
+    string port = 11;
+    string ssl = 12;
+    string connection_id = 13;
+    string connection_path = 14;
+    string custom_credentials_json = 15; // this field will be used in case we are using Egeria as the catalog since egeria does not support the above fields at the moment. Also this field will be used in case of WKC also when remoteStatus is REMOTE_CUSTOM (when a local asset is uploaded in WKC GUI), then we get a credentials json from WKC. we are not sure which fields will be present. so this field will be sending the json as it is received from WKC.
+    bool is_remote = 16;
+}
```
shlomitk1, line 26: Most of this information describes DataSource. We have another proto for passing this information. 
I would include only the first five fields.

shlomitk1 (changes requested):
data_catalog_response and data_credential_response are mixed up.

pkg/connectors/protos/data_credential_response.proto
```diff
@@ -6,7 +6,27 @@ syntax = "proto3";
...
+    string resource_instance_id = 6;
+    string url = 7;
+    string bucket = 8;
+    string database = 9;
+    string host = 10;
+    string port = 11;
+    string ssl = 12;
+    string connection_id = 13;
+    string connection_path = 14;
+    string custom_credentials_json = 15; // this field will be used in case we are using Egeria as the catalog since egeria does not support the above fields at the moment. Also this field will be used in case of WKC also when remoteStatus is REMOTE_CUSTOM (when a local asset is uploaded in WKC GUI), then we get a credentials json from WKC. we are not sure which fields will be present. so this field will be sending the json as it is received from WKC.
```
shlomitk1, line 24: The idea of the taxonomy is that any component accessing credentials will know what keys to expect. It is the responsibility of a connector (egeria or other) to provide the credentials exactly as requested, or report an error otherwise.

rohithdv: The data credential proto affects the interaction between the manager and connector (either WKC Connector when WKC is used as a catalog or Vault Connector when Egeria is used as a catalog). 

_Scenario: Using WKC as a catalog_ 
Actually in our existing code of WKC Connector, we are already providing most of these fields in the context of supporting credential information returned from WKC. WKC through its connection API provides us with information about these fields. You can see example JSON returned by WKC Connection API [here](https://github.ibm.com/data-mesh-research/WKC-connector/blob/43100b73fb150c7acac7ac8dad8304f5fa0568d1/java/src/main/resources/json/connection/connection_metadata.json#L25). 

We are just transferring the information we got from WKC to the proto through these fields. In fact, in the current codebase in the WKC Connector, we are already doing it but in the form of a single string json. You can check the existing WKC Connector code [here](https://github.ibm.com/data-mesh-research/WKC-connector/blob/43100b73fb150c7acac7ac8dad8304f5fa0568d1/java/src/main/java/com/ibm/wkc/ConnectionMetaDataHelper.java#L101). 

Also in the case of WKC asset which is uploaded through the WKC GUI (not through a WKC Connection) we do not have any connection information but only have some json credential information as given in the example [here](https://github.ibm.com/data-mesh-research/WKC-connector/blob/43100b73fb150c7acac7ac8dad8304f5fa0568d1/java/src/main/resources/json/asset/asset_metadata.json#L243). Hence in this case we just take the json and put it in the `custom_credentials_json` field in the proto. 

_Scenario: Using Egeria as a catalog_ 
In case of Egeria, to the best of our understanding, we do not have information about what fields egeria provides us as part of connection credentials. In fact in our code now, we are assuming the credential information returned about Egeria is something like [here](https://github.com/IBM/the-mesh-for-data/blob/973a84f7bc1b97175c1c6c24366175cf67d5e4ca/connectors/vault/user_vault_init.sh#L39). Hence, we are including such a response in the `custom_credentials_json` field in the proto. 

To summarise, In this PR, we have now structured it  by enlisting the fields corresponding to connection credentials  separately to make it compatible with WKC and Egeria. Anyways the manager module can take whichever fields it expects from the protobuf data structure. I think it does not harm in any way if we have additional information.

pkg/connectors/protos/data_credential_response.proto
rohithdv: Some thoughts are provided [here](https://github.com/IBM/the-mesh-for-data/pull/137#discussion_r524485252). Please let me know your views. Thanks.

pkg/connectors/protos/data_credential_response.proto
shlomitk1: I understand that API of the external service (Egeria, WKC, ...) allows us to get a lot of information - connection details, credentials,... However, we have a clear separation between connection details that are not a secret, and credentials (secret_key,access_key,password) that are kept as a secret, e.g. stored in Vault. 
The connection details are already taken care of in data_catalog_response proto. This PR refers to changes to data_credentials_response.
Two main issues here:
1. data_credentials_response should not contain connection details, as there is another interface (manager-connector) for getting this information
2. The issue that this PR should fix is that currently, connectors pass some json that they received from an external catalog to manager, manager transfers (via secret provider) the same json fields to the modules. However, modules expect very specific information which is using strict keywords, e.g. access_key and secret_key. There can be many different catalogs with the corresponding connectors, for each the credentials may take a different form. Therefore, the connectors are responsible to provide the expected information regardless of the catalog-connector communication. Manager is responsible to store the expected information as a secret and pass it further to the active modules. 
The connector may receive credentials as a part of a large json string. It is supposed to extract the specific fields and provide them to manager. If these fields are not there, the custom json string won't help. In this case the connector should report an error. To detect errors at the early stage is important. Otherwise, we'll get a problem after the modules are orchestrated, and some of them terminate with error because they can't get the expected credentials, and when it happens, it will be much more difficult to detect the problem.

rohithdv: got it. is there any document or interface or file which lists what fields the modules expect ? then we could design the credentials proto based on this interface. If tomorrow we add more modules which expect some other field then we should have a way to change the credential proto also easily. also then it would be easy for the connector to flag an error if any of the fields which the modules expect is not provided by the corresponding catalog.

third_party/egeria/deploy.sh
```diff
@@ -3,11 +3,11 @@
 set -e
 set -x
 
-: ${WITHOUT_OPENSHIFT=true}
+: ${WITHOUT_OPENSHIFT=false}
 
 CHART=odpi-egeria-lab
 RELEASE=lab
-NAMESPACE=egeria-catalog
+NAMESPACE=irltest1
```
shlomitk1, line 10: Why was the namespace changed?

rohithdv: sorry it was edited by mistake during testing. restored it now. Thanks.

shlomitk1 (approved):
Changed manager appropriately.

PR #142, opened 2020-11-19
Multicluster: kind support

froesef:
This PR Adds support to spin up two kind clusters in the same dev environment. 
These kind clusters share one registry so that images only have to be uploaded once.

### DIFF
--- a/hack/tools/create_kind.sh
+++ b/hack/tools/create_kind.sh
@@ -59,37 +59,16 @@ certs_delete() {
 }
 
 kind_delete() {
-        bin/kind version
-        bin/kind delete cluster --name kind
-}
-
-kind_delete_control() {
-        bin/kind version
-        bin/kind delete cluster --name control
+        bin/kind delete cluster --name $1
 }
 
 kind_create() {
-        bin/kind version
-        bin/kind create cluster --name kind \
+        bin/kind create cluster --name $1 \
              -v 4 --retain --wait=0s \
-             --config ./kind-config.yaml \
+             --config ./$2 \
              --image=kindest/node:$K8S_VERSION
-        bin/kubectl config use-context kind-kind
-        for node in $(kind get nodes --name kind); do
-          bin/kubectl annotate node "${node}" "tilt.dev/registry=kind-registry:5000";
-          docker cp ../registry/ca.crt "$node":/usr/local/share/ca-certificates
-          docker exec "$node" update-ca-certificates
-        done
-}
-
-kind_create_control() {
-        bin/kind version
-        bin/kind create cluster --name control \
-             -v 4 --retain --wait=0s \
-             --config ./kind-control-config.yaml \
-             --image=kindest/node:$K8S_VERSION
-        bin/kubectl config use-context kind-control
-        for node in $(kind get nodes --name control); do
+        bin/kubectl config use-context kind-$1
+        for node in $(kind get nodes --name $1); do
           bin/kubectl annotate node "${node}" "tilt.dev/registry=kind-registry:5000";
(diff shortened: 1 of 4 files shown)


website/content/contribute/code/flow/index.md
```diff
@@ -120,6 +120,17 @@ Running end to end tests:
 make e2e
 ```
 
+# Building in a multi cluster environment
+
+As the-mesh-for-data is can run in a multi-cluster environment there is also a test environment
```
roee88, line 125:
```suggestion
As {{< name >}} can run in a multi-cluster environment there is also a test environment
```

hack/tools/create_kind.sh
```diff
@@ -63,14 +63,33 @@ kind_delete() {
...
         bin/kubectl config use-context kind-kind
-        for node in $(kind get nodes); do
+        for node in $(kind get nodes --name kind); do
+          bin/kubectl annotate node "${node}" "tilt.dev/registry=kind-registry:5000";
+          docker cp ../registry/ca.crt "$node":/usr/local/share/ca-certificates
+          docker exec "$node" update-ca-certificates
+        done
+}
+
+kind_create_control() {
```
roee88, line 85: why not create a function that accepts name and yaml location

hack/tools/create_kind.sh
```diff
@@ -83,6 +102,15 @@ case "$op" in
         registry_delete || true
         certs_delete || true
         kind_delete || true
+        kind_delete_control || true
+        ;;
+    multi)
```
roee88, line 107: I wonder if we would want to create/delete one cluster independently

hack/tools/kind-control-config.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+  extraPortMappings:
+  - containerPort: 30333  # Razee dash API node port in K8s
+    hostPort: 3333  # Razee dash API on host
+  - containerPort: 30080  # Razee dash UI node port in K8s
+    hostPort: 8080  # Razee dash UI on host
```
roee88, line 12: Eventually we may not want to expose the UI to the outsiders but rather just port-forward to access it.

roee88 (commented):
I wonder if

hack/tools/kind-control-config.yaml
froesef: Yes but I think for local development in kind it's easier to have this exposed right now.

hack/tools/create_kind.sh
```diff
@@ -63,14 +63,33 @@ kind_delete() {
         bin/kind delete cluster --name kind
 }
 
+kind_delete_control() {
+        bin/kind version
```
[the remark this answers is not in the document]
froesef: Removed

hack/tools/create_kind.sh
froesef: I used parameters for create and delete now. Does that also solve this comment?

roee88: See https://github.com/IBM/the-mesh-for-data/pull/142#pullrequestreview-534300178 to understand what I meant, but we are good for now.

website/content/contribute/code/flow/index.md
```diff
@@ -120,6 +120,17 @@ Running end to end tests:
 make e2e
 ```
 
+# Building in a multi cluster environment
+
+As {{< name >}} is can run in a multi-cluster environment there is also a test environment
```
roee88, line 125: Why "is can"?

roee88 (approved):

PR #148, opened 2020-11-20
Multicluster: Razee installation on kind

froesef:


roee88:
Quick comment, this should be in third_party.

hack/tools/create_kind.sh
```diff
@@ -30,22 +30,33 @@ registry_create() {
...
+    if [ ! -f ../registry/registry.crt ]; then
+      openssl genrsa -out ../registry/registry.key 2048
+      openssl req -new -key ../registry/registry.key -out ../registry/registry.csr -subj '/C=US/ST=NY/O=IBM/CN=kind-registry' -extensions EXT -config <(printf "[dn]\nCN=kind-registry\n[req]\ndistinguished_name = dn\n[EXT]\nsubjectAltName=DNS:kind-registry,DNS:localhost\nkeyUsage=digitalSignature\nextendedKeyUsage=serverAuth")
+      openssl x509 -req -in ../registry/registry.csr -CA ../registry/ca.crt -CAkey ../registry/ca.key -CAcreateserial -out ../registry/registry.crt
+    fi
 }
 
 install_certs() {
     if [[ "$OSTYPE" == "linux-gnu"* ]]; then
         sudo cp ../registry/ca.crt /usr/local/share/ca-certificates
```
roee88, line 48: Not related to this PR but I think it isn't true for all linux machines (e.g., RHEL uses a different folder + binary). Need to at least document this.

hack/tools/create_kind.sh
```diff
@@ -54,6 +65,7 @@ install_certs() {
 certs_delete() {
     if [[ "$OSTYPE" == "darwin"* ]]; then
       security remove-trusted-cert  ../registry/ca.crt
+      security delete-certificate -c themeshfordata
     fi
     rm -rf ../registry
```
roee88, line 70: Not related to this PR but don't we need to clean it up from the global CA folder as well?

roee88 (commented):
Can you please explain why we must vendor modified versions of the Razee YAMLs?

froesef:
The vendoring is because we want to set the local environment variable at some point for development. When using the all-in-one.yaml from razee it downloads the deployments via RemoteResource, which is nice because you get updates but if you change your deployment manually (e.g. to set the local authentication environment variable) it's also bad because it will overwrite your manually adapted deployment with what is defined remotely. 
So for now I thought the easiest is to use the yamls so we can adapt them instead of pulling them from razee. Another solution would be to use a templated razee version if that is availabe.

hack/tools/create_kind.sh
froesef: I'm not sure what you mean with the global CA folder. Can you please elaborate? 
The default is to store the certificate in the user cert store. And this removes it from the default user cert store. (Adding it in the admin cert store would require the `-d` flag when adding the trusted cert.) What I noticed is that for trusting/untrusting a certificate it's in need for the user password.

roee88: My question is where do you remove the registry certificate from the user cert store in ubuntu (/usr/local/share/ca-certificates/)?. Also do you think we should create a file with a more unique name than ca.crt? Also after removal a call to update-ca-certificates with the --fresh (or -f) flag is needed.

hack/tools/create_kind.sh
roee88: I assumed that

> The vendoring is because we want to set the local environment variable at some point for development. When using the all-in-one.yaml from razee it downloads the deployments via RemoteResource, which is nice because you get updates but if you change your deployment manually (e.g. to set the local authentication environment variable) it's also bad because it will overwrite your manually adapted deployment with what is defined remotely.
> So for now I thought the easiest is to use the yamls so we can adapt them instead of pulling them from razee. Another solution would be to use a templated razee version if that is availabe.

I think it's worth asking on the Razee slack channel if there is a better approach than vendoring it. But okay for now.

roee88:
> The vendoring is because we want to set the local environment variable at some point for development. When using the all-in-one.yaml from razee it downloads the deployments via RemoteResource, which is nice because you get updates but if you change your deployment manually (e.g. to set the local authentication environment variable) it's also bad because it will overwrite your manually adapted deployment with what is defined remotely.
> So for now I thought the easiest is to use the yamls so we can adapt them instead of pulling them from razee. Another solution would be to use a templated razee version if that is availabe.

I think it's worth asking on the Razee slack channel if there is a better approach than vendoring it. But okay for now.

roee88 (approved):

hack/tools/create_kind.sh
froesef: That's a good question. Currently it's not removed from the ubuntu ca-certificates. I did not test this on a ubuntu machine for some time. I'll check it out.

hack/tools/create_kind.sh
```diff
@@ -57,16 +57,21 @@ install_certs() {
...
 }
 
 certs_delete() {
-    if [[ "$OSTYPE" == "darwin"* ]]; then
-      security remove-trusted-cert  ../registry/ca.crt
+    if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+      # This has been tested on Ubuntu. On other distributions this may vary.
+      # If developers use different distributions please add a branch here that works.
+      sudo rm /usr/local/share/ca-certificates/themeshfordata-ca.crt
+      sudo update-ca-certificates
```
roee88, line 72: I read that without --fresh some dangling symlinks are kept in /etc/ssl

roee88 (approved):

Issue #133, opened 2020-11-12
Ingress security of connectors in m4d-system namespace


revit13:
This is sub issue of the Control plane security feature: https://github.com/IBM/the-mesh-for-data/issues/57

Goals:
1) Restrict access to connectors:
   1.1) Default behavior: Enable ingress to connectors *only* from the manager and other connectors if needed .
   1.2) add connector-connector ingress policy if needed by specific configuration
2) mTls between the connectors and the manager

Tasks:
1) Install istio as third party (https://github.com/IBM/the-mesh-for-data/issues/132)
2) Inject istio side cars to the manager and connectors 
3) Apply istio authorization policy and authentication policy to achieve the above goals
4) Document the security elements in the documentation

PR #152, opened 2020-11-22
[RFC] Initial support for control plane security

revit13:
This patch is the first step towards securing the traffic between the components in m4d-system namespace . It adds the following features using Istio:
1) Enforce connectors ingress traffic to be only from the manager and other connectors if needed.
2) Mutual TLS is enabled for the traffic between the manager and the connectors. 

I would appreciate your comments is particular regarding the following:

- Testing: I am doing two kind of manual testing: 
1) in kiali I check that the traffic between the manager and the connectors is encrypted (marked with a small lock) 
2) validate that the connectors can not be accessed from a dummy container (getting rbac error)
I am not sure if that type of testing is needed to be automated.

 - Authorization policy is done by applying AuthorizationPolicy CRD  for each connector. Principles is the source of the requests (please see https://istio.io/latest/docs/concepts/security/#principals). These Principles are hard coded in the yaml file because the traffic between the connectors is connectors specific. For example, egeria has ingress traffic from the manager and opa connector. So I did not think that using kustomization tool is required here but I might be wrong.

- In this patch Envoy sidecars are injected selectively to the manager and connectors  but there is another option to automatically inject them to newly created pods in m4d-system namespace. I decided to  go with the selective approach to avoid injecting the Envoy sidecar to pods that are not part of the M4D core components like opa. I can change that if automatic injection is more reasonable.

Tested on kind v0.8 with Istio 1.7.4 and currently testing on OpenShift.

** UPDATE: The code was tested also with Istio 1.8.


roee88:
Quick comments:
1. Leave testing to future PR, it will take time to settle on the right way to implement this feature anyway.
2. We need a simpler approach. You require too much from connectors instead of having a strict set of simple requirements and moving the complex parts to the core of the project. Prefer simple but sufficient given reasonable assumptions to anything complex. For example, if we require a specific label in connector pods to indicate that it's an m4d connector and create a single authorization policy based on it, then it does have some implications but could be sufficient (here the assumptions would be: only m4d administrators can install things to m4d-system, all installed connectors are trusted. 
3. Depending on Istio installation, automatic injection behaves differently. Options are: no auto injection, auto injection if namespace is labeled and pod is not labeled with "don't inject", auto injection if namespace is labeled and pod is labeled with "do inject". We can make the assumption (document it) that auto injection is enabled. Then add namespace label and appropriate pod labels to everything we deploy in m4d-system. Move out third parties like vault if this is an issue.
4. I'm already lost with all the kustomize overlays in the project. Let's avoid adding kustomize complexity unless we have to. Hopefully a future move to Helm will give us more flexibility.

simanadler:
Against what/whom does this solution protect?  Can you describe the potential attack vector that it solves? The m4d-system namespace and the components (manager and connectors) that run in it are deployed by the IT admin via M4D scripts.  Am I correct in assuming that m4d-system is not accessible by others, and that other workloads can't just be set to run in m4d-system?  If so, I'm not clear against what we are trying to protect and why the isolation between the manager and the connectors is required, but I could well be missing something.  If we are protecting from the admin, couldn't he just change the policies?

roee88:
@simanadler the connectors can currently be accessed by anything in any namespace. The encryption parts are because the traffic is currently not encrypted so pods that happen to run in the same node may br able to sniff traffic that includes things like credentials.

revit13:
@roee88 Thanks for your comments. I tried to address them in the new version. Thanks

connectors/helpers/base/egr_connector.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
   name: egr-connector
   labels:
     app: egr-connector
+    type: connector
```
roee88, line 10: To avoid conflicts labels should include a prefix. For example `m4d.ibm.com/componentType`

website/content/docs/setup/istio/index.md
```diff
@@ -1,11 +1,32 @@
...
+
+Upon {{< name >}} deployment the Envoy sidecar is automatically injected to pods in the control-plane namespace (called 4d-system be default). To support automatic injection Istio's [MutatingAdmissionWebhook](https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#automatic-sidecar-injection) should be turned on in the cluster.
+
+To check if MutatingAdmissionWebhook is turned on in the cluster run:
+
+```bash
+kube-apiserver -h | grep enable-admission-plugins
+```
+
+To turn it on in Kind run:
```
roee88, line 17: I don't think that this is needed. I used istio with kind before with auto injection enabled and never had to use kubeadm.

revit13: OK, will remove this sentence then.

connectors/helpers/base/egr_connector.yaml
revit13: OK,

revit13:
I create a new PR for #133 due to rebase problems. Please ignore this one. tnx

Issue #132, opened 2020-11-11
Move Istio to third_party


roee88:
Currently Istio is installed with the `make kind` command. Instead, it should move to `third_party` similar to how we install other cluster dependencies like cert-manager. When it becomes a strong dependency it should be added to cluster-prepare.

revit13:
Please note that there is an issue in kubeflow that prevents it from using istio 1.6 and up due to usage of deprecated  ClusterRbacConfig  resource in kubeflow  https://github.com/kubeflow/manifests/issues/1535.

roee88:
> Please note that there is an issue in kubeflow that prevents it from using istio 1.6 and up due to usage of deprecated ClusterRbacConfig resource in kubeflow kubeflow/manifests#1535.

Kubeflow compatibilities should not block us, they will catch up and if needed we can change to a simpler Jupyter sample that doesn't require Kubeflow. 

@hunchback Istio 1.8 is out. I have read that installation backward-compatibility is supposed to be better from this point forward. Let's switch to 1.8 with istioctl installation using YAML input (kind IstioOperator) instead of flags.

roee88:
@revit13 Following https://github.com/IBM/the-mesh-for-data/pull/154#issuecomment-731839677 and the comments below.

Put the quick start aside for a minute. A proper installation of the project would not require any install script but would just be applying a YAML or a Helm chart to install the project. It will not install any third party. See how Kubeflow is moving away from this approach. Instead, installation instructions should clearly state what is required and direct the user to install on his own as a prerequisite. Sure, we can also put hints to a script that we use in the CI *if it exists* but give it the right context. This is why we need to lower the number of strong requirements. The core of the project has dependencies on cert-manager, Vault and (optionally) Istio.

All of our installation scripts for anything that is not part of the project are for CI usage only. When we will run CI against an OpenShift cluster we will choose what to install during CI and what apriori. Things like cert manager and Istio would probably be installed to the cluster outside of the CI scope. For other things, if the CI needs it then we may script it but definitely not for production level (e.g., our Vault installation is the most basic install).



Back to the quick start. It's actually targeting a quick evaluation of the project using the default connectors stack. It does not require everything to be installed in a script and can definitely include "Install Istio" as part of the instructions, if needed.

revit13:
[...]

PR #154, opened 2020-11-22
istio: moved to third_party (now installing 1.8.0)

hunchback:
to install a new version run:

```
make -C third_party/istio undeploy
rm hack/tools/bin/istioctl
make -C third_party/istio deploy
make -C third_party/istio deploy-wait
```


roee88:
Please see https://github.com/IBM/the-mesh-for-data/issues/132#issuecomment-731708401

revit13:
Hi Aiden, I agree with Roee's comment in the issue that we should use istio 1.8 now that its out. 
Also, addonComponents.kiali.enabled is deprecated; the samples/addons/ deployments is used instead.
 (I used it in my patch - https://github.com/revit13/the-mesh-for-data/blob/control-plane-security-133/hack/tools/create_istio.sh)
Another comment is regarding OpenShift - I think the installation on OpenShift  might require additional commands. I've read that Istio 1.8 has profile for OpenShift so this is another reason to use Istio 1.8 https://developer.ibm.com/blogs/istio-1-8-release-blog/
I am not familiar with the istio Operator that Roee mentioned but I will try to experiment with to see how to install it.
Also, I found some error running the commands in create_istio.sh -- it seems that install should be used instead of apply in istio_install() function. Please also see my version above.

roee88:
I'm not sure that we need OpenShift installation here. Let's focus on kind for this PR.

revit13:
@roee88 In https://ibm.github.io/the-mesh-for-data/docs/setup/quickstart/  it mentioned that there is an option to run  on OpenShift with WITHOUT_OPENSHIFT=false. Thus I think we need to handle also OpenShift installation.

roee88:
@revit13 moving the discussion to the issue, this isn't the place to discuss it.

hunchback:
@roee88 / @revit13 - in the https://github.com/istio/istio/releases/ page version 1.7.5 is labeled as the latest release while 1.8.0 is still new and thus not yet designated as the latest release. So best we remain with 1.7.5.

froesef:
Do we need to install istio on every cluster setup? Do we currently use it in any tests? This increases the runtime of the CI by 2 min as the istio setup is not running in parallel. (Or possibly we could start it to run in parallel to the other prepare scripts)

Makefile
```diff
@@ -18,12 +18,14 @@ test:
 
 .PHONY: cluster-prepare
 cluster-prepare:
+	$(MAKE) -C third_party/istio deploy
```
roee88, line 21: As mentioned in the issue, let's wait with adding it to cluster prepare until this becomes a strong dependency (i.e., when we merge tests for control plane security)

hack/tools/install_istio.sh
```diff
@@ -5,7 +5,7 @@
 
 source ./common.sh
 
-export ISTIO_VERSION=1.6.2
+export ISTIO_VERSION=${ISTIO_VERSION:=1.7.5}
```
roee88, line 8: Update to 1.8

third_party/istio/Makefile
```diff
@@ -0,0 +1,19 @@
+ROOT_DIR:=../..
+include $(ROOT_DIR)/Makefile.env
+include $(ROOT_DIR)/.mk/cluster.mk
+
+.PHONY: deploy
+deploy: $(TOOLBIN)/kubectl $(TOOLBIN)/istioctl
+	cd $(TOOLS_DIR); ./create_istio.sh
```
roee88, line 7: Please update the installation to match the instructions in https://istio.io/latest/docs/setup/install/istioctl/

roee88: Provide parameters to the installation using YAML file instead of flags

roee88 (changes requested):

hack/tools/install_istio.sh
hunchback: ok

Makefile
hunchback: ok

third_party/istio/Makefile
hunchback: ok

hack/tools/create_istio.sh
```diff
@@ -46,25 +51,6 @@ istio_config() {
     kubectl label namespace default istio-injection=enabled --overwrite
```
roee88, line 51: why do we need to label the default namespace?

third_party/istio/Makefile
```diff
@@ -0,0 +1,19 @@
...
+
+.PHONY: deploy-wait
+deploy-wait: 
+	$(TOOLBIN)/kubectl wait --for=condition=available -n istio-system deployment/istiod --timeout=180s
```
roee88, line 11: Is 180s an arbitrary choice?

roee88 (changes requested):

hack/tools/create_istio.sh
```diff
@@ -12,21 +12,26 @@ istio_status() {
...
-        --set values.gateways.istio-ingressgateway.enabled=true
-        --set values.gateways.istio-egressgateway.enabled=true
-        --set values.meshConfig.accessLogFile="/dev/stdout"
-        --set values.kiali.enabled=true
-        --set values.istiocoredns.enabled=true
-"
-
 istio_install() {
-    bin/istioctl manifest apply $manifest_args
+    cat <<EOF | bin/istioctl install -y -f -
```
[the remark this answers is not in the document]
hunchback: currently will leave as is (as there is also .mk/tools.mk which istalls the istioctl binary into hack/tools/bin and the travis caching done on it ... so I rather leave all the related scripts to installation here (even if some are for 3rd party components not used in all scenarios)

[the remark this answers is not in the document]
hunchback: ok, moved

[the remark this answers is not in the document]
hunchback: ok

third_party/istio/Makefile
hunchback: yes

hack/tools/create_istio.sh
hunchback: will remove

hack/tools/create_istio.sh
```diff
@@ -12,21 +12,26 @@ istio_status() {
...
+    egressGateways:
+    - name: istio-egressgateway
+      enabled: true
+    ingressGateways:
+    - name: istio-ingressgateway
+      enabled: true
+  meshConfig:
+    accessLogFile: /dev/stdout
+  namespace: istio-system
+EOF
```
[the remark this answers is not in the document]
hunchback: ok

roee88 (approved):
I would set timeout to 5m because CI is usually slow. But we can keep 2m and change in a followup PR if we encounter timeout issues.

PR #165, opened 2020-11-25
Initial support for control-plane security

revit13:
This patch address issue #133 and it is the first step towards securing the traffic between the components in m4d-system namespace . 
RFC PR https://github.com/IBM/the-mesh-for-data/pull/152 was posted for that issue and this PR addresses the comments received.
It adds the following features using Istio:

    - Enforce connectors ingress traffic to be only from the manager and other connectors if needed.
    - Mutual TLS is enabled for the traffic between the manager and the connectors.


revit13:
This pr addresses comments given in https://github.com/IBM/the-mesh-for-data/pull/152.

revit13:
@roee88 The last changes applied to the patch address review comments. In particular Istio policies that are applied to secure the control-plane do not require further connector/workload specific adjustments. Thanks

roee88 (commented):
I think it would be better to have this as a kustomize layer in manager rather than a separate root level directory. Anyway, I would welcome an overview of alternative approaches to achieve the same as we discussed.

revit13:
This last changes address the comment regarding having the policies as kustomize layer in the manager by adding two new configs to the manager:
1) A new network-policy configuration which is applied in the default deployment and applies kubernetes networkpolicy for policy authorization.
2) A control-plane security configuration  which in addition to the above config it also applies istio policies.

Please note that Kind cluster default CNI did not work for me and I had to replace it with Calico CNI.
Tested on Kind.

manager/config/network-policies/authorization.yaml
```diff
@@ -0,0 +1,15 @@
...
+  podSelector:
+    matchLabels:
+      m4d.ibm.com/componentType: connector
+  policyTypes:
+  - Ingress
+  ingress:
+  - from:
+    - namespaceSelector:
+        matchLabels:
+          m4d.com/namespaceType: control-plane
```
roee88, line 15: https://github.com/ahmetb/kubernetes-network-policy-recipes/blob/master/04-deny-traffic-from-other-namespaces.md

revit13: Thanks! I fixed the policy.

manager/Makefile
```diff
@@ -94,6 +94,15 @@ undeploy_it:
 	cd config/manager && $(ABSTOOLBIN)/kustomize edit set image controller=${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/manager:${DOCKER_TAGNAME}
 	$(TOOLBIN)/kustomize build --load_restrictor none config/integration-tests | $(TOOLBIN)/kubectl delete -f -
 
+deploy_cps: $(TOOLBIN)/kustomize $(TOOLBIN)/kubectl
```
roee88, line 97: not a fan of shortcuts like `cps`, make has auto completion.

manager/config/control-plane-security/authentication.yaml
```diff
@@ -0,0 +1,12 @@
+# Copyright 2020 IBM Corp.
+# SPDX-License-Identifier: Apache-2.0
+
+apiVersion: "security.istio.io/v1beta1"
+# This policy enables mutual TLS traffic between every two workloads that have sidecar.
+kind: "PeerAuthentication"
+metadata:
+  name: "premissive-mtls-in-control-plane"
```
roee88, line 8: I see that other kustomize resources specify `namespace: system` (which is later altered automatically with the added prefix `m4d-`). This is relevant to all of the resources you added.

manager/config/control-plane-security/kustomization.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+# Value of this field is prepended to the
+# names of all resources, e.g. a deployment named
+# "wordpress" becomes "alices-wordpress".
+# Note that it should also match with the prefix (text before '-') of the namespace
+# field above.
+namePrefix: m4d-
+
+resources:
+- ../network-policies
```
roee88, line 15: If it's part of default then why is it added here?

manager/config/network-policies/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+# Copyright 2020 IBM Corp.
+# SPDX-License-Identifier: Apache-2.0
+
+resources:
+- authorization.yaml
```
roee88, line 5: I suggest renaming `authorization.yaml` to `connectors-ingress.yaml` to give it the right context (the fact that it's in `network-policies` folder fills the rest of the context)

website/content/docs/overview/architecture/index.md
```diff
@@ -30,4 +30,12 @@ The information below is outdated
 
 {{< bullet n=8 >}} Processing jobs for preparing the environment for the application are running. The figure shows an example of creating an implicit copy job to copy data from PostgresSQL to Ceph.
 
-{{< bullet n=9 >}} The application reads from and writes data to whitelisted targets. Requests are handled by `M4DModule` instances. The application can not interact with non-whitelisted targets.
\ No newline at end of file
+{{< bullet n=9 >}} The application reads from and writes data to whitelisted targets. Requests are handled by `M4DModule` instances. The application can not interact with non-whitelisted targets.
+
+### Control Plane Security
```
roee88, line 35: To match the documentation guidelines: 

```suggestion
### Control plane security
```

website/content/docs/overview/architecture/index.md
```diff
@@ -30,4 +30,12 @@ The information below is outdated
...
+
+The traffic between the pilot and the connectors is secured by lavergaing the following Kubernetes and Istio security features:
+- Istio authentication policy(\*) enables mutual TLS between the pilot and the connectors.
+- Kubernetes authorization policy controls the ingress traffic of the connectors.
+
+(\*) Applying Istio authentication policy is not part of the default {{< name >}} deployment.
```
roee88, line 41: Keeping `pilot` because this all page is outdated:
```suggestion
- A Kubernetes `NetworkPolicy` controls ingress traffic to connectors.
- Istio enables mutual TLS between the pilot and the connectors. Note that this is not part of the default {{< name >}} deployment and needs to be enabled explicitly.
```

website/content/docs/overview/connectors/index.md
```diff
@@ -48,3 +48,8 @@ A PDP returns a list of enforcement actions given a set of policies and specific
 
 Policies are therefore defined externally in the policy manager of choice. {{< name >}} provides a package to help writing data policies in OPA. Otherwise, data stewards are expected to keep using the policy manager that they already use, as long as there is a connector to it.
 
+# Securing the Connector Ingress Traffic
```
roee88, line 51: Please read the contribution guidelines for documentation
```suggestion
# Securing the connector ingress traffic
```

website/content/docs/overview/connectors/index.md
```diff
@@ -48,3 +48,8 @@ A PDP returns a list of enforcement actions given a set of policies and specific
...
+
+[Kubernetes NetworkPolicy](https://kubernetes.io/docs/concepts/services-networking/network-policies/) resource is used to ensure that the connector ingress is limited to traffic only from workloads within the control-plane namespace.
```
roee88, line 53: style: Kubernetes [`NetworkPolicy`](https://kubernetes.io/docs/concepts/services-networking/network-policies/)

website/content/docs/setup/quickstart/index.md
```diff
@@ -11,6 +11,7 @@ Ensure that you have the following:
 - `make`
 - `kubectl` version 1.16 and above
 - `helm` version 3.3 and above
+- `Istio` version 1.6.1 and above (Istio is only required if the manager control-plane-security configuration is used. Please see the manager README [config section](https://{{< github_base >}}/{{< github_repo >}}/tree/master/manager#config)).
```
roee88, line 14: It might be better for the quickstart to exclude Istio (it actually does exclude it, nothing in here requires Istio)

website/content/docs/setup/control-plane-security/index.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Istio installation notes
+
+If control-plane security configuration is used the Envoy sidecar is automatically injected to pods in the control-plane namespace (called 4d-system be default). To support automatic injection Istio's [MutatingAdmissionWebhook](https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#automatic-sidecar-injection) should be turned on in the cluster.
+
+To check if MutatingAdmissionWebhook is turned on in the cluster run:
+
+```bash
+kube-apiserver -h | grep enable-admission-plugins
+```
```
roee88, line 23: Please remove. We don't assume people have `kube-apiserver`

website/content/docs/setup/control-plane-security/index.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Steps to secure the Ingress Traffic of a New Connector
+
+Upon {{< name >}} deployment a new NetworkPolicy CRD is applied. This policiy
+allows ingress traffic to any connector labeled ```m4d.ibm.com/componentType: connector``` only from workloads within the control-plane namespace.
+
+Given that, a new label:```m4d.ibm.com/componentType: connector``` should be added to the components of a new connector (pods, service, deployment).
```
roee88, line 30: only `Pods` are required.

website/content/docs/setup/control-plane-security/index.md
```diff
@@ -0,0 +1,30 @@
+---
+title: Control-plane Security
+linktitle: Control-plane Security
+weight: 30
+---
+
+Upon {{< name >}} deployment [Kubernetes NetworkPolicy](https://kubernetes.io/docs/concepts/services-networking/network-policies/) resource is applied to ensure that the connector ingress is limited to traffic only from workloads within the control-plane namespace. 
+In addition, Istio authentication policy can be applied upon installation of {{< name >}} to ensure mutual TLS between the pilot and connectors. 
+Istio authentication policy is not part of the default {{< name >}} deployment but it can be applied by running the manager control-plane-security configuration as described in the manager README [config section](https://{{< github_base >}}/{{< github_repo >}}/tree/master/manager#config).
```
roee88, line 9: Please add concrete steps in a bash block for deploying these Istio stuff. 
```bash
command 1
command 1
```

Also, it's not an authorization policy. Just say Istio resources.

roee88 (changes requested):
Can you check if it would make sense to have `deploy-manager` and `undeploy-manager` that unifies all of the different `deploy_*` targets? For example, by getting the kustomize profile to apply as a make variable

manager/Makefile
revit13: OK

manager/config/control-plane-security/authentication.yaml
revit13: Sure will add it. I used namespace: m4d-system in kustomization.yaml and missed it. Will change that.

manager/config/control-plane-security/kustomization.yaml
revit13: I thought it should include it, but thinking on it again I'll remove it. tnx

revit13:
@roee88 Thanks for the review! I made fixes based on your comments. Regarding changing the manager's Makefile to use unified deploy-manager: I am checking that now. If the change will make sense I will address it in another pr (and issue) if that's OK with you. Thanks

roee88:
> If the change will make sense I will address it in another pr (and issue) if that's OK with you. Thanks

Sure! Thanks for working on this PR, I will re-review soon.

manager/config/control-plane-security/authentication.yaml
```diff
@@ -0,0 +1,13 @@
...
+  namespace: system
+spec:
+  mtls:
+    # PERMISSIVE: Workloads accept both mutual TLS and plain text traffic
+    mode: PERMISSIVE
```
roee88, line 13: Don't we need another `PeerAuthentication` with `STRICT` mode targeting connectors workloads?

website/content/docs/setup/control-plane-security/index.md
```diff
@@ -0,0 +1,34 @@
...
+This is optional and must be explicitly enabled. Before you begin ensure that:
+1. Istio 1.6 or above is installed
+1. Istio [automatic sidecar injection](https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#automatic-sidecar-injection) is enabled
+1. Istio [automatic mutual TLS](https://istio.io/latest/docs/tasks/security/authentication/authn-policy/#auto-mutual-tls) is enabled (this is the default)
+
+Before installing {{< name >}} run the following:
+```bash
+kubectl create namespace m4d-system
+kubectl label namespace m4d-system istio-injection=enabled
+make -C manager deploy_control_plane_security
```
roee88, line 27: I kept this but eventually (not now) we should remove the need for `make`.

roee88 (approved):
I added two last comments and also edited the website content. Please check.

manager/config/control-plane-security/authentication.yaml
revit13: We can however because the manager and other connectors have sidecars Istio auto mTLS feature guarantees that the mTLS is used for the connectors.

revit13: It seems that PeerAuthentication  STRICT mode is not needed in our case because of Istio auto mTL feature which automatically enables mTLS between all workloads with sidecar (as in our case - the incoming services to the connectors have sidecar). STRICT mode is needed to prevent non mTLS traffic when not all services have sidecar.  For this case destination rule should also be used.

roee88: You explanation is actually for why a destination rule is not needed. Destination rules are applied to the source (cf. destination) and can dictate whether an outgoing connection needs to be with mTLS. I agree that this is not needed here.

However, my understanding is that PeerAuthentication applies to the destination and set what it's willing to accept. Even if our sources are assumed to be configured to initial mTLS (via destination rules / default), the destination should still be set to only accept mTLS, even for just the case of misconfiguration.

PR #169, opened 2020-11-25
Omit svc.cluster.local suffix for services within the cluster

revit13:

### DIFF
--- a/manager/config/prod/deployment_configmap.yaml
+++ b/manager/config/prod/deployment_configmap.yaml
@@ -6,7 +6,7 @@ kind: ConfigMap
 metadata:
   name: m4d-config
 data:
-  SECRET_PROVIDER_URL: "http://secret-provider.m4d-system.svc.cluster.local:5555/get-secret"
+  SECRET_PROVIDER_URL: "http://secret-provider:5555/get-secret"
   SECRET_PROVIDER_ROLE: "demo"
   #catalog connector
   CATALOG_CONNECTOR_URL: "egr-connector:50084"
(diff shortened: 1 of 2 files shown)


manager/config/prod/deployment_configmap.yaml
```diff
@@ -6,7 +6,7 @@ kind: ConfigMap
 metadata:
   name: m4d-config
 data:
-  SECRET_PROVIDER_URL: "http://secret-provider.m4d-system.svc.cluster.local:5555/get-secret"
+  SECRET_PROVIDER_URL: "http://secret-provider:5555/get-secret"
```
roee88, line 9: This won't work, it is passed to modules that run in different namespace

roee88 (changes requested):
Added comment in secret provider.
Also, have you tested it on kind/OpenShift?

manager/config/prod/deployment_configmap.yaml
revit13: I tested it on Kind. The suffixes  in /etc/resolve are: m4d-system.svc.cluster.local svc.cluster.local cluster.local
That's why I omitted the m4d-system

revit13: I validated and the arrow module did not fail... maybe the address is cached somehow? is there a way to look at the logs? thanks

revit13: I've added the m4d-system to secret-provider. Thanks

roee88 (approved):

PR #173, opened 2020-12-01
Use python slim image for secret provider

froesef:

Hi @tomersolomon1! Can we use the slim python image? This would speed up builds quite a bit as the python image is currently more than 500MB ...


tomersolomon1 (approved):
Seems to be working, good idea 👍

Issue #144, opened 2020-11-19
Golang API for Multi-Cluster Env Info


simanadler:
|   |   |
| - | - |
| Design doc | [link] |
| Parent issue | #134 |
| Required issues | # |

# Description

The M4DAppliction controller currently creates a Blueprint that spans a single cluster. It now needs to deploy to multiple clusters.  This issue is to address the additions required to inform M4DApplication what clusters are available and in what geography each cluster runs.  In the future other metadata about the cluster may also be required.

# Requirements

1. Golang API that returns a list containing a unique cluster identifier and/or cluster name, and metadata associated with the cluster.  Metadata must include geography.
2. This API should work in single and multi-cluster M4D deployments

roee88:
```go
type ClusterMetadata struct {
  Region string
  Zone string
}

type Cluster struct {
  Name string
  Metadata ClusterMetadata
}

type ClusterManager interface {
  List() ([]Cluster, error)
}
```
Would this suffice as a starting point? @simanadler 

Only callable from coordinator cluster, no differentiation between coordinator/remote because it just lists clusters you can deploy to. Only returns clusters with metadata (cluster registration is considered as completed only when metadata is available to coordinator). Error is when we can't reach RazeeDash API. Currently using region and zone until 145 is resolved. 

@tomersolomon1 is this easily implementable?

tomersolomon1:
Under the assumption Razee can copy a configmap from remote cluster to local cluster it should be easy to implement. Personally I haven't tried it yet, but I understand @roee88 @froesef  it's possible? Under this my idea is the follows:
1. Create a configmap holding this data in each remote cluster (during cluster registration)
2. Propagate the configmap to the coordinator cluster (using the watchkeeper)
3.  The manager goes over all these configmaps

If (2) is not possible, we can pull that configmap from the razeedash-api using graphql.
Alternatively, after connecting the remote cluster, just create that configmap in the coordinator cluster.

I'm wondering about the mapping between the `ClusterGroup` name of the cluster and the cluster-name. On one hand, we currently put only one cluster per cluster group. So it makes sense they'll have the same name or be something like `<cluster-name>-group` (like just adding a suffix).
On the other hand, maybe in the future we'd like to have multiple cluster in the same cluster group? In that case, maybe adding a field with the cluster group name to that configmap is the right way to go?
[...]

Issue #136, opened 2020-11-15
Multicluster blueprint orchestration


roee88:
|   |   |
| - | - |
| Design doc | [link] |
| Parent issue | #134 |
| Required issues | # |

# Description

The M4DAppliction controller currently creates a Blueprint that spans a single cluster. This issue is to address the changes required to support managing blueprints in multiple clusters. Taking the separation of concerns approach, this issue describes a control plane component (can even be just a go package) that handles orchestration logic. It shouldn't include logic for blueprint generation.

# Requirements

1. The input is a description of a multicluster blueprint
1. Using [Razee](https://razee.io/)
1. Create/Update/Delete in a remote cluster
1. Monitor status as updated by a remote cluster
1. The resources sent to remotes describe a blueprint rather than individual modules (Razee does not support deploying Helm charts)

roee88:
The implementation can either directly use Razee or use an abstraction layer for CRUD-like operations on remote resources. For such an abstraction, if desired, a separate issue should be opened (and added to the "Required issues" list in the header of this issue). I would personally go with something that directly uses Razee first and only later try to generalize, otherwise we may fall into performance pitfalls. We can start a discussion about it here.

roee88:
Following the existing codebase, namespace creation is also needed. The recommendations we got:
1. Watch Blueprints by type rather than using a label
2. Use a list of resources to create namespace+blueprint, need to validate first that namespace creation is synchronous in k8w api.

roee88:
@tomersolomon1 @froesef can you please list what's missing in the current implementation?

froesef:
After the latest change to a single namespace which is m4d-blueprints the deletion of blueprints works as expected. Closing..

PR #176, opened 2020-12-02
multicluster: Initial commit of ClusterManager interface

froesef:
Initial version of the ClusterManager interface


manager/controllers/app/multicluster/multi_cluster_manager.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/ibm/the-mesh-for-data/manager/apis/app/v1alpha1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+)
+
+type ClusterManager interface {
+	GetClusters() ([]Cluster, error)
+	GetBlueprint(cluster string, namespace string, name string) (*v1alpha1.Blueprint, error)
+	CreateBlueprint(cluster string, blueprint *v1alpha1.Blueprint) (DeploymentDetails, error)
+	UpdateBlueprint(cluster string, blueprint *v1alpha1.Blueprint) (DeploymentDetails, error)
```
roee88, line 14: I suggest to remove `DeploymentDetails` and only add something like it if and when needed

manager/controllers/app/multicluster/multi_cluster_manager.go
```diff
@@ -0,0 +1,75 @@
...
+	DeleteBlueprint(cluster string, namespace string, name string) error
+}
+
+type ClusterMetadata struct {
+	Region string
+	Zone   string
```
roee88, line 20: OK for this PR since it follows the issue. But be aware that some clusters can span across multiple zones and we will need to deal with that later.

manager/controllers/app/multicluster/multi_cluster_manager.go
```diff
@@ -0,0 +1,75 @@
+package multicluster
```
roee88, line 1: I suggest to start moving things to the root level `/pkg`, starting with new stuff like multicluster.

manager/controllers/app/multicluster/multi_cluster_manager.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"fmt"
+	"github.com/ibm/the-mesh-for-data/manager/apis/app/v1alpha1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+)
+
+type ClusterManager interface {
+	GetClusters() ([]Cluster, error)
```
roee88, line 11: I thought that you suggested having a separate interface for GetClusters(). I don't mind if we don't split but just asking if that's really what you decided eventually. 

```go
type ClusterLister interface {
  GetClusters() ([]Cluster, error)
}
type ClusterManager interface {
  ClusterLister 
  GetBlueprint(cluster string, namespace string, name string) (*v1alpha1.Blueprint, error)
  ...
}

**Review by tomersolomon1 — APPROVED** [review]

File: manager/controllers/app/multicluster/multi_cluster_manager.go [hunk 5]
```diff
@@ -0,0 +1,75 @@
...
+
+type Cluster struct {
+	Name     string
+	Metadata ClusterMetadata
+}
+
+type DeploymentDetails struct {
+}
+
+type DummyMultiClusterManager struct {
```
roee88, line 31: If this is just for testing then move it to `*_test.go`

manager/controllers/app/multicluster/multi_cluster_manager_test.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"github.com/onsi/gomega"
+	apps "k8s.io/api/apps/v1"
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"testing"
+)
+
+func TestMultiCluster(t *testing.T) {
```
roee88, line 12: Are we testing decode here? Name the function accordingly.

manager/controllers/app/multicluster/multi_cluster_manager.go
roee88, line 14: Do we really need both Create and Update? (vs something that is similar to apply)

froesef: This is an initial version. I think in future it makes sense to merge it once we know how razee really behaves.

roee88: NP, you will definitely get to it when you implement the controller anyway

simanadler:
@froesef Which issue/design doc does this address? Can you include a link?  Thanks!

manager/controllers/app/multicluster/multi_cluster_manager.go
tomersolomon1: I don't mind, @froesef do you see a reason not to do that?

roee88:
@froesef please provide more details in the text of this PR. Specifically, what components will use this interface and for what. We don't have an issue linked to this PR, it seems to address https://github.com/IBM/the-mesh-for-data/issues/144 but also parts being part of https://github.com/IBM/the-mesh-for-data/issues/136

shlomitk1 (approved):

roee88 (approved):

PR #178, opened 2020-12-02
generation of multiple blueprint specifications

shlomitk1:

In addition to the module selection, decision is made in what geography the module will run based on user requirements and the existing set of M4DModules. For each geography, an appropriate blueprint specification will be generated.
This is an initial version that does not work with cluster metadata information. Thus, the blueprints/modules are generated per a geography and not a specific cluster.


simanadler:
@shlomitk1 I understand this is a first step, and thus you simplified the problem to distributing per geography.  However, is there an underlying assumption that there will be only one cluster per geography? If so, I don't think this is something we should assume.  Perhaps starting with the assumption that the clusterID=geography and coding using clusterID would result in less re-write in the next PR?

manager/controllers/app/modules/moduleinterface.go
```diff
@@ -70,90 +69,116 @@ func (m *Selector) GetError() string {
...
+	geo := m.selectGeography(item)
 	// append moduleinstances to the list
 	instances = append(instances, ModuleInstanceSpec{
-		AssetID: item.AssetID,
-		Module:  m.GetModule(),
-		Args:    args,
+		AssetID:   item.AssetID,
+		Module:    m.GetModule(),
+		Args:      args,
+		Geography: geo,
```
simanadler, line 78: Is this meant to remain geo in the future or will it change to cluster ID?

manager/controllers/app/modules/moduleinterface.go
```diff
@@ -70,90 +69,116 @@ func (m *Selector) GetError() string {
...
 	})
 	for _, dep := range m.GetDependencies() {
 		instances = append(instances, ModuleInstanceSpec{
-			AssetID: item.AssetID,
-			Module:  dep,
-			Args:    args,
+			AssetID:   item.AssetID,
+			Module:    dep,
+			Args:      args,
+			Geography: geo,
```
simanadler, line 85: Same as previous comment

simanadler (commented):
See comment I put in PR discussion

shlomitk1:
> 
> 
> @shlomitk1 I understand this is a first step, and thus you simplified the problem to distributing per geography. However, is there an underlying assumption that there will be only one cluster per geography? If so, I don't think this is something we should assume. Perhaps starting with the assumption that the clusterID=geography and coding using clusterID would result in less re-write in the next PR?

This is the first step. There will be a functionality of determining clusterID based on selected geography. Also, processingGeography in m4dapplication will be replaced by workloadClusterID. Assuming N >= 1 clusters per geography.

manager/controllers/app/modules/moduleinterface.go
shlomitk1: Not sure. A clusterID as the first step. Perhaps, to extend to a list of possible clusterIDs to make it possible to decide later on clusterID that is common to many modules. I would start with a single clusterID.

manager/controllers/app/modules/moduleinterface.go
shlomitk1: CluserID as well.

simanadler (approved):
Approving as a first step

PR #180, opened 2020-12-03
local cluster manager impl

tomersolomon1:
Implementation of the local cluster manager (which retrieves the data from a configmap in the `m4d-system`)

### DIFF
--- a/pkg/multicluster/local/local_cluster_manager.go
+++ b/pkg/multicluster/local/local_cluster_manager.go
@@ -13,11 +13,11 @@ const (
 	clusterMetadataConfigmapName string = "cluster-metadata"
 )
 
-type LocalClusterManager struct {
+type ClusterManager struct {
 	Client client.Client
 }
 
-func (cm *LocalClusterManager) GetClusters() ([]multicluster.Cluster, error) {
+func (cm *ClusterManager) GetClusters() ([]multicluster.Cluster, error) {
 	clusterMetadataConfigmap := corev1.ConfigMap{}
 	namespacedName := client.ObjectKey{
 		Name:      clusterMetadataConfigmapName,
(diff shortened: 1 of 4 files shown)


pkg/multicluster/local/local_cluster_manager.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"context"
+	"github.com/ibm/the-mesh-for-data/manager/apis/app/v1alpha1"
+	"github.com/ibm/the-mesh-for-data/pkg/multicluster"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	m4dSystemNs                  string = "m4d-system"
```
roee88, line 12: change to a `namespace` parameter to the LocalClusterManager (because I think that we can eventually detect it, but definitely not part of the scope of local_cluster_manager)

pkg/multicluster/local/local_cluster_manager.go
```diff
@@ -0,0 +1,61 @@
...
+
+func (cm *LocalClusterManager) UpdateBlueprint(cluster string, blueprint *v1alpha1.Blueprint) error {
+	return nil
+}
+
+func (cm *LocalClusterManager) DeleteBlueprint(cluster string, namespace string, name string) error {
+	return nil
+}
+
+func CreateLocalClusterManager(client client.Client) multicluster.ClusterManager {
```
roee88, line 57: `NewLocalClusterManager`

pkg/multicluster/local/local_cluster_manager.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (cm *LocalClusterManager) GetClusters() ([]multicluster.Cluster, error) {
+	clusterMetadataConfigmap := corev1.ConfigMap{}
+	namespacedName := client.ObjectKey{
+		Name:      clusterMetadataConfigmapName,
+		Namespace: m4dSystemNs,
+	}
+	if err := cm.Client.Get(context.Background(), namespacedName, &clusterMetadataConfigmap); err != nil {
+		return nil, err
```
roee88, line 27: wrap the error and provide more context about the error

pkg/multicluster/local/local_cluster_manager_test.go
```diff
@@ -0,0 +1,49 @@
...
+			},
+		},
+	}
+	cl := fake.NewFakeClientWithScheme(s, objs...)
+
+	cm := CreateLocalClusterManager(cl)
+	var actualClusters []multicluster.Cluster
+	var err error
+	if actualClusters, err = cm.GetClusters(); err != nil {
+		t.Error("TestMultiCluster failed")
```
roee88, line 36: Too generic. Consider something like `t.Errorf("unexpected error in GetClusters: %v", err)`

pkg/multicluster/local/local_cluster_manager.go
```diff
@@ -0,0 +1,61 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	m4dSystemNs                  string = "m4d-system"
+	clusterMetadataConfigmapName string = "cluster-metadata"
+)
+
+type ClusterManager struct {
```
roee88, line 16: ClusterManager is already the interface name

tomersolomon1: I tried before to call it `LocalClusterManager`, but Go's linter didn't like it and called it `stuttering`, and suggested to call it just `ClusterManager`. Bellow is a SOF question about it.https://stackoverflow.com/questions/45885985/ways-to-avoid-stuttering-in-go-package-and-struct-names

tomersolomon1: I agree we need to be consistent. Another option is to call the interface `ClusterManagerInterface` or `IClusterManager`. @froesef  What do you think?

By the way, a name is considered to be stuttering if the package name is a strict prefix of it:
https://github.com/golang/lint/blob/master/lint.go#L944

froesef: I was running into the same issue. I called it `MultiClusterManager` in dummy package. But I'm fine with `ClusterManager` and then the package name would define which one it is.

tomersolomon1: @froesef I changed it to be `ClusterManager` in both implementations

roee88 (approved):
@froesef @tomersolomon1 

In a follow-up please add to the tests a check that the implementation matches the interface.

```go 
var _ InterfaceName = &ImplementationName{}
```

PR #181, opened 2020-12-06
Removed minikube related scripts and targets

### LINKED ISSUE #150: codebase: remove minikube related targets
If someone wants to use minikube then we have nothing against it, but our codebase should not include minikube related stuff as our automated testing uses kind. Also, minikube setup is different depending on os and external dependencies, the existing targets didn't work for all anyway.

roee88:
Fix #150

If anyone wants to use minikube we still allow it, but we don't support scripts that we don't use in the CI.
Contributors should use Kind instead.


rohithdv (approved):
I think the connectors codebase is now not using minikube as we used to earlier (with exclusive script and yamls for minikube). So the changes done in this PR should be fine.

PR #185, opened 2020-12-07
Razee multicluster impl no tests

tomersolomon1:
Razee cluster-manager implementation. This PR doesn't include tests. Tests require pushing stuff to different cluster, and will be handled in future PR.


pkg/multicluster/razee/razee.go
```diff
@@ -0,0 +1,350 @@
...
+		return "", err
+	}
+	return result.AddGroup.Uuid, nil
+}
+
+//nolint:golint,unused
+type IAMRazeeClient struct {
+	apiKey string
+	client *graphql.Client
+}
```
roee88, line 189: Not needed

pkg/multicluster/razee/razee_cluster_manager.go
```diff
@@ -0,0 +1,241 @@
...
+	return nil
+}
+
+//nolint:golint,unused
+func channelName(cluster string, namespace string, name string) string {
+	return "m4d-" + cluster + "-blueprint-" + namespace + "-" + name
+}
+
+//nolint:golint,unused
+func NewRazeeManager(url string, login string, password string, orgId string) multicluster.ClusterManager {
```
roee88, line 222: Shouldn't this accept a RazeeClient instead?

pkg/multicluster/razee/razee_cluster_manager.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	if removeChannel.Success {
+		r.log.Info("Successfully deleted channel " + removeChannel.UUID)
+	}
+	return nil
+}
+
+//nolint:golint,unused
+func channelName(cluster string, namespace string, name string) string {
+	return "m4d-" + cluster + "-blueprint-" + namespace + "-" + name
```
roee88, line 218: can we swap to `m4d.ibm.com/<cluster>/<namespace>/<name>`? or is it illegal?

roee88 (commented):
Thanks, looks good. @tomersolomon1 I did a first pass review so I will review the actual logic tomorrow and let @froesef know in case of any suspected issue.

pkg/multicluster/razee/razee_cluster_manager.go
tomersolomon1: Not sure I understand. Maybe you mean Round-Tripper?

pkg/multicluster/razee/razee_cluster_manager.go
tomersolomon1: Nothing is illegal in wonderland:
```
> db.channels.find()
{ "_id" : "a3fe3acb-155f-4fc2-90da-ba17fbb90855", "uuid" : "15b09186-5103-4fbe-bddd-e9ad785e7ebe", "org_id" : "e9077db1-f82f-4cdf-a0e6-06dc7528e0f6", "name" : "m4d.ibm.com/cluster-1/namespace-2/resource-3", "versions" : [ ], "created" : ISODate("2020-12-07T16:20:48.410Z"), "__v" : 0 }
> 
```

pkg/multicluster/razee/razee.go
tomersolomon1: Right

pkg/multicluster/razee/razee_cluster_manager.go
froesef: We have two different RazeeClients at the moment. Both of them have a different initialization. I would hide the Razee specific stuff behind these functions and not expose it further up.

roee88: What two clients? All of the New*Client functions return RazeeClient. 

My suggestion was that NewRazeeManager  accept a RazeeClient. Usage is for example:
`NewRazeeClient(NewGithubAPIKeyClient(...))` but you will be able to initiate it with Local or IAM the same way.

froesef: `con` and `razeeClient`. We currently use multiple clients to razee as the library does not support all functions yet. (The plan is to eventually implement everything in sat-con library and then have only that one)

froesef (approved):

pkg/multicluster/razee/razee_cluster_manager.go
roee88: Does `con` only supports local authentication?

roee88 (approved):
Approving to avoid blocking further work, but let's have a sync tomorrow because I do think this needs some quick changes in a follow-up.

Issue #135, opened 2020-11-15
Multicluster blueprint generation


roee88:
|   |   |
| - | - |
| Design doc | https://ibm.box.com/s/62xh8iwu64aas4awb2fwtad8betn1e07 |
| Parent issue | #134 |
| Required issues | # |

# Description

The M4DAppliction controller currently creates a blueprint that spans a single cluster. This issue is to address the changes required to support managing blueprints in multiple clusters. Taking the separation of concerns approach, this shouldn't include orchestration logic. 

# Requirements

1. Create a single description of a blueprint that spans multiple clusters 
    - An auditable structure 
    - Describes desired state in all clusters
    - Not necessarily a CRD or persisted anywhere, though it can be if that is what will be decided
    - Future: would also be useful for picking an optimal blueprint from multiple choices, may be used to describe a cross-cluster data plane graph, etc.
1. Generation is based on:
    1. Input from the various connectors in the control plane (primary cluster)
    1. Registered M4DModules in the control plane (primary cluster)
    1. Registered remote clusters and the metadata of them: geo-location, cluster features (to check if a M4DModule is applicable in the cluster) - _current assumption: all modules installed in all clusters_
1. Generation is triggered by:
    1. A change in M4DApplication spec
    1. Once every X minutes to account for policy changes (future)
1. Keep track of actual state to reflect reconcile status
1. Efficiency
    1. No blueprint generation triggers other than what is listed here
    1. No operation if no changes to desired state are required
    1. Complete reconciliation as fast as possible (e.g., allow remotes to reconcile in parallel to each other)

roee88:
@shlomitk1 can this be closed now?

shlomitk1:
Closing.

PR #187, opened 2020-12-07
Using cluster environment metadata for Plotter generation

shlomitk1:
Related to https://github.com/IBM/the-mesh-for-data/issues/135
Blueprints in Plotter are mapped by cluster name rather than geography.
Initial interaction with ClusterLister interface inside m4dapplication controller:
- in main.go local ClusterManager is used
- unit tests: a mockup ClusterLister is used
- integration in kind with existing cluster: a configmap is created to be used by local ClusterManager.


manager/controllers/app/suite_test.go
```diff
@@ -75,6 +75,17 @@ var _ = BeforeSuite(func(done Done) {
 	if os.Getenv("USE_EXISTING_CONTROLLER") == "true" {
 		logf.Log.Info("Using existing controller in existing cluster...")
 		k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
+		Expect(k8sClient.Create(context.Background(), &v1.ConfigMap{
```
froesef, line 78: This could be part of the environment setup in line 57. Basically another K8s object in the testdata folder that is also loaded. But we can sort the exact testing environment etc out later so this is just a note here. (Then you also would not need the mockup listener)

froesef (approved):

manager/controllers/app/suite_test.go
shlomitk1: This is very immature. For an existing controller, for example, both cases should be considered (multi/single cluster config). This is an initial version that will make CI pass in a single-cluster environment, but more work should be done to guarantee an effective testing.

tomersolomon1 (approved):

PR #189, opened 2020-12-07
Multicluster: local razee secret

froesef:
This PR improves the local development environment that is used for multi cluster development.
It creates a secret with razee credentials that can be mounted to a pod in order to connect with razee.
Currently the automated setup supports razee local as authentication for local development.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,6 +87,10 @@ docker-build-local:
 helm:
 	$(MAKE) -C modules helm
 
+.PHONY: helm-chart-push
+helm-chart-push:
+	$(MAKE) -C modules helm-chart-push
+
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -87,6 +87,10 @@ docker-build-local:
 helm:
 	$(MAKE) -C modules helm
 
+.PHONY: helm-chart-push
+helm-chart-push:
+	$(MAKE) -C modules helm-chart-push
```
roee88, line 92: Where is this target used? I couldn't find usage

roee88 (approved):

Makefile
froesef: Good point. I thought I did but I use the target from modules directly. I'll remove it.

PR #194, opened 2020-12-09
multicluster: Plotter controller

froesef:
This PR introduces a first version of a plotter. 
I'd be happy about some first feedback and comments over what really needs to be solved in this PR and what might be follow up issues.

Currently it will create the blueprints in a remote namespace and watch their status.
If updates occur on the plotter object these will also be updated in the remote blueprints.

Pat of #136.


manager/controllers/app/plotter_controller.go
```diff
@@ -0,0 +1,246 @@
...
+		} else {
+			random := rand.String(5)
+			randomNamespace := "m4d-" + random
+			blueprint := &app.Blueprint{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "Blueprint",
+					APIVersion: "app.m4d.ibm.com/v1alpha1",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        "blueprint",
```
shlomitk1, line 175: Do we want to match the plotter name, for an easier debug (in case more than one plotter is active)?

shlomitk1 (approved):

manager/config/multi-controller/kustomization.yaml
```diff
@@ -0,0 +1,21 @@
...
+- manager_patch.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: data-catalog-mock
+  newName: kind-registry:5000/m4d-system/data-catalog-mock
+  newTag: latest
+- name: policycompiler
+  newName: kind-registry:5000/m4d-system/serverpolicycompiler-mock
+  newTag: latest
```
roee88, line 21: Why do we need this? I don't think that these are used in the default profile

manager/config/multi-controller/manager_patch.yaml
```diff
@@ -0,0 +1,30 @@
+# Copyright 2020 IBM Corp.
```
roee88, line 1: Can be move multi cluster configurations to a subfolder of `config`?

- `manager/config/multicluster/coordinator` for this file
- `manager/config/multicluster/remote` for what you called follower

manager/config/multi-controller/kustomization.yaml
roee88: same for other files

manager/controllers/app/plotter_controller.go
```diff
@@ -0,0 +1,246 @@
...
+	Log            logr.Logger
+	Scheme         *runtime.Scheme
+	ClusterManager multicluster.ClusterManager
+}
+
+// +kubebuilder:rbac:groups=app.m4d.ibm.com,resources=plotters,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=app.m4d.ibm.com,resources=plotters/status,verbs=get;update;patch
+
+// Reconcile receives a Plotter CRD
+//nolint:dupl
```
roee88, line 38: [out of scope feedback] I have some code for a reusable ReconcileCycle that can use instead of duplicating (we can create a variation of it based on our common code). I will open a separate issue.

manager/controllers/app/plotter_controller.go
```diff
@@ -0,0 +1,246 @@
...
+	hasFinalizer := ctrlutil.ContainsFinalizer(plotter, finalizerName)
+
+	// If the object has a scheduled deletion time, delete it and its associated resources
+	if !plotter.DeletionTimestamp.IsZero() {
+		// The object is being deleted
+		if hasFinalizer { // Finalizer was created when the object was created
+			// the finalizer is present - delete the allocated resources
+
+			for cluster, blueprint := range plotter.Status.Blueprints {
+				// TODO Check namespace deletion. Some finalizers leave namespaces in terminating state
```
roee88, line 91: [out of scope feedback] let's ensure that we code it only once

manager/controllers/app/plotter_controller.go
```diff
@@ -0,0 +1,246 @@
...
+			if !reflect.DeepEqual(blueprintSpec, remoteBlueprint.Spec) {
+				r.Log.V(1).Info("Blueprint specs differ...")
+				remoteBlueprint.Spec = blueprintSpec
+				remoteBlueprint.ObjectMeta.Annotations = map[string]string(nil) // reset annotations
+				err := r.ClusterManager.UpdateBlueprint(cluster, remoteBlueprint)
+				if err != nil {
+					return ctrl.Result{}, err
+				}
+				return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+			}
```
roee88, line 150: Wouldn't this trigger redundant updates between the time you push an update to the time its status is reflected in razeedash api?

manager/controllers/app/plotter_controller.go
```diff
@@ -0,0 +1,246 @@
...
+
+			if !reflect.DeepEqual(blueprintSpec, remoteBlueprint.Spec) {
+				r.Log.V(1).Info("Blueprint specs differ...")
+				remoteBlueprint.Spec = blueprintSpec
+				remoteBlueprint.ObjectMeta.Annotations = map[string]string(nil) // reset annotations
+				err := r.ClusterManager.UpdateBlueprint(cluster, remoteBlueprint)
+				if err != nil {
+					return ctrl.Result{}, err
+				}
+				return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
```
roee88, line 149: Not very effective to requeue inside the iteration (in case we have multiple blueprints)

manager/controllers/app/plotter_controller.go
```diff
@@ -0,0 +1,246 @@
...
+			blueprint := &app.Blueprint{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "Blueprint",
+					APIVersion: "app.m4d.ibm.com/v1alpha1",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        "blueprint",
+					Namespace:   randomNamespace,
+					ClusterName: cluster,
+					Labels:      map[string]string{"razee/watch-resource": "debug"},
```
roee88, line 178: [out of scope feedback] You use debug for getting the spec, I'm not sure that this is the right approach and I think that getting "what I pushed" would fit better. Leave the remote reconcile to Razee.

manager/main.go
```diff
@@ -186,3 +197,24 @@ func initVaultConnection() (*api.Client, error) {
...
+func NewClusterManager(mgr manager.Manager) multicluster.ClusterManager {
+	setupLog := ctrl.Log.WithName("setup")
+	if user, razeeLocal := os.LookupEnv("RAZEE_USER"); razeeLocal {
+		razeeURL := strings.TrimSpace(os.Getenv("RAZEE_URL"))
+		password := strings.TrimSpace(os.Getenv("RAZEE_PASSWORD"))
+		orgID := strings.TrimSpace(os.Getenv("RAZEE_ORG_ID"))
+
+		setupLog.Info("Using razee at " + razeeURL + " orgId " + orgID)
+		return razee.NewRazeeManager(strings.TrimSpace(razeeURL), strings.TrimSpace(user), password, orgID)
+	} else if apiKey, satConf := os.LookupEnv("IAM_API_KEY"); satConf {
```
roee88, line 212: Should rename to something more concrete (IAM is not a unique indicator, it's used everywhere).

manager/controllers/app/plotter_controller.go
roee88, line 178: We should avoid razee specific labels, it's the job of the cluster manager to add it.

manager/controllers/app/plotter_controller.go
```diff
@@ -0,0 +1,246 @@
...
+			plotter.Status.Blueprints[cluster] = blueprintMini
+			err := r.ClusterManager.CreateBlueprint(cluster, blueprint)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			isReady = false
+		}
+	}
+
+	// TODO do loop of statuses vs spec for removed specs
```
roee88, line 197: Here too I don't think it should be from statuses.

manager/controllers/app/plotter_controller.go
```diff
@@ -0,0 +1,246 @@
...
+
+	if isInitialReconcile {
+		// Return after initial deployment of blueprints
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+	}
```
roee88, line 202: Why?

roee88 (approved):
LGTM. All comments are for follow-ups.

PR #204, opened 2020-12-13
make: add run-integration-tests

hunchback:
Signed-off-by: hunchback <aidan.shribman@gmail.com>

can be used for local development by running:

```
make run-integration-tests
```

from top-level

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,8 +17,8 @@ test:
 	$(MAKE) -C manager test
 
 .PHONY: run-integration-tests
-run-integration-tests: export DOCKER_HOSTNAME=kind-registry:5000
-run-integration-tests: export DOCKER_NAMESPACE=m4d-system
+run-integration-tests: export DOCKER_HOSTNAME?=kind-registry:5000
+run-integration-tests: export DOCKER_NAMESPACE?=m4d-system
 run-integration-tests:
 	$(MAKE) kind
 	$(MAKE) cluster-prepare
--- a/manager/Makefile
+++ b/manager/Makefile
@@ -123,8 +123,8 @@ wait_for_manager:
 	$(TOOLBIN)/kubectl wait --for=condition=available -n ${KUBE_NAMESPACE} deployment/m4d-controller-manager --timeout=120s
 
 .PHONY: run-integration-tests
-run-integration-tests: export DOCKER_HOSTNAME=kind-registry:5000
-run-integration-tests: export DOCKER_NAMESPACE=m4d-system
+run-integration-tests: export DOCKER_HOSTNAME?=kind-registry:5000
+run-integration-tests: export DOCKER_NAMESPACE?=m4d-system
 run-integration-tests: wait_for_manager
 	NO_SIMULATED_PROGRESS=true USE_EXISTING_CONTROLLER=true USE_EXISTING_CLUSTER=true go test ./... -v -run TestMotionAPIs -count 1
 	NO_SIMULATED_PROGRESS=true USE_EXISTING_CONTROLLER=true USE_EXISTING_CLUSTER=true go test ./... -v -run TestAPIs -count 1


hunchback:
output of running tests:

```
aidan@AIDANSHRIBMAN-SDFUE6N:~/projects/the-mesh-for-data/manager$ make run-integration-tests
Makefile:69: warning: overriding recipe for target 'docker-build'
../hack/make-rules/docker.mk:18: warning: ignoring old recipe for target 'docker-build'
../hack/tools/bin/kubectl wait --for=condition=available -n m4d-system deployment/m4d-controller-manager --timeout=120s
deployment.apps/m4d-controller-manager condition met
NO_SIMULATED_PROGRESS=true USE_EXISTING_CONTROLLER=true USE_EXISTING_CLUSTER=true go test ./... -v -run TestMotionAPIs -count 1
?       github.com/ibm/the-mesh-for-data/manager        [no test files]
testing: warning: no tests to run
PASS
ok      github.com/ibm/the-mesh-for-data/manager/apis/app/v1alpha1      2.850s [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/ibm/the-mesh-for-data/manager/apis/motion/v1alpha1   0.027s [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/ibm/the-mesh-for-data/manager/controllers/app        0.048s [no tests to run]
?       github.com/ibm/the-mesh-for-data/manager/controllers/app/modules        [no test files]
?       github.com/ibm/the-mesh-for-data/manager/controllers/mockup     [no test files]
=== RUN   TestMotionAPIs
Running Suite: Controller Suite
===============================
Random Seed: 1607851590
Will run 2 of 2 specs

••

Ran 2 of 2 Specs in 4.030 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestMotionAPIs (4.03s)
PASS
ok      github.com/ibm/the-mesh-for-data/manager/controllers/motion     4.062s
?       github.com/ibm/the-mesh-for-data/manager/controllers/utils      [no test files]
?       github.com/ibm/the-mesh-for-data/manager/unittests      [no test files]
NO_SIMULATED_PROGRESS=true USE_EXISTING_CONTROLLER=true USE_EXISTING_CLUSTER=true go test ./... -v -run TestAPIs -count 1
?       github.com/ibm/the-mesh-for-data/manager        [no test files]
testing: warning: no tests to run
PASS
ok      github.com/ibm/the-mesh-for-data/manager/apis/app/v1alpha1      2.469s [no tests to run]
testing: warning: no tests to run
PASS
ok      github.com/ibm/the-mesh-for-data/manager/apis/motion/v1alpha1   0.055s [no tests to run]
=== RUN   TestAPIs
Running Suite: Controller Suite
===============================
Random Seed: 1607851598
Will run 9 of 9 specs

••••••••
------------------------------
• [SLOW TEST:5.320 seconds]
M4DApplication Controller
/home/aidan/projects/the-mesh-for-data/manager/controllers/app/m4dapplication_controller_test.go:151
  M4DApplication e2e
  /home/aidan/projects/the-mesh-for-data/manager/controllers/app/m4dapplication_controller_test.go:522
    Test end-to-end for M4DApplication with arrow-flight module
    /home/aidan/projects/the-mesh-for-data/manager/controllers/app/m4dapplication_controller_test.go:539
------------------------------


Ran 9 of 9 Specs in 13.076 seconds
SUCCESS! -- 9 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestAPIs (13.08s)
PASS
ok      github.com/ibm/the-mesh-for-data/manager/controllers/app        13.154s
?       github.com/ibm/the-mesh-for-data/manager/controllers/app/modules        [no test files]
?       github.com/ibm/the-mesh-for-data/manager/controllers/mockup     [no test files]
testing: warning: no tests to run
PASS
ok      github.com/ibm/the-mesh-for-data/manager/controllers/motion     0.012s [no tests to run]
?       github.com/ibm/the-mesh-for-data/manager/controllers/utils      [no test files]
?       github.com/ibm/the-mesh-for-data/manager/unittests      [no test files]
aidan@AIDANSHRIBMAN-SDFUE6N:~/projects/the-mesh-for-data/manager$
```

Makefile
```diff
@@ -16,6 +16,22 @@ test:
 	$(MAKE) -C pkg/policy-compiler test
 	$(MAKE) -C manager test
 
+.PHONY: run-integration-tests
```
froesef, line 19: I'm not sure if a bash script in `hack` would be better to run the test instead of putting more code into the make file.

ronenkat (commented):
@hunchback which issue discuss the use-case of running the integration tests?
I am not sure I understand when are we using this, and how this related to the CI.

Makefile
hunchback: this is an issue of taste, my personal take is that having a script from hack calling back into the make system creates a complex look (notice that all the other scripts are self-contained in that respect). so the natural place for such a sequence is in the makefile itself, as to its' length, this will reduce as we improve and current acceptable complexity of setting up m4d (I am working on that separately). for example: the `secret-provider` steps need to be merged into the m4d core deployment (and thus will go), I have a big problem with the architecture of test/services and think this code should be integrated as a mock code within the connectors (and thus will go), etc.

connectors/opa/go.mod
```diff
@@ -8,7 +8,7 @@ go 1.13
 
 require (
 	github.com/golang/protobuf v1.4.2
-	github.com/hashicorp/go-retryablehttp v0.5.4
+	github.com/hashicorp/go-retryablehttp v0.6.6
```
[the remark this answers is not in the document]
hunchback: agree, this should not be here, this was a result of building the opa connector previously, which indicates that the `go.mod` there is stale, but going forward we should probably eliminate the private go.mod and use the top-level go.mod. in any case will remove this change from this PR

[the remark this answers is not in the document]
hunchback: this is a result of build/deploying opa connector, its' a result of having stale go.mod dependencies in opa and elsewhere, so note that if I remove this (and other connector go.mod changes I introduced) then each time the user will run the test he will get all these changes locally ... which is not what we want ... we want the code to be updated so that anyone running the test will not trigger additional changes to the code base.

Makefile
roee88: I agree with @froesef that a script would be cleaner here, even if it's just a matter of taste. I also find the exports confusing, at least I wasn't aware of this kind of syntax.

manager/config/manager/kustomization.yaml
```diff
@@ -7,5 +7,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: ghcr.io/the-mesh-for-data/manager
+  newName: kind-registry:5000/m4d-system/manager
```
roee88, line 10: why the change?

hack/tools/create_kind.sh
```diff
@@ -97,7 +98,7 @@ case "$op" in
     cleanup)
         header_text "Uninstalling kind cluster"
         registry_delete || true
-        # certs_delete || true # Deactivated for now to simplify development when recreating a kind cluster
+        certs_delete || true
```
roee88, line 101: why did you reactivate? Not saying that you shouldn't, but needs to understand

Makefile
```diff
@@ -16,6 +16,22 @@ test:
...
+run-integration-tests: export DOCKER_HOSTNAME=kind-registry:5000
+run-integration-tests: export DOCKER_NAMESPACE=m4d-system
```
roee88, line 21: Why are these defined here?

Makefile
```diff
@@ -16,6 +16,22 @@ test:
...
+	$(MAKE) cluster-prepare
+	$(MAKE) docker
+	$(MAKE) -C test/services docker-all
+	$(MAKE) cluster-prepare-wait
+	$(MAKE) -C secret-provider configure-vault
+	$(MAKE) -C secret-provider deploy
+	$(MAKE) -C manager deploy_it
+	$(MAKE) -C manager wait_for_manager
+	$(MAKE) helm
+	$(MAKE) -C manager run-integration-tests
```
roee88, line 33: Does it match 1-1 the script you deleted? In a previous PR changes here made the CI fail when building master and I had to revert the PR

manager/config/manager/kustomization.yaml
hunchback: so that the change is stable when we run the integration-tests ... the root problem which should probably be addressed is that our use of kustomize is altering the file `kustomization.yaml`, so to fix it properly we need to rename the file to `kustomization.yaml.in` and make `kustomization.yaml` an ephemeral file (which has an entry in .gitignore) ... but as I am anyway working on a new helm system to replace eventually kustomize when we may not need this change in the end.

hack/tools/create_kind.sh
hunchback: this caused for a silent failure (which was very difficult to track down), as the `hack/registry` file was not refreshed causing the mounting of our local kind-cluster registry to have missing CA files, which in turn trigger the mysterious x503 errors which came and went. No (which this change) the issue will be detected during kind cluster creation and will be easy to detect and solve (solved in my case by rebooting the docker service which was holding the `hack/registry` directory and not allowing it to be recreated correctly)

Makefile
hunchback: as I want the default of `run-integration-tests` to work as expected using the local registry, while generally I want all other operations to work by default on the global registry (for example when building and pushing docker images or helm charts).

Makefile
hunchback: yes, it matches, and works, I did not change .travis yet as I don't want to destabilize the CI, but will do that later

manager/config/manager/kustomization.yaml
roee88: would it affect applying the default profile or are we setting the image there separately in a patch?

hack/tools/create_kind.sh
roee88: Got it, thanks.

Makefile
roee88: OK

hunchback:
this change was already merged in #42 (as it served for testing it), so the only change remaining is the ability to alter the env variables, and updating I will push now

roee88 (approved):

PR #206, opened 2020-12-13
manager: change chart resource multi to single

hunchback:
change array of resources (each a helm chart) to contain only a single value.


shlomitk1 (approved):

manager/apis/app/v1alpha1/blueprint_types.go
```diff
@@ -103,9 +103,9 @@ type ComponentTemplate struct {
 	// +required
 	Kind string `json:"kind"`
 
-	// Resources contains the location of the helm chart with info detailing how to deploy
+	// Resource contains the location of the helm chart with info detailing how to deploy
 	// +required
-	Resources []string `json:"resources"`
+	Resource string `json:"resource"`
```
roee88, line 108: Shouldn't we name this chart instead of resource?

roee88 (approved):

manager/apis/app/v1alpha1/blueprint_types.go
hunchback: will change

manager/config/manager/kustomization.yaml
```diff
@@ -7,5 +7,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: ghcr.io/the-mesh-for-data/manager
+  newName: kind-registry:5000/m4d-system/manager
```
shlomitk1, line 10: @hunchback Why was this line changed?

PR #207, opened 2020-12-13
Reset error in beginning of reconcile loop and check for existing cha…

froesef:
…nnels during creation

### DIFF
--- a/pkg/multicluster/razee/razee_cluster_manager.go
+++ b/pkg/multicluster/razee/razee_cluster_manager.go
@@ -16,6 +16,7 @@ import (
 	"net/http"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"strconv"
+	"strings"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/multicluster/razee/razee_cluster_manager.go
```diff
@@ -193,6 +194,19 @@ func (r *ClusterManager) CreateBlueprint(cluster string, blueprint *v1alpha1.Blu
 		return err
 	}
 
+	// Check if channel exists
+	existingChannel, err := r.con.Channels.ChannelByName(r.orgId, channelName)
+	if err != nil {
+		if !strings.HasPrefix(err.Error(), "Query channelByName error. Could not find the channel with name") {
```
roee88, line 200: We should check if Razee exposed error codes instead and if so open an issue in satcon to expose it. For now, maybe set this as a constant?

roee88 (approved):

pkg/multicluster/razee/razee_cluster_manager.go
froesef: I checked. Not found is actually not a 404 but a 200 with an error code. (returned from razee)

Issue #184, opened 2020-12-07
helm3 packaging for m4d


hunchback:
# Overview

To help simplify deployment of the-mesh-for-data we want the entire core (including connectors) to have a helm based deployment (optionally moving CRD into a separate deployment).

# Stage 0: Helm Chart: manager CRDs

the reason for have a separate helm chart for CRDs is:

- we may have several helm charts which depend on the CRDs requiring it be deployed independently
- we want to be able to upgrade the CRDs independently from core
- that this is an acceptable practice see rancher/fleet for example

so the result would be supporting:

```
helm install -g m4d-crd
```

This is how Rancher Fleet works: https://github.com/rancher/fleet/tree/master/charts/fleet-crd

And is defined as a plausible option (Method 2) here: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

# Stage 1: Helm Chart:  manager

The idea is to replace the current kustomize based manager deployment with a helm chart wrapper which enables to configure where the helm modules and module images are downloaded from (the docker registry location).

# Stage 2: Helm Chart: manager + connectors

Next we add connectors to the deployment: 
- opa-connector
- vault-connector
- egeria-connector (or egeria replacement)

Replacing their current `bash` based deployment with extending the helm chart, and exposing configuration of: the service endpoints (and credentials) needed for the connectors to connect.

This work _may_ also include merging the code into the manager as library code (currently the manager communicates to each connector which exposes the service at a designated port number)

# Stage 3: Helm chart: manager + connectors + 3rd-party

In small (to evaluation/demo) deployments you may want to deploy _some_ of the 3rd party tools, thus the proposal is to add these as configurable options such that one can either deploy m4d with a local/private instance of a required 3rd party of rather point to an existing shared 3rd party

specifically we should consider:
- vault server
- cert-manager server
- TBD

# Stage 4: Helm chart: manager + connectors + 3rd-party + modules

The current system of module management was intended to enable external parties to extend m4d w/o having to change the core of m4d, this approach makes sense for either super-large-and-complex OSS (which m4d is not, as the lines-of-code its 1/100 or 1/1000 or projects such as envoy, k8s, istio and others), or it was suppose to serve for extending a closed source project (which again is not the case of m4d).

So the proposal is to eliminate all together the module notion, merge it into the core of m4d and have it all managed via a single helm chart.

# Stage 5: make architecture multi-cluster compatible/agnostic

[...]

PR #208, opened 2020-12-13
charts: added m4d-crd helm

hunchback:
this PR is part of addressing #184 

added chart which contains m4d-crd, the decision to separate CRDs is:
- we may have several helm charts which depend on the CRDs requiring it be deployed independently
- we want to be able to upgrade the CRDs independently from core
- that this is an acceptable practice see rancher/fleet for example

the is not yet combined into CI, once the entire helm based deployment is ready the integration test will be changed accordingly, but for now I tested manually:

```
charts/m4d-crd$ helm install . --generate-name
NAME: chart-1607887038
LAST DEPLOYED: Sun Dec 13 21:17:19 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None

charts/m4d-crd$ helm uninstall chart-1607887038
release "chart-1607887038" uninstalled
```

### DIFF
--- a/charts/m4d-crd/Chart.yaml
+++ b/charts/m4d-crd/Chart.yaml
@@ -1,0 +1,4 @@
+apiVersion: v2
+name: m4d-crd
+description: M4D CustomResourceDefinitions
+version: 0.0.0
--- a/charts/m4d-crd/templates/crds.yaml
+++ b/charts/m4d-crd/templates/crds.yaml
@@ -1,0 +1,3346 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.3.0
+  creationTimestamp: null
+  name: batchtransfers.motion.m4d.ibm.com
+spec:
+  additionalPrinterColumns:
+  - JSONPath: .spec.source.description
+    name: Source
+    type: string
+  - JSONPath: .spec.destination.description
+    name: Destination
+    type: string
+  - JSONPath: .spec.schedule
+    name: Schedule
+    type: string
+  - JSONPath: .status.status
+    name: Status
+    type: string
+  group: motion.m4d.ibm.com
+  names:
+    kind: BatchTransfer
+    listKind: BatchTransferList
+    plural: batchtransfers
+    singular: batchtransfer
+  scope: Namespaced
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      description: BatchTransfer is the Schema for the batchtransfers API
+      properties:
+        apiVersion:
+          description: 'APIVersion defines the versioned schema of this representation
+            of an object. Servers should convert recognized schemas to the latest
+            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          description: BatchTransferSpec defines the state of a BatchTransfer. The
+            state includes source/destination specification, a schedule and the means
+            by which data movement is to be conducted. The means is given as a kubernetes
+            job description. In addition, the state also contains a sketch of a transformation
+            instruction. In future releases, the transformation description should
+            be specified in a separate CRD.
+          properties:
+            destination:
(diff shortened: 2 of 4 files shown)


roee88:
I looked at the code in this PR. It would be better to open a dedicated issue for that so we can discuss the requirements first.

shlomitk1 (commented):
I have several comments regarding crds.yaml
1) Can the CRDs be split into several files?
2) I thought all kustomization will be replaced by a chart. If so, some artifacts are missing, e.g. deployment, configmap, secrets, ...

hunchback:
> I looked at the code in this PR. It would be better to open a dedicated issue for that so we can discuss the requirements first.

@roee88 this should be discussed within #184 as the alternatives are either CRDs as a separate helm chart or integrated within the core helm chart, so if we decide on the later we will have to re-merge the issue back into #184 .... as to #184 I am going to add information supporting the usage of using a separate helm chart (and providing examples from other projects)

hunchback:
> I have several comments regarding crds.yaml
> 
> 1. Can the CRDs be split into several files?
> 2. I thought all kustomization will be replaced by a chart. If so, some artifacts are missing, e.g. deployment, configmap, secrets, ...

1. no, as generation of the final CRDs as done via a single kustomize command (config/crds directory)
2. the helm for CRDs is just a first step (see #184 for more details)

manager/Makefile
```diff
@@ -50,6 +50,11 @@ install: $(TOOLBIN)/kustomize $(TOOLBIN)/kubectl manifests
 uninstall: $(TOOLBIN)/kustomize $(TOOLBIN)/kubectl manifests
 	$(TOOLBIN)/kustomize build config/crd | $(TOOLBIN)/kubectl delete -f -
 
+# Install CRDs into a charts/m4d-crds
+.PHONY: charts-crd
+charts-crd: $(TOOLBIN)/kustomize $(TOOLBIN)/kubectl manifests
```
roee88, line 55: Do we want to just put it in `manifests` directly? Otherwise the chart might accidently be outdated if someone pushed a PR that changes the CRDs but forgot to run charts-crd

charts/m4d-crd/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: m4d-crd
+description: M4D CustomResourceDefinitions
+version: 0.0.0
+appVersion: 0.0.0
```
roee88, line 5: appVersion is optional and we should probably omit it

charts/m4d-crd/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: m4d-crd
+description: M4D CustomResourceDefinitions
+version: 0.0.0
```
roee88, line 4: Would you like to open an issue for a release process?  You might have an idea on how to set this version before a release without manual edits.

charts/m4d-crd/templates/crds.yaml
```diff
@@ -0,0 +1,3346 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
roee88, line 1: should the folder be named `m4d-crds` rather than `m4d-crd`?

charts/m4d-crd/Chart.yaml
hunchback: will omit

charts/m4d-crd/Chart.yaml
hunchback: no, automation, I want this to be a human decision as changing CRDs may not change each release (this is usually the case in other products)

charts/m4d-crd/templates/crds.yaml
hunchback: no, in other projects such as fleet its' named `fleet-crd`

charts/m4d-crd/Chart.yaml
roee88: You mean that the version here is not the same as the M4D version?

manager/Makefile
hunchback: for now I don't want to enforce any automation, will do once I have a fully helm based deployment and then will change the integration test to be helm based (and then will add automation to ensure consistency, plus ensure that `make verify` ensures crds are up-to-date)

charts/m4d-crd/Chart.yaml
roee88: I see that fleet just fixes it as "0.0.0" so I guess no need to change ever?

manager/Makefile
roee88: OK, I'll need to remember this one.

charts/m4d-crd/templates/crds.yaml
roee88: Yeah, I saw that in fleet. Still thought that `crds` is more fitting. Doesn't really matter though.

roee88 (approved):

PR #209, opened 2020-12-14
manager: improve plotter reconcile loop

froesef:
Continues the the processing of other blueprint specs in the plotter reconcile loop if blueprints are updated instead of returning for each update or not found blueprint.


roee88 (approved):
Generally what I do is keep `requeueNeeded, err` variables before loops and then only after the loop return a result based on it (rather than fail early). This PR addresses it but not for errors. Please consider this for the future.

froesef:
I will revisit the error handling. About the requeueing as there is no external mechanism right now that gives notifications the current strategy is to periodically requeue. Although I'd like to introduce some kind of backoff if the state does not seem to progress further. (e.g. if the plotter is in state ready only check status every minute or so)

PR #211, opened 2020-12-14
travis: release using github token

hunchback:
make use of the existing github token for adding more files to release (currently only README.md)


roee88 (approved):
LGTM.
BTW, why is the release task not a stage like the others?

Issue #105, opened 2020-11-01
BacthTransfer deletion fails


shlomitk1:
When an m4dapplication is deleted, two things happen: the relevant blueprint is deleted, and the allocated namespace is deleted.
Deletion of a blueprint causes uninstall of a Helm release which deletes a batchtransfer resource if exists. However, batchtransfer is not deleted, and the namespace is stuck in Terminating status.

froesef:
This is due to the fact that the BatchTransfer is starting a finaliser pod when it gets deleted. But when the namespace is terminating it cannot create any new objects in there. So we need to wait for deleting the namespace until all the finalisers and objects have been deleted.

roee88:
Sounds like an issue in the finalizer design, because normally deleting a namespace shouldn't block (think of human users of the motion API as well). Was this design pattern taken from another project? If so, can you share which?

froesef:
No I came up with the design pattern myself. The thought behind was that the mover knows best how to talk to the stores. 
Removing a file from COS would probably not be a problem in go. But when accessing other stores like Kafka would bring in all Kafka dependencies into the manager then. (e.g. when the finalizer there should remove a topic) Or for a jdbc connection a finalizer can remove a created table. And as all those data sources are available in the mover it's easier to start a pod that should quickly execute a delete action on a data store.

roee88:
So short-term we need to do cleanup in our own code correctly and add a warning in the movement API comments (to appear in the generated docs). How about deleting the blueprint and only upon its actual deletion delete the namespace? Or do we have to delete the individual BatchTransfers?

Long term I think that we need to find an alternative design pattern, though it might be a difficult task.

shlomitk1:
M4DApplication owns two resources: Blueprint and Namespace. It can delete Blueprint first, verify that Blueprint is deleted, and only then delete the namespace. Blueprint uninstalls helm charts and dies. This is not enough. Prior to uninstalling a release, it should "remember" the list of associated resources (it already manages them for status updates), and after uninstall - wait till all are deleted, and only then remove the finalizer and die. By waiting I refer to requeue every X seconds.

roee88:
> Prior to uninstalling a release, it should "remember" the list of associated resources (it already manages them for status updates), and after uninstall - wait till all are deleted

No option to uninstall with a wait? Not saying that it's better, just checking the options.

shlomitk1:
I did not think about that... It's definitely worth checking.

PR #214, opened 2020-12-15
fix blueprint deletion

shlomitk1:

- Fixes https://github.com/IBM/the-mesh-for-data/issues/105
This PR removes blueprint finalizers only after ensuring that there are no helm releases attached to this blueprint. Prior to deleting the blueprint namespace, a check is done to ensure that the blueprint no longer exists.
- Also, tests have been modified to avoid conflicts between m4dapplications having the same name.
- Fake Helm implementation has been changed to return a different status after uninstall (release is nil after uninstall).


manager/controllers/app/blueprint_controller.go
```diff
@@ -99,7 +99,9 @@ func (r *BlueprintReconciler) reconcileFinalizers(blueprint *app.Blueprint) erro
 				r.Log.V(0).Info("Error while deleting owned resources: " + err.Error())
 				return err
 			}
-
+			if r.hasExternalResources(blueprint) {
+				return errors.NewPlain("helm release uninstall is still in progress")
```
roee88, line 103: Are you relying on exponential backoff here? I think treating this case differently in `Reconcile` and using RequeueAfter would be a better choice.

shlomitk1: Added a different Result for this case.

roee88 (approved):

Issue #216, opened 2020-12-20
test: vault connector errors on Mac


simanadler:
When running "make test" on my Mac I got the following:

go: finding github.com/lib/pq v1.8.0
go: finding github.com/patrickmn/go-cache v2.1.0+incompatible
?   	github.com/ibm/the-mesh-for-data/connectors/vault	[no test files]
?   	github.com/ibm/the-mesh-for-data/connectors/vault/testutil	[no test files]
?   	github.com/ibm/the-mesh-for-data/connectors/vault/vault_utils	[no test files]
# github.com/shirou/gopsutil/disk
../../../../pkg/mod/github.com/shirou/gopsutil@v2.20.9+incompatible/disk/disk_darwin.go:67:51: cannot use stat.Mntfromname[:] (type []byte) as type []int8 in argument to common.IntToString
../../../../pkg/mod/github.com/shirou/gopsutil@v2.20.9+incompatible/disk/disk_darwin.go:68:49: cannot use stat.Mntonname[:] (type []byte) as type []int8 in argument to common.IntToString
../../../../pkg/mod/github.com/shirou/gopsutil@v2.20.9+incompatible/disk/disk_darwin.go:69:50: cannot use stat.Fstypename[:] (type []byte) as type []int8 in argument to common.IntToString
../../../../pkg/mod/github.com/shirou/gopsutil@v2.20.9+incompatible/disk/disk_darwin.go:85:43: cannot use stat.Fstypename[:] (type []byte) as type []int8 in argument to common.IntToString
FAIL	github.com/ibm/the-mesh-for-data/connectors/vault/vaultconn-bl [build failed]
FAIL

simanadler:
I also tried: NO_SIMULATED_PROGRESS=true USE_EXISTING_CLUSTER=true make -C manager test
which failed as follows:

Makefile:74: warning: overriding commands for target `docker-build'
../hack/make-rules/docker.mk:18: warning: ignoring old commands for target `docker-build'
../hack/tools/bin/controller-gen object:headerFile=../hack/boilerplate.go.txt,year=2020 paths="./..."
../hack/tools/bin/controller-gen "crd:trivialVersions=true" rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
go test ./... -coverprofile cover.out
?   	github.com/ibm/the-mesh-for-data/manager	[no test files]
2020/12/20 13:08:18 Unauthorized
[...]

PR #217, opened 2020-12-20
vault: fix darwin with go.mod gopsutil pkg upgrade

hunchback:
based on https://github.com/shirou/gopsutil/issues/853 I attempt to fix darwin build as reported in issue #216 

@simanadler - please check on your Mac if this fixes the issue


simanadler (approved):
Yes, this solves the problem on the Mac

PR #228, opened 2020-12-22
vault: refactor

hunchback:
cleanup vault deployment:
- leave only stuff related to deployment of the service
- move the secret-provider what is related to its' setup

### DIFF
--- a/third_party/vault/deploy.sh
+++ b/third_party/vault/deploy.sh
@@ -37,11 +37,8 @@ case "$1" in
     undeploy)
         undeploy
         ;;
-    wait_for_vault)
-      wait_for_vault
-      ;;
     *)
-        echo "usage: $0 [deploy|undeploy|wait_for_vault]"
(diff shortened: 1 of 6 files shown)


third_party/vault/deploy.sh
```diff
@@ -37,9 +37,6 @@ case "$1" in
     undeploy)
         undeploy
         ;;
-    wait_for_vault)
-      wait_for_vault
-      ;;
     *)
         echo "usage: $0 [deploy|undeploy|wait_for_vault]"
```
shlomitk1, line 41: wait_for_vault should be removed from usage instructions as well

shlomitk1 (approved):
Please see my comment.

third_party/vault/deploy.sh
hunchback: fixed

shlomitk1 (approved):

Issue #220, opened 2020-12-21
Replace the use of flight in `testdata/e2e/module-read.yaml` with a dummy module


roee88:
In https://github.com/IBM/the-mesh-for-data/blob/master/manager/testdata/e2e/module-read.yaml the real flight module is being used, creating a dependency on https://github.com/IBM/the-mesh-for-data-flight-module for testing. The suggestion is to replace it with a dummy instead. Would there be any issues with it?

PR #229, opened 2020-12-22
replacing read module with dummy for testing

shlomitk1:
Fixes https://github.com/IBM/the-mesh-for-data/issues/220

This PR replaces a real arrow-flight module chart by a dummy m4d-template module.


roee88 (approved):

hunchback:
this is an end-to-end test, do we not want to use the real flight module, such as is the case with the other modules ?

roee88:
If you think that there's value in testing with flight or any other real module then open an issue about it. The current tests doesn't do anything with it, it tests the manager and its usage in a read flow.

PR #231, opened 2020-12-23
Charts vault

hunchback:
added vault setup (using standard helm) plus vault setup (as part of m4d chart)

### DIFF
--- a/charts/Makefile
+++ b/charts/Makefile
@@ -12,14 +12,19 @@ cert-manager:
 	kubectl create namespace cert-manager
 	helm repo add jetstack https://charts.jetstack.io
 	helm repo update
-	helm install cert-manager jetstack/cert-manager \
-	  --namespace cert-manager --version v1.1.0 --set installCRDs=true \
-	  --wait --timeout 120s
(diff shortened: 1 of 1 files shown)


charts/Makefile
```diff
@@ -12,14 +12,19 @@ cert-manager:
...
+		--namespace cert-manager \
+		--wait --timeout 120s
 
 .PHONY: vault
 vault:
 	@echo "Installing vault ..."
-	cd ../third_party/vault && $(MAKE) deploy deploy-wait
+	helm repo add hashicorp https://helm.releases.hashicorp.com
+	helm install vault hashicorp/vault --version 0.8.0 \
+                --set "global.openshift=false" \
```
roee88, line 25: Why is it always false?

hunchback: this setup is on a local kind cluster, so its' false, in our instructions that we prepare later on we will explain setup for Openshift as well

roee88: Eventually one should be able to install all charts on OpenShift as well (for evaluation) just like any other Kubernetes cluster.

roee88 (approved):

PR #233, opened 2020-12-23
website: update dev running integration tests

hunchback:
notice updated documentation for developers

### DIFF
--- a/website/content/contribute/code/flow/index.md
+++ b/website/content/contribute/code/flow/index.md
@@ -83,41 +83,77 @@ USE_EXISTING_CONTROLLER=true NO_SIMULATED_PROGRESS=true USE_EXISTING_CLUSTER=tru
   should be used. E.g. in integration tests running against an existing setup a controller is already existing
   in the Kubernetes cluster and should not be started by the test as two controllers competing may influence the test.
 
-Setup default local kind cluster with a local image registry:
-```bash
-make kind
+# Running integration tests
+
+## Pre requisite
+
+The integration tests make use of the local docker registry such enables using
+current docker images (rather than official images) during tests. The docker
+registry runs on localhost:5000 but in order that it be accessible from within
+the k8s cluster it has to be designated a FQDN named `kind-registry`.
+
+To support local image registry host resolution append the following to /etc/hosts:
+
+```
+127.0.0.1       kind-registry
 ```
 
-By default the docker setup points to the official docker.io registry. When developing
-locally and building/pushing docker containers the docker host and namespace should be changed. When 
-using a local kind setup with a local registry the environment can be pointed to it by setting the following 
-environment variables. (Please note that you also have to add an entry to your /etc/hosts file):
+## Running in one step
+
+With the following you will then setup a kind cluster with the local registry,
+build and push current docker images and finally run the integration
+tests on it:
 
 ```bash
+make run-integration-tests
+```
+
+## Running step by step
+
+It is also possible to call the commands step by step, which sometimes is
+useful if you want to only repeat a specific step which failed without having
+to rerun  the entire sequence
+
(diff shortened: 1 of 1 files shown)


shlomitk1 (approved):

website/content/contribute/code/flow/index.md
```diff
@@ -83,41 +83,17 @@ USE_EXISTING_CONTROLLER=true NO_SIMULATED_PROGRESS=true USE_EXISTING_CLUSTER=tru
...
-Setup default local kind cluster with a local image registry:
-```bash
-make kind
-```
-
-By default the docker setup points to the official docker.io registry. When developing
-locally and building/pushing docker containers the docker host and namespace should be changed. When 
-using a local kind setup with a local registry the environment can be pointed to it by setting the following 
-environment variables. (Please note that you also have to add an entry to your /etc/hosts file):
+To support local image registry host resolution append the following to /etc/hosts:
```
simanadler, line 86: Not clear what the issue is that you're telling the developer to address, nor is it clear to me what exactly you want him to do.

website/content/contribute/code/flow/index.md
```diff
@@ -83,41 +83,17 @@ USE_EXISTING_CONTROLLER=true NO_SIMULATED_PROGRESS=true USE_EXISTING_CLUSTER=tru
...
 ```
 
-Deploy on the cluster. This will install the CRDs, dependencies such as the certificate manager and the controller
-itself. The default will pull the images from docker.io. If a local development setup is used please make sure
-that $DOCKER_HOSTNAME is set to the registry that should be used. 
-
-```bash
-make deploy
-```
+With the following you will then setup a kind cluster with the local registry
```
simanadler, line 92: Grammar is a bit funky here.  I would suggest something like ... The following command (1) sets up a kind cluster that uses a local registry containing the required images, and then runs a suite of integration tests.  These test check .... _provide information about what is tested_

simanadler (changes requested):

website/content/contribute/code/flow/index.md
hunchback: Let me explain here and will be happy if you help phrase it.

So that we can use the local registry (which is a local docker image exposing the docker registry v2 protocol on localhost:5000), so that we can access it from within our k8s cluster we need to assign a FQDN to it, in our case we have assigned the FQDN of `kind-registry`, and so that its' associated with the localhost we need to add an entry into `/etc/hosts`.

Happy if you can take a go at phrasing it in an easy to understand way.

Thanks!

website/content/contribute/code/flow/index.md
hunchback: ok based on your suggestion I plan:

The following command sets up a kind cluster that uses a local registry (exposed at kind-registry:5000) containing the required images, and then runs a suite of integration tests. These test check that the m4d core works against  a set of mocked services.

website/content/contribute/code/flow/index.md
roee88: Are you sure that we have to? There is no such requirement in https://kind.sigs.k8s.io/docs/user/local-registry/ (though it doesn't use FQDN) and we do add kind-registry to the same docker network. 

@froesef please advise.

froesef: `localhost:5000` would work fine for docker images. But helm cannot work with a local registry that is using https. (At least the last time I checked some months ago). So the only way that worked to push both docker images and helm charts to the registry to `kind-registry:5000` (using https) As this is not a proper address that is available everywhere we have to register it manually in `/etc/hosts`.

roee88: Got it, thanks.

Issue #226, opened 2020-12-22
Remove processingGeography from m4dapplication spec


shlomitk1:
M4DApplication spec includes a selector that connects m4dapplication to the user workload. This selector should include the cluster name in which the workload is running. If no cluster name exists, a local cluster is assumed. 
This cluster name should be used by the manager to determine the processing geography instead of specifying the geography in the spec.

shlomitk1:
PR that addresses this issue has been merged.

PR #234, opened 2020-12-23
obtain processing geography from cluster selector

shlomitk1:
Signed-off-by: SHLOMITK@il.ibm.com 

This PR is related to https://github.com/IBM/the-mesh-for-data/issues/226

1. Extended the existing LabelSelector of m4dapplication by ClusterSelector which contains both the label selector and the cluster name of the running workload. 
2. Removed processingGeography from the spec. It is received from ClusterSelector, using cluster metadata. If no cluster is specified in the selector, a local cluster is assumed and is read from the cluster-metadata config.
3. Removed a cloned selector from the blueprint. Not sure it is required at the blueprint level running in a remote cluster.
4. Minor changes to silence a lint checker.


manager/apis/app/v1alpha1/cluster_selector.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ClusterSelector is a label query over a set of resources in the specified cluster.
+type ClusterSelector struct {
+	// Cluster name
+	// +required
+	ClusterName string `json:"clusterName"`
```
simanadler, line 14: I see that clustername is required, but in your general comments didn't you say that if it's not provided you would take the information from the local cluster?

manager/apis/app/v1alpha1/cluster_selector.go
```diff
@@ -0,0 +1,20 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ClusterSelector is a label query over a set of resources in the specified cluster.
+type ClusterSelector struct {
```
simanadler, line 11: I thought @roee88 wanted two selectors in the spec rather than combining them into a single structure, but it could very well be that I didn't understand him correctly.  @roee88 is this what you wanted?

manager/apis/app/v1alpha1/m4dapplication_types.go
```diff
@@ -49,7 +44,9 @@ type M4DApplicationSpec struct {
 
 	// Selector enables to connect the resource to the application
 	// Application labels should match the labels in the selector.
-	Selector metav1.LabelSelector `json:"selector"`
+	// For some flows the selector may not be used.
+	// +optional
+	Selector ClusterSelector `json:"selector"`
```
simanadler, line 49: See previous comment

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -381,3 +390,30 @@ func (r *M4DApplicationReconciler) GetAllModules() (map[string]*app.M4DModule, e
 	}
 	return moduleMap, nil
 }
+
+// GetProcessingGeography determines the geography of the workload cluster.
+// If no workload has been specified, a local cluster is assumed.
+func (r *M4DApplicationReconciler) GetProcessingGeography(applicationContext *app.M4DApplication) (string, error) {
+	clusterName := applicationContext.Spec.Selector.ClusterName
+	if clusterName == "" {
```
simanadler, line 398: Related to previous comment ... Why check for an empty string instead of nil?

manager/apis/app/v1alpha1/cluster_selector.go
shlomitk1: Cluster selector itself may not be provided in the spec. If it is provided, it will have both fields defined.

manager/controllers/app/m4dapplication_controller.go
shlomitk1: It is not nil. It will be an empty structure with all strings in it equal to an empty string.

manager/apis/app/v1alpha1/cluster_selector.go
roee88: No it's fine as is. I would just name `ClusterSelector` differently because it's not just a cluster selector. Maybe even just `Selector`.

manager/apis/app/v1alpha1/cluster_selector.go
roee88: @shlomitk1  why not make this optional as well?

shlomitk1: OK.

manager/apis/app/v1alpha1/cluster_selector.go
shlomitk1: Renamed.

ronenkat:
@shlomitk1 you are making changes to the API.
Do we have a policy about backward compatibility? if not, can we define one? i.e., from which version whould we start being backward compatible? how do we manager versions?
@roee88 do we have these guidelines?

roee88:
> @shlomitk1 you are making changes to the API.
> Do we have a policy about backward compatibility? if not, can we define one? i.e., from which version whould we start being backward compatible? how do we manager versions?
> @roee88 do we have these guidelines?

We are not yet maintaining any backward compatibility, this is why the API is marked v1alpha1 (we're not special, that's what everyone else does).

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -381,3 +390,30 @@ func (r *M4DApplicationReconciler) GetAllModules() (map[string]*app.M4DModule, e
...
+		clusterMetadataConfigmap := corev1.ConfigMap{}
+		namespacedName := client.ObjectKey{
+			Name:      "cluster-metadata",
+			Namespace: utils.GetSystemNamespace(),
+		}
+		if err := r.Client.Get(context.Background(), namespacedName, &clusterMetadataConfigmap); err != nil {
+			return "", err
+		}
+		return clusterMetadataConfigmap.Data["Region"], nil
```
roee88, line 407: That's an implementation detail of `local.ClusterManager`. If we need an indication of the local cluster or an indication of the primary cluster then let's add it to the cluster manager interface instead.

roee88 (changes requested):

manager/controllers/app/m4dapplication_controller.go
shlomitk1: Added a method GetLocalCluster to the ClusterLister interface.

razee implementation of this function is not yet done - I need help with this. Local implementation uses GetClient() of manager which is not passed to razee manager. Either to pass this argument, or is there a "razee native" way to use the local config?

roee88: @froesef ^

shlomitk1: Using local implementation to get local cluster information

pkg/multicluster/local/local_cluster_manager.go
```diff
@@ -19,31 +19,25 @@ type ClusterManager struct {
...
-
-func (cm *ClusterManager) GetLocalCluster() (multicluster.Cluster, error) {
 	clusterMetadataConfigmap := corev1.ConfigMap{}
 	namespacedName := client.ObjectKey{
 		Name:      clusterMetadataConfigmapName,
 		Namespace: cm.Namespace,
 	}
 	if err := cm.Client.Get(context.Background(), namespacedName, &clusterMetadataConfigmap); err != nil {
-		return multicluster.Cluster{}, err
+		wrappedError := fmt.Errorf("Error in GetClusters: %w", err)
```
roee88: We sometimes use this method and sometimes errors.Wrap. If you have a preference let's document it in the contribution guidelines and stick to one option.

roee88 (approved):

pkg/multicluster/local/local_cluster_manager.go
shlomitk1: I've reverted this file to the master branch, it is not a part of this PR :-)

PR #239, opened 2020-12-30
charts: added opa deployment to m4d

hunchback:
added to the m4d chart an option for also installing the opa server (which for now is installed by default global.opa=true), this is currently part of the foundational services required for running a basic m4d. as opa does not have an official helm chart (as cert-manager and vault have) for now it is integrated (as a conditional option) within m4d, going forward when opa has an official helm chart then this will be removed and the sequence (of deploying m4d) will be changed to include running `helm install opa opa`

### DIFF
--- a/charts/m4d/templates/opa.yaml
+++ b/charts/m4d/templates/opa.yaml
@@ -1,0 +1,238 @@
+{{- if .Values.global.opa }}
+apiVersion: v1
+data:
+  action_struct.rego: "package data_policies\n\n#general enforcment action structure\nenforcement_action_struct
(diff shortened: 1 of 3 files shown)


charts/m4d/templates/opa.yaml
```diff
@@ -0,0 +1,238 @@
+{{- if .Values.global.opa }}
+apiVersion: v1
+data:
+  action_struct.rego: "package data_policies\n\n#general enforcment action structure\nenforcement_action_struct
```
roee88, line 4: OMG

roee88 (approved):

PR #241, opened 2020-12-31
make: add run-deploy-tests target

hunchback:
the deploy phase in travis was silently failing (as described in the associated issue), as a result there is no guarantee that the images placed in the public docker repo are actually ok and can execute.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,17 @@ run-integration-tests:
 	$(MAKE) helm
 	$(MAKE) -C manager run-integration-tests
 
+.PHONY: run-deploy-tests
+run-deploy-tests:
+	$(MAKE) kind
+	$(MAKE) cluster-prepare
+	$(MAKE) -C third_party/opa deploy
+	kubectl apply -f ./manager/config/prod/deployment_configmap.yaml
+	kubectl create secret generic user-vault-unseal-keys --from-literal=user-vault-root=''
+	$(MAKE) -C connectors deploy
+	kubectl get pod --all-namespaces
+	kubectl wait --for=condition=ready pod --all-namespaces --all --timeout=120s
+
 .PHONY: cluster-prepare
 cluster-prepare:
 	$(MAKE) -C third_party/cert-manager deploy
--- a/.travis.yml
+++ b/.travis.yml
@@ -101,17 +101,10 @@ jobs:
 
   - stage: Deploy
     name: "Kind deploy"
-    env:
-    - COMPONENT=connectors
-    - CLUSTER=kind
     install:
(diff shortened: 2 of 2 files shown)


manager/config/prod/deployment_configmap.yaml
```diff
@@ -42,3 +42,4 @@ data:
   #external vault  variables
   USER_VAULT_ADDRESS: "http://vault:8200/"
   USER_VAULT_PATH: "external"
+  USER_VAULT_TOKEN: ""
```
shlomitk1, line 45: Any vault token should be in a secret and not in a configmap. 
I remember that there was a script that was reading a vault token from vault secret, creating a new secret to be deployed prior to the connector deployment. @rohithdv can provide more details.

shlomitk1 (changes requested):
Please remove USER_VAULT_TOKEN from m4d-config.

manager/config/prod/deployment_configmap.yaml
hunchback: its to enable the test to pass using a dummy value

shlomitk1: USER_VAULT_TOKEN is relevant only for vault connector and is defined in the following way:
```
       env:
        - name: USER_VAULT_TOKEN
          valueFrom:
            secretKeyRef:
              name: user-vault-unseal-keys
              key: user-vault-root
```
What test was not passing? Perhaps the relevant secret (user-vault-unseal-keys) has not been configured.

hunchback: no. I am intentionally doing a simple dummy setup so not referencing a k8s Secret object but rather defining inline a dummy vault "" to enable the deployment test to run

hunchback: ok. I removed the field from the configmap

hunchback:
> Please remove USER_VAULT_TOKEN from m4d-config.

done

shlomitk1 (approved):

Makefile
```diff
@@ -33,6 +33,19 @@ run-integration-tests:
 	$(MAKE) helm
 	$(MAKE) -C manager run-integration-tests
 
+.PHONY: run-deploy-tests
+run-deploy-tests: export DOCKER_HOSTNAME?=ghcr.io
+run-deploy-tests: export DOCKER_NAMESPACE?=the-mesh-for-data
```
roee88, line 38: Why do we need these?

hunchback: removed

roee88 (approved):

PR #242, opened 2020-12-31
Connectors refactor

hunchback:
make each connector deployment self-contained, remove dead code, ensure deployment tests still passes


connectors/helpers/deploy.sh
```diff
@@ -1,95 +0,0 @@
...
-}
-
-undeploy() {
-        $WITHOUT_VAULT || vault_delete
-        connectors_delete
-        kube_cluster_info
-}
-
-deploy() {
-        $WITHOUT_VAULT || vault_delete
```
revit13, line 81: Regarding $WITHOUT_VAULT  - it is used in several places for example in hack/install.sh so it seems you need to address it in the refactor code.

hunchback: @revit13 I checked with its' used and I was not able to see how my change effects any of below, kindly provide details, for example how is hack/install.sh effected

```
aidan@AIDANSHRIBMAN-SDFUE6N:~/projects/the-mesh-for-data$ grep -R WITHOUT_VAULT *
hack/install.sh:: ${WITHOUT_VAULT:=false}
hack/install.sh:$WITHOUT_VAULT || make -C third_party/vault deploy
hack/install.sh:WITHOUT_PORT_FORWARD=true WITHOUT_VAULT=false make deploy
manager/controllers/utils/vault.go:     return os.Getenv("RUN_WITHOUT_VAULT") == "1"
manager/controllers/utils/configuration.go:     SetIfNotSet("RUN_WITHOUT_VAULT", "1", t)
Binary file manager/bin/manager matches
secret-provider/deploy/deploy.sh:: ${WITHOUT_VAULT=true}
aidan@AIDANSHRIBMAN-SDFUE6N:~/projects/the-mesh-for-data$
```

rohithdv (approved):
I have tested connectors deployment on openshift and they seem to be working fine. 

However I had one query since I am trying to understand the latest changes to make targets. When I tried to run the "make run-deploy-tests" target in my local Mac machine I was observing that vault pod was in Init:0/1 state. Is there something I am missing in the configuration ? This is just a query. I am attaching the screenshot below. Thanks. 

image

PR #243, opened 2020-12-31
Charts connectors

hunchback:
add connectors deployment to the m4d helm chart (can be disabled using connectors.egeria=false etc).


roee88:
There is no concensus on putting these connectors in the same m4d chart. We will soon add a built in catalog and it might fit better as an optional install. I'm keeping this PR as draft and we will open a discussion.

hunchback:
@roee88 - this is one of the major elements needed to support a pure end-user deployment - adding this as part of m4d does not prohibit installing other connectors separatly - further more it will enable alignment of parameters such as tcp port.

roee88:
I don't argue that having an easy default install that works out of the box is needed. I just don't think that we get it with what we will have after this PR since the current connectors are more complex than something that just works out of the box. I encourage you to first finish the m4d chart in terms of eliminating the use of kustomize altogether.

Issue #221, opened 2020-12-21
Ingest Scenario Support


simanadler:
Add support for copying data from an external system into the managed environment and ensuring it is cataloged in the data catalog.  There are several scenarios in which new data is created:
1) Data in external systems are ingested for the first time into the environment, and thus the new data set needs to be cataloged.
2) Data is being appended or is updating an existing data set, and thus the data set already exists in the catalog.  _Does classification need to be re-run after new data is added?  Does new metadata need to be added? _

For a discussion about the ingest flow implementation options and what was chosen: https://docs.google.com/document/d/1j5NBs-xN8y8hxM01g2ZwgGvROQGIV-yqk-LwbJsfUoo/edit?usp=sharing

For information about policies relating to these scenarios see: https://docs.google.com/document/d/1rQ4ykFfOZ4mfD6rrd1XgInKSiJ0DPBoHu9xwawJecFI/edit?usp=sharing

roee88:
Why are these part of the same issue?

roee88:
Focusing on ingest, my understanding is that you wanted a module to do the ingest copy. In that case, what I imagined is `M4DApplication` with the regular read parameters but with these changes/extensions: 
1. Support read not from a catalog (cf. creating a private catalog for the ingestion)
2. Adding usage hints/parameters like we discussed in https://github.com/IBM/the-mesh-for-data/issues/93 (useful for implicit copy as well)
3. Make the selector optional (once we add cluster selector it could also be used as an optional parameter)
4. Add the cataloging functionality

simanadler:
@roee88 I put ingest and write in the same issue because I wanted to make sure we aligned the approach for both in order to not "break the architecture".  If you prefer to separate them that's fine.

As for the additions to M4DApplication, my thoughts were the same.  I am working on updating the M4DApplication yaml.  Would you like me to do a draft PR so it can be reviewed before additional work is done?

roee88:
> @roee88 I put ingest and write in the same issue because I wanted to make sure we aligned the approach for both in order to not "break the architecture". If you prefer to separate them that's fine.

I think it would be more meaningful to separate these two. They require two very different changes.

> As for the additions to M4DApplication, my thoughts were the same. I am working on updating the M4DApplication yaml. Would you like me to do a draft PR so it can be reviewed before additional work is done?

Sure

roee88:
[...]

Issue #104, opened 2020-11-01
Feature: provision a bucket dynamically


shlomitk1:
https://docs.google.com/document/d/1PzaQtED9NUq10REfxl0wMFOl247SKc0XvAUHpxjSYjY/edit?usp=sharing

When implicit copy is applied, the mesh for data needs to provide a bucket for copy destination. Currently, a single bucket is created manually, and the bucket credentials are stored in vault manually as well. 
Requirement: to provision a bucket dynamically when copy is required, to manage the bucket life cycle together with its credentials.

simanadler:
@shlomitk1 I think this issue needs to be a bit broader.  For the ingest flow (a future feature) we will also need to dynamically allocate storage.  In general I think the full scope of this task is as follows, although we will of course need to implement it gradually.

* IT person needs to somehow inform the M4D of the available storage accounts, their types, and the credentials for accessing them.  Cloud object store (COS) are one example, but there will be other types such as databases, etc.

When storage is required (temporary for implicit copy, or more permanent for the ingest scenario) M4D manager needs to allocate the appropriate storage in the relevant type of data store, ex: buckets in COS, in an account pre-allocated by the enterprise IT team.

As for the life cycle of the allocated bucket (or other type of storage) ...
* For implicit copy, M4D manager should release both the data and the bucket when the processing is complete - i.e. the M4DApplication instance is deleted
* For data that was ingested into the data lake (and cataloged in the Data Catalog during the process), there will have to be a separate flow for removing data from the data lake.  This is a future feature.  

@roee88 @ronenkat This is related to the DLF discussions

shlomitk1:
Things left to do:
- fix marking buckets as persistent (dependency on DLF)
- support multiple endpoints/regions for a single account
- check errors of deleting a dataset, ensure that m4dapplication deletion is not blocked
- analyze Dataset errors
- Register secret in vault - move to pkg/vault

shlomitk1:
Implemented.

Issue #93, opened 2020-10-28
Manager: Logic for choosing modules


simanadler:
The pilot component (M4DApplication controller) contains the logic for choosing the modules to be included in the blueprint.  Currently logic implemented (hard coded) is for example .... (@shlomitk1 keep me honest here)

(1) If the data set protocol/format and the protocol/format requested by the user do not match, then make an implicit copy of the data ,storing it such that it is readable via the protocol/format requested by the user.  
(2) If the governance action required on the data set are not supported by the read module, and it is supported by the implicit copy module ... then make an implicit copy.  Otherwise no need for implicit copy, and read will be done from the source directly.

There could be other rules that an organization might want to implement that would influence how to build the blueprint which we do not currently take into account.  For example ...
(3) If the data set is very large, aim to not use implicit copy.  Prefer direct use of the source.
(4) If performance is important, keep the data as close to the compute as possible.

We need a mechanism (1) for an enterprise to define these rules, (2) M4D should leverage these rules rather than hard coding it in the manager logic

shlomitk1:
@simanadler Current logic supports rule (1) only. There is a strict separation of READ and COPY actions that are not mixed. If no read module supports the required actions, an error will be returned.

roee88:
> There is a strict separation of READ and COPY actions that are not mixed

I thought that COPY policies is applied on top of READ policies in case of a copy. I'm not sure what the right approach is. @simanadler do you have any input here?

> The pilot component (M4DApplication controller) contains the logic for choosing the modules to be included in the blueprint. Currently logic implemented (hard coded) is for example .... (@shlomitk1 keep me honest here)
> 
> (1) If the data set protocol/format and the protocol/format requested by the user do not match, then make an implicit copy of the data ,storing it such that it is readable via the protocol/format requested by the user.
> (2) If the governance action required on the data set are not supported by the read module, and it is supported by the implicit copy module ... then make an implicit copy. Otherwise no need for implicit copy, and read will be done from the source directly.
> 
> There could be other rules that an organization might want to implement that would influence how to build the blueprint which we do not currently take into account. For example ...
> (3) If the data set is very large, aim to not use implicit copy. Prefer direct use of the source.
> (4) If performance is important, keep the data as close to the compute as possible.
> 
> We need a mechanism (1) for an enterprise to define these rules, (2) M4D should leverage these rules rather than hard coding it in the manager logic.
> 
> _Should these rules be handled by the policy manager??_

[...]

PR #248, opened 2021-01-03
Ingest flow: blueprint creation logic

simanadler:
Partially addresses issue #221, the ingest scenario.  
Enables the user to initiate an ingest flow and builds the relevant blueprint.  It does NOT yet contain:
* Register of the new data asset in the data catalog
* Allocation of permanent buckets.  Currently uses the mechanism for allocating implicit copy buckets, which are deleted when M4DApplication is deleted.  This will be addressed during integration with DLF.
* Write path flow, which @roee88 suggested be addressed separately.
* DataStage module.  Currently used existing copy module - S3 to S3.

Updated: Changed to take the ingested data set info from an "inline" catalog instead of declaring the connection parameters and context in M4DApplication.  Dependent on implementation of inline catalog and related credential issues. @roee88 

Updated 2: Now handles credentials and catalogs as described in updated Issue 221.

@ronenkat FYI


hack/tools/.srl
```diff
@@ -0,0 +1 @@
+E72C58454F0D02E0
```
shlomitk1, line 1: What is this?

manager/apis/app/v1alpha1/m4dapplication_types.go
```diff
@@ -8,7 +8,57 @@ import (
...
+	// No values means there is no sensitive data.
+	// +optional
+	Sensitivity []DataSensitivity `json:"sensitivity,omitempty"`
+
+	// IFdetails indicates the protocol and format expected by the data by the Data Scientist's application
+	// +required
+	IFdetails InterfaceDetails `json:"ifDetails"`
+
+	// Recurrence indicates how often the copy should take place or if streamed input is needed
+	// +optional
```
shlomitk1, line 57: You may want to add a comment about default being None.

manager/apis/app/v1alpha1/m4dapplication_types_write.go
```diff
@@ -0,0 +1,270 @@
+// Copyright 2020 IBM Corp.
```
shlomitk1, line 1: This file should not be committed

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -25,6 +25,7 @@ import (
 	"github.com/ibm/the-mesh-for-data/manager/controllers/utils"
 	pb "github.com/ibm/the-mesh-for-data/pkg/connectors/protobuf"
 	"github.com/ibm/the-mesh-for-data/pkg/multicluster"
+	m "github.com/ibm/the-mesh-for-data/pkg/multicluster"
```
shlomitk1, line 28: There is already an inclusion of multicluster (a line above), no need for duplication

manager/controllers/app/moduleinstance.go
```diff
@@ -75,13 +90,96 @@ func (m *ModuleManager) GetCopyDestination(item modules.DataInfo, destinationInt
...
+	if err != nil {
+		m.Log.Info("StructToInterfaceDetails failed for: ")
+		utils.PrintStructure(item, m.Log, "Ingest modules.DataInfo")
+
+		return instances, err
+	}
+
+	sourceDataStore := &app.DataStore{
+		Connection:         item.DataDetails.GetDataStore(),
+		CredentialLocation: utils.GetDatasetVaultPath(item.AssetID),
```
shlomitk1, line 126: This won't work for ingest. CredentialLocation is a part of the provided external data structure.

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -274,9 +287,51 @@ func (r *M4DApplicationReconciler) reconcile(applicationContext *app.M4DApplicat
 	return ctrl.Result{}, nil
 }
 
-// ConstructDataInfo gets a list of governance actions and data location details for the given dataset and fills the received DataInfo structure
+// ConstructIngestDataInfo already has the information about the external data set's endpoint,
+// but it adds the governance decisions. It also already has the link to the credentials, which was provided by the user.
+// NOTE: Assumes the credentials were put into the M4D credential manager by the user!!
```
shlomitk1, line 292: Assuming credentials are already in vault, the appropriate module still needs to access them, i.e. the vault path should be provided to the module. See my comment in module_instance.go - management of credentials location should be different.

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -274,9 +287,51 @@ func (r *M4DApplicationReconciler) reconcile(applicationContext *app.M4DApplicat
...
+	req.AppInterface = &externalDataset.IFdetails
+	req.Flow = app.Copy // This is how we know it's an ingest flow
+
+	// Check if we are allowed to write the data to any of the geographies available to us
+	// If yes, we will allocated the storage later
+	excludedGeos := ""
+	for _, cluster := range clusters {
+		req.Geo = cluster.Metadata.Region
+		if err := TempLookupPolicyDecision(externalDataset, r.PolicyCompiler, req, input, pb.AccessOperation_COPY); err != nil { // TODO: Change to regular policy compiler function
+			return AnalyzeError(input, r.Log, req.AssetID, err, "Policy Compiler")
```
shlomitk1, line 313: No asset id - pass ""

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -274,9 +287,51 @@ func (r *M4DApplicationReconciler) reconcile(applicationContext *app.M4DApplicat
...
+// but it adds the governance decisions. It also already has the link to the credentials, which was provided by the user.
+// NOTE: Assumes the credentials were put into the M4D credential manager by the user!!
+func (r *M4DApplicationReconciler) ConstructIngestDataInfo(externalDataset app.ExternalDataContext, req *modules.DataInfo, input *app.M4DApplication, clusters []m.Cluster) error {
+	dataDetails := pb.DatasetDetails{
+		DataOwner:  "",
+		DataStore:  externalDataset.ExternalStore.Connection,
+		DataFormat: externalDataset.ExternalStore.Format,
+		Geo:        string(externalDataset.Residency),
+	}
+	req.AssetID = externalDataset.ExternalStore.Connection.Name // Not cataloged so no id. Name isn't unique though.  OK?
```
shlomitk1, line 300: It may be left empty until a new id is created, in case we want to store this info here.

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -274,9 +287,51 @@ func (r *M4DApplicationReconciler) reconcile(applicationContext *app.M4DApplicat
...
+	req.Actions = make(map[app.ModuleFlow]modules.Transformations)
+	req.AppInterface = &externalDataset.IFdetails
+	req.Flow = app.Copy // This is how we know it's an ingest flow
+
+	// Check if we are allowed to write the data to any of the geographies available to us
+	// If yes, we will allocated the storage later
+	excludedGeos := ""
+	for _, cluster := range clusters {
+		req.Geo = cluster.Metadata.Region
+		if err := TempLookupPolicyDecision(externalDataset, r.PolicyCompiler, req, input, pb.AccessOperation_COPY); err != nil { // TODO: Change to regular policy compiler function
```
shlomitk1, line 312: We need to split the lookup into different actions - one for read/copy an existing asset, another - for writing a non-cataloged asset. The API is entirely different: dataset decisions vs. general decisions. 
Not sure whether the API for writing data holds enough inputs (data residency? sensitivity?) But this is another issue.

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -274,9 +287,51 @@ func (r *M4DApplicationReconciler) reconcile(applicationContext *app.M4DApplicat
...
+	req.DataDetails = &dataDetails
+	req.Credentials = nil
+	req.Actions = make(map[app.ModuleFlow]modules.Transformations)
+	req.AppInterface = &externalDataset.IFdetails
+	req.Flow = app.Copy // This is how we know it's an ingest flow
+
+	// Check if we are allowed to write the data to any of the geographies available to us
+	// If yes, we will allocated the storage later
+	excludedGeos := ""
+	for _, cluster := range clusters {
```
shlomitk1, line 310: In the future, some geographies must take precedence (for example, the same geography as specified).

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -274,9 +287,51 @@ func (r *M4DApplicationReconciler) reconcile(applicationContext *app.M4DApplicat
...
+		if excludedGeos != "" {
+			excludedGeos += ", "
+		}
+		excludedGeos += cluster.Metadata.Region
+	}
+
+	// We haven't found any geographies to which we are allowed to write
+	setCondition(input, req.AssetID, "Writing to all geographies denied: "+excludedGeos, "Policy Compiler", true)
+
+	return nil // Should this return an error?
```
shlomitk1, line 327: nil is fine - the user will receive an error message, and no further processing will be done. Error is returned in case of an unexpected event, such as grpc communication problem, problems communicating with vault, etc.

manager/controllers/app/moduleinstance.go
```diff
@@ -25,8 +25,23 @@ type ModuleManager struct {
...
+	- and register new data set in data catalog
+- If Data Context Flow=Write
+   - Write is always required, and always close to compute
+   - Implicit Copy is used on demand, e.g. if a write module does not support the existing source of data or governance actions
+   - Transformations are always done at workload location
+   - If not external data, then register in data catalog
+
+Updates to add ingest:
+- If no label selector assume ingest of external data
+	- run Copy module close to destination (determined based on governance decisions)
```
shlomitk1, line 42: I don't see how this is expressed in the code.

manager/controllers/app/storage.go
```diff
@@ -18,8 +18,9 @@ import (
 // +kubebuilder:rbac:groups=app.m4d.ibm.com,resources=m4dbuckets,verbs=get;list;watch;update;
 // +kubebuilder:rbac:groups=app.m4d.ibm.com,resources=m4dbuckets/status,verbs=get;update;patch
 
-// FindAvailableBucket finds an available storage asset
-func FindAvailableBucket(c client.Client, log logr.Logger, owner types.NamespacedName, id string, prefixBase string, canShare bool) *app.M4DBucket {
+// FindAvailableBucket finds an available storage asset - in the relevant geo
+// TODO: Implement check for geo.  Needed for ingest
+func FindAvailableBucket(c client.Client, log logr.Logger, owner types.NamespacedName, id string, prefixBase string, canShare bool, geo string) *app.M4DBucket {
```
shlomitk1, line 23: Are we assuming that a bucket can be allocated in any registered cluster?

hack/tools/.srl
simanadler: No idea. It got added during one of the merges

manager/apis/app/v1alpha1/m4dapplication_types.go
simanadler: done

manager/controllers/app/m4dapplication_controller.go
simanadler: fixed

hack/tools/.srl
shlomitk1: It is not in the master branch. I guess this was auto-generated at some stage, but there is no need to commit this file.

manager/controllers/app/moduleinstance.go
simanadler: You're right.  Fixed

manager/controllers/app/m4dapplication_controller.go
simanadler: Yes.  fixed

manager/controllers/app/m4dapplication_controller.go
simanadler: Agreed.  This is a temporary place holder until we do that.

manager/controllers/app/m4dapplication_controller.go
simanadler: Right now we use the first geography that is allowed.  In the future there may be additional logic, but I think that will be related to issue #93 and does not need to be addressed now

manager/controllers/app/m4dapplication_controller.go
simanadler: OK.  That's what I thought.  Removed comment in code

manager/controllers/app/moduleinstance.go
simanadler: The copy module SelectCluster determines where it will run.  No changes were required to it.  I sets the cluster to the destination if no transforms are required, which is fine.

manager/controllers/app/storage.go
simanadler: No, that's why I added the geo parameter.  The bucket should be allocated in a cluster in that geo.  Note though that although I added the parameter, the logic was not yet implemented.  I assume that will be done as part of the overall work on issue #104 (provision buckets dynamically)

manager/apis/app/v1alpha1/m4dapplication_types.go
```diff
@@ -72,6 +72,7 @@ type DataContext struct {
 	IFdetails InterfaceDetails `json:"ifDetails"`
 
 	// Recurrence indicates how often the copy should take place or if streamed input is needed
+	// Default is none
```
roee88, line 75: unless you add dots this looks bad in the generated docs:

> description: Recurrence indicates how often the copy should take place or if streamed input is needed Default is none

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -290,16 +289,29 @@ func (r *M4DApplicationReconciler) reconcile(applicationContext *app.M4DApplicat
...
+func (r *M4DApplicationReconciler) ConstructIngestDataInfo(externalDataset app.ExternalDataContext, req *modules.DataInfo, input *app.M4DApplication, clusters []multicluster.Cluster) error {
 	dataDetails := pb.DatasetDetails{
 		DataOwner:  "",
 		DataStore:  externalDataset.ExternalStore.Connection,
 		DataFormat: externalDataset.ExternalStore.Format,
 		Geo:        string(externalDataset.Residency),
 	}
+	// Store link to credentials rather than the credentials themselves
+	// This is a hack, but works for now.
+	creds := pb.Credentials{
```
shlomitk1, line 301: Please see https://github.com/IBM/the-mesh-for-data/pull/251. It keeps a path rather than the credentials themselves. Removes the need in this work-around.

manager/apis/app/v1alpha1/m4dapplication_types.go
```diff
@@ -57,11 +58,17 @@ type ExternalDataContext struct {
...
+*/
 
 // DataContext indicates data set chosen by the Data Scientist to be read from or written to by his application,
 // and includes information about the data format and technologies used by the application
 // to access the data.
 type DataContext struct {
+	// DataCatalogID indicates the data catalog that contains the information about the data set.  If no value is provided the assumption is there is only one data catalog.
+	// TODO: Temporary pending design for multi-catalog support.
+	// +optional
+	DataCatalogID string `json:"dataCatalogID"`
```
shlomitk1, line 70: Can this be a part of DataSetID? Today DataSetID is a json that includes catalog id and asset id. We could add another attribute to identify the catalog (egeria, inline, ...)

manager/apis/app/v1alpha1/m4dapplication_types.go
```diff
@@ -57,11 +58,17 @@ type ExternalDataContext struct {
 	// +optional
 	Recurrence RecurrenceType `json:"recurrence,omitempty"`
 }
+*/
```
shlomitk1, line 61: Is RecurrenceType commented out? Why?

manager/apis/app/v1alpha1/m4dapplication_types.go
```diff
@@ -75,6 +82,12 @@ type DataContext struct {
 	// Default is none
 	// +optional
 	Recurrence RecurrenceType `json:"recurrence,omitempty"`
+
+	// Cataloged indicates whether or not the data set has already been registered in the organization's data catalog.
+	// When a user chooses the data set from the catalog this should be true.  In flows where a new data set is
+	// being created in the M4D managed environment this should be false.
+	// +optional
+	Cataloged bool `json:"cataloged,omitempty"`
```
shlomitk1, line 90: Do we specify both an external catalog id and the state of the asset? If data comes from an external catalog, then it is not cataloged in the organization catalog.
Another question: supporting multiple catalogs - what does it mean? If we support multiple catalogs then we can also catalog the asset in more than one catalog.

manager/apis/app/v1alpha1/m4dapplication_types.go
```diff
@@ -57,11 +58,17 @@ type ExternalDataContext struct {
...
 
 // DataContext indicates data set chosen by the Data Scientist to be read from or written to by his application,
 // and includes information about the data format and technologies used by the application
 // to access the data.
 type DataContext struct {
+	// DataCatalogID indicates the data catalog that contains the information about the data set.  If no value is provided the assumption is there is only one data catalog.
+	// TODO: Temporary pending design for multi-catalog support.
+	// +optional
+	DataCatalogID string `json:"dataCatalogID"`
+
```
shlomitk1, line 71: Why optional? Manager needs to know what connector/catalog to access. Also, this should be an enum rather than a free string, e.g. Egeria, inline, ...

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -286,42 +282,45 @@ func (r *M4DApplicationReconciler) reconcile(applicationContext *app.M4DApplicat
...
+// TODO: Resolve credential issues
+func (r *M4DApplicationReconciler) ConstructNewDataInfo(dataset app.DataContext, req *modules.DataInfo, input *app.M4DApplication, clusters []multicluster.Cluster) error {
+	// Call the DataCatalog service to get info about the dataset
+	if err := GetConnectionDetails(dataset.DataSetID, dataset.DataCatalogID, req, input); err != nil {
+		return AnalyzeError(input, r.Log, dataset.DataSetID, err, "Catalog Connector")
+	}
+
+	// In cataloged data sets we are reading the actual credentials from the external credential manager.
+	// In this case we are assuming that the user put the credentials in our internal credential manager, and thus we only need to get the link.
+	// TODO: Need a way to get the link.  From where will we get it?  The internal data catalog?  Need a relevant API.  Temp code is a temporary place holder.
```
shlomitk1, line 299: I suggest the following scenario:
- Admin creates a secret (or several secrets) in m4d-system of the coordinator cluster. This secret holds the credentials. 
- Credential manager can access the secret (based on datasetId/catalogId)
- Credential manager returns the credentials to manager 
- Manager stores the credentials in vault and writes the path in a relevant blueprint.

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -351,11 +350,11 @@ func (r *M4DApplicationReconciler) ConstructDataInfo(datasetID string, req *modu
 		setCondition(input, datasetID, req.Actions[app.Read].Message, "Policy Compiler", true)
 	}
 	// Call the DataCatalog service to get info about the dataset
-	if err := GetConnectionDetails(datasetID, req, input); err != nil {
+	if err := GetConnectionDetails(datasetID, "", req, input); err != nil {
```
shlomitk1, line 353: shouldn't be dataCatalogId instead?

manager/controllers/app/policy_compiler_interface.go
```diff
@@ -12,6 +12,7 @@ import (
 )
 
 // ConstructApplicationContext constructs ApplicationContext structure to send to Policy Compiler
+// TODO: Add DataCatalogID
```
shlomitk1, line 15: Do we expect policyManager to access the catalog searching for data, or we need to add some data context here? We have got it anyway from the catalog connector.

manager/controllers/app/policy_compiler_interface.go
```diff
@@ -33,11 +34,11 @@ func ConstructApplicationContext(datasetID string, input *app.M4DApplication, op
...
 
 // TempLookupPolicyDecision is a temporary function used to simulate the policy compiler, until the policy
-// compiler is updated to receive the externalDataset in place of the datasetID for external data sets.
+// compiler is updated to handle uncataloged data.
 // TODO: Update LookupPolicyDecisions after policy compiler is updated
-func TempLookupPolicyDecision(externalDataset app.ExternalDataContext, policyCompiler pc.IPolicyCompiler, req *modules.DataInfo, input *app.M4DApplication, op pb.AccessOperation_AccessType) error {
-	// For now approve if destGeo is the same as the residency of the data set
-	if req.Geo == string(externalDataset.Residency) {
+func TempLookupPolicyDecision(dataset app.DataContext, policyCompiler pc.IPolicyCompiler, req *modules.DataInfo, input *app.M4DApplication, op pb.AccessOperation_AccessType) error {
+	// For now approve if destGeo is Turkey
```
shlomitk1, line 40: I suggest to remove this hard-coded function, and compare instead req.Geo with whatever you got from the catalog connector in DatasetDetails (there is some Geo there as well)

samples/kubeflow/m4dapplication.yaml
```diff
@@ -14,6 +14,7 @@ spec:
     role: Security
   data:
     - dataSetID: "{\"ServerName\":\"cocoMDS3\",\"AssetGuid\":\"ASSET_ID\"}"
+      cataloged: true
```
shlomitk1, line 17: If 'cataloged' were to be replaced with 'catalog_requested', the default false would remove the need to change the existing resources. Kind of backward compatible, in this sense.

simanadler: This flag is just meant to indicate that the manager shouldn't look for the data in the enterprise catalog.  Your approach seems to indicate that the user is requesting that the data be cataloged.  He is not.  The manager's logic is making that decision. For example, if a user is writing to a data store outside of the data lake (future scenario) we do not want to catalog it.

manager/apis/app/v1alpha1/m4dapplication_types.go
simanadler: fixed

manager/controllers/app/m4dapplication_controller.go
simanadler: Agreed, but it only solves part of the problem. There is no API to get the credential path from the inline catalog.

manager/apis/app/v1alpha1/m4dapplication_types.go
simanadler: No, it's not.  The entire ExternalDataContext is commented out.  I will remove it.  Recurrence is in DataContext now.

manager/apis/app/v1alpha1/m4dapplication_types.go
simanadler: @shlomitk1 Not sure what you mean by "state of the asset".
Regarding multiple catalogs, I will let @roee88 answer.  I think for the ingest scenario we will assume two catalogs.  One is the "inline" catalog via which the data sets to be ingested are provided, and the other is the enterprise catalog where the data sets are registered after they are successfully ingested into the data lake.

@roee88 Should the data sets be removed from the "inline" catalog after they are registered in the enterprise catalog?

manager/apis/app/v1alpha1/m4dapplication_types.go
simanadler: @shlomitk1 Agreed.  I'm waiting for the "inline" catalog design details before finalizing this.  @roee88 is working on this I believe.

manager/controllers/app/m4dapplication_controller.go
simanadler: The data owner provides the credentials, not an admin.  If I understood @roee88 correctly, the plan is for the data owner to put the credentials directly into the M4D vault (like the data user does for the credentials to external systems such as the enterprise catalog) and put the link to them into the inline data catalog.  The manager then needs to get that link, which is passed to the modules in the blueprint.
@roee88 Did I understand correctly?

manager/apis/app/v1alpha1/m4dapplication_types.go
shlomitk1: @simanadler I mean that if dataCatalogId=inline, you don't need the 'cataloged' field. 
My concern is the user experience. Along with specifying the data identificators, you also need to explicitly state that it is cataloged, since a dataset is not cataloged by the default. I would like to remove the need in this property entirely, or to change the default (cataloged -> external)

manager/controllers/app/m4dapplication_controller.go
simanadler: Yes, fixed

manager/controllers/app/policy_compiler_interface.go
simanadler: @roee88 said that the policyManager will get the data from the catalog in the same way that it does it today when using egaria and OPA.

manager/controllers/app/m4dapplication_controller.go
shlomitk1: If you mean VaultConnector that retrieves the credentials that the user has put in the user vault, then the connector will return the credentials, not the link. It will access the user vault instance according to the link, retrieve the credentials and pass to manager that will register them in the internal vault and pass to the modules. Data owner does not have an access to m4d-system and m4d vault. That's why VaultConnector uses an additional set of variables for indicating url, token, root (USER_VAULT_TOKEN, USER_VAULT_PATH, etc.)

manager/controllers/app/m4dapplication_controller.go
simanadler: Code now takes the credentials from the inline catalog and copies them into vault

manager/apis/app/v1alpha1/m4dapplication_types.go
simanadler: Removed

manager/controllers/app/policy_compiler_interface.go
simanadler: Will do so when policies are finalized

manager/config/crd/bases/app.m4d.ibm.com_m4dapplications.yaml
```diff
@@ -78,11 +84,13 @@ spec:
                     required:
                     - protocol
                     type: object
+                  recurrence:
```
ronenkat, line 87: I suggest we move this to the catalog as metadata, I don't think the manager should be aware of this. Only the code that interacts with the "copy module" should be aware of this.

go.mod
```diff
@@ -28,7 +29,7 @@ require (
 	github.com/stretchr/testify v1.6.1
 	golang.org/x/net v0.0.0-20201110031124-69a78807bb2b // indirect
 	golang.org/x/sys v0.0.0-20201119102817-f84b799fce68 // indirect
-	google.golang.org/grpc v1.28.1
+	google.golang.org/grpc v1.31.1
```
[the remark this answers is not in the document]
shlomitk1: I'm reverting these changes.

manager/config/crd/bases/app.m4d.ibm.com_m4dapplications.yaml
shlomitk1: @simanadler ^

shlomitk1: We could move the recurrence property to the inline catalog as a part of dataset metadata. However, if there is a request to read data from the enterprise catalog and refresh with some frequency, do we have a way to do this? Can any metadata be inserted to the catalog? I guess not.

shlomitk1 (approved):

manager/config/crd/bases/app.m4d.ibm.com_m4dapplications.yaml
ronenkat: @shlomitk1 I think that scenario is out of scope for our initial implementation.
But generally catalogs support metadata, which can include refresh frequency, and we cab get it from there.

shlomitk1: OK, removing the recurrence definition from m4dapplication.

manager/controllers/app/modules/moduleinterface.go
```diff
@@ -17,22 +17,30 @@ type Transformations struct {
 	Allowed            bool
 	EnforcementActions []pb.EnforcementAction
 	Message            string
+	// geography relevant for the read/write operation
+	// indicates where the workload runs (for read access) or where to write the data to (for copying data to another location)
+	Geo string
 }
```
simanadler, line 23: Why is this in the Transformations?  This is contextual information that was one of the parameters used to make the decision as to whether to do the transformation, and while it's true that the manager will have to decide where to run the transformation ... I'm not sure why Geo should be part of the Transformations.

manager/controllers/app/modules/moduleinterface.go
```diff
@@ -17,22 +17,30 @@ type Transformations struct {
...
 
 // DataInfo defines all the information about the given data set
+// TODO: Add support for recurrence in modules
 type DataInfo struct {
 	// Data asset unique identifier, not necessarily the same string appearing in the resource definition
 	AssetID string
-	// Processing geography
-	Geo string
 	// Application interface
```
simanadler, line 30: If there are no transformations, then we don't know the processing Geo?

manager/controllers/app/moduleinstance.go
```diff
@@ -75,21 +90,12 @@ func (m *ModuleManager) GetCopyDestination(item modules.DataInfo, destinationInt
 	}, nil
 }
 
-// SelectModuleInstances selects the necessary read/copy/write modules for the blueprint
+// SelectModuleInstances selects the necessary read/copy/write modules for the blueprint for a given data set
+// Write path is not yet implemented
 func (m *ModuleManager) SelectModuleInstances(item modules.DataInfo) ([]modules.ModuleInstanceSpec, error) {
 	instances := make([]modules.ModuleInstanceSpec, 0)
```
simanadler, line 96: @shlomitk1 I see you have merged the two functions I had into one very large function, whose flow is hard to understand in my opinion.

manager/controllers/app/moduleinstance.go
```diff
@@ -104,41 +110,65 @@ func (m *ModuleManager) SelectModuleInstances(item modules.DataInfo) ([]modules.
...
+			return instances, errors.New(readSelector.GetError())
+		}
+		// logic for deciding whether copy module is required
+		copyRequired, interfaces, actionsOnCopy = m.getCopyRequirements(item, readSelector)
+		// is copy allowed?
+		if !item.Actions[pb.AccessOperation_WRITE].Allowed {
+			return instances, errors.New(app.CopyNotAllowed)
+		}
+	} else {
+		// COPY FLOW
```
simanadler, line 147: @shlomitk1 Any time we're not doing READ we're assuming it's an ingest flow? 
I don't think anyone reading this code could figure out that we're loading new data into the environment.  "Copy flow" comment is not sufficient to understand this.

manager/controllers/app/moduleinstance.go
```diff
@@ -104,41 +110,65 @@ func (m *ModuleManager) SelectModuleInstances(item modules.DataInfo) ([]modules.
...
+		copyRequired = true
+		actionsOnCopy = item.Actions[pb.AccessOperation_WRITE].EnforcementActions
+		interfaces = []*app.InterfaceDetails{item.AppInterface}
+		// is copy allowed?
+		if !item.Actions[pb.AccessOperation_WRITE].Allowed {
```
simanadler, line 152: Comment says copy, but action is write.

manager/controllers/app/m4dapplication_controller_test.go
```diff
@@ -469,28 +487,26 @@ var _ = Describe("M4DApplication Controller", func() {
...
-
-				// Check the generated blueprint
-				// There should be a single read module with two datasets
-				numReads := 0
-				for _, step := range blueprint.Spec.Flow.Steps {
-					if step.Template == readPathModule.Name {
-						numReads++
-						Expect(len(step.Arguments.Read)).To(Equal(2))
-					}
+			// Check the generated blueprint
```
simanadler, line 497: @shlomitk1 Why are these changes in this PR?

manager/controllers/app/m4dapplication_controller_test.go
```diff
@@ -606,44 +708,32 @@ var _ = Describe("M4DApplication Controller", func() {
...
-				// Extra long timeout as deploying the arrow-flight module on a new cluster may take some time
-				// depending on the download speed
-				By("Expecting M4DApplication to eventually be ready")
-				Eventually(func() bool {
-					Expect(k8sClient.Get(context.Background(), applicationKey, application)).To(Succeed())
-					return application.Status.Ready
-				}, timeout*10, interval).Should(BeTrue(), "M4DApplication is not ready after timeout!")
-			}
+			// The plotter has to be created
+			plotter := &apiv1alpha1.Plotter{}
```
simanadler, line 718: @shlomitk1 Why is this part of this PR?

manager/controllers/app/suite_test.go
```diff
@@ -100,6 +101,7 @@ var _ = BeforeSuite(func(done Done) {
 		Expect(err).ToNot(HaveOccurred())
 		err = NewBlueprintReconciler(mgr, "Blueprint", fakeHelm).SetupWithManager(mgr)
 		Expect(err).ToNot(HaveOccurred())
+		SetupPlotterController(mgr, local.NewManager(mgr.GetClient(), "m4d-system"))
```
simanadler, line 104: @shlomitk1 Part of this PR?

simanadler (commented):
@shlomitk1 Please see my comments about the changes you made

roee88 (changes requested):
Did not yet reviewed the files that Shlomit just commented on

manager/controllers/app/modules/moduleinterface.go
shlomitk1: There is no meaning to the Geo outside the operation context. This structure is used for communication with policy manager. As we discussed with @rohithdv, there is a single geography field passed to the connector - workload geography for the read, and the destination for the write. In the ingest flow we don't have a workload geography at all...

manager/controllers/app/modules/moduleinterface.go
shlomitk1: Workload geography is stored in the geo field of the READ action structure. It's where we want to read the data.

manager/controllers/app/moduleinstance.go
shlomitk1: OK. I'll return the flow argument to the structure.

manager/controllers/app/moduleinstance.go
shlomitk1: I'll change the comment.

manager/controllers/app/m4dapplication_controller_test.go
shlomitk1: I've fixed the tests along the way.

manager/controllers/app/suite_test.go
shlomitk1: The tests did not test what was expected. I can split into two PRs.

manager/apis/app/v1alpha1/m4dapplication_types.go
```diff
@@ -8,18 +8,24 @@ import (
...
 	// +required
 	// +kubebuilder:validation:MinLength=1
 	DataSetID string `json:"dataSetID"`
 
 	// IFdetails indicates the protocol and format expected by the data by the Data Scientist's application
 	// +required
 	IFdetails InterfaceDetails `json:"ifDetails"`
+
+	// DestinationCatalog indicates the catalog service and catalog in which a data set being imported should be registered.
+	// This parameter should contain datacatalogservice, and catalog.  This should be populated only when Cataloged is false.
```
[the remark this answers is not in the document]
shlomitk1: I've removed the Cataloged field, forgot to edit the comment.

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -283,31 +278,66 @@ func (r *M4DApplicationReconciler) reconcile(applicationContext *app.M4DApplicat
...
-	// policies for READ operation
-	if err := LookupPolicyDecisions(datasetID, r.PolicyCompiler, req, input, pb.AccessOperation_READ); err != nil {
-		return AnalyzeError(input, r.Log, datasetID, err, "Policy Compiler")
-	}
-	if !req.Actions[app.Read].Allowed {
-		setCondition(input, datasetID, req.Actions[app.Read].Message, "Policy Compiler", true)
-	}
+// ConstructDataInfo gets a list of governance actions FOR READING and data location details for the given dataset and fills the received DataInfo structure
+// Assumes that the data is registered in the data catalog.
+func (r *M4DApplicationReconciler) ConstructDataInfo(req *modules.DataInfo, input *app.M4DApplication, clusters []multicluster.Cluster) error {
```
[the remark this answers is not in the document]
shlomitk1: OK.

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -283,31 +278,66 @@ func (r *M4DApplicationReconciler) reconcile(applicationContext *app.M4DApplicat
...
+	if req.WorkloadSelected {
+		workloadGeography, err = r.GetProcessingGeography(input)
+		if err != nil {
+			setCondition(input, datasetID, err.Error(), "", true)
+			return err
+		}
+
+		if req.Actions[pb.AccessOperation_READ], err = LookupPolicyDecisions(datasetID, r.PolicyCompiler, input,
+			pb.AccessOperation{Type: pb.AccessOperation_READ, Destination: workloadGeography}); err != nil {
+			return AnalyzeError(input, r.Log, datasetID, err, "Policy Compiler")
```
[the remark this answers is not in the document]
shlomitk1: Yes, there is an issue to remove it.

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -283,31 +278,66 @@ func (r *M4DApplicationReconciler) reconcile(applicationContext *app.M4DApplicat
...
+		}
+		if !req.Actions[pb.AccessOperation_READ].Allowed {
+			setCondition(input, datasetID, req.Actions[pb.AccessOperation_READ].Message, "Policy Compiler", true)
+		}
+	}
+
+	// policies for WRITE operation
+	// if workload exists, the write will be done to the workload geography depending on the decision logic
+	// Otherwise, select any of the available geographies
+	if req.WorkloadSelected {
```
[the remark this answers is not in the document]
shlomitk1: WorkloadSelected is determined using the relevant part of the cluster selector (not the cluster name, the workload selector)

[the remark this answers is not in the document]
roee88: I know, my comment is that the selection for the target should respect the cluster name if specified.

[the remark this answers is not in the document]
shlomitk1: ok, got it.

manager/apis/app/v1alpha1/m4dapplication_types.go
```diff
@@ -8,18 +8,24 @@ import (
...
+	// +optional
+	DestinationCatalog string `json:"destinationCatalog,omitempty"`
```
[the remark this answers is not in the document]
shlomitk1: I don't follow the structure. In the ingest case, why the asset will be under read?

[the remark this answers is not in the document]
roee88: Removed to avoid confusion with flow

```yaml
selector: ...
appInfo: ...
data:
  - asset: ...
    preferences:
      protocol: ...
      format: ...
      copy:
        catalog: ... # effectively an explicit copy
        ...
```

manager/controllers/app/modules/moduleinterface.go
shlomitk1: Transformations name was misleading. Renamed to Operations. The idea is that the controller collects in advance governance actions for possible read and write operations, and selects modules later based on this information.

manager/controllers/app/moduleinstance.go
shlomitk1: Tried to simplify the function. The idea is that the read module is selected when there is a workload, and copy is selected both when the workload requires a copy, or the user requested explicitly (ingest flow). No need for special casing.

manager/apis/app/v1alpha1/m4dapplication_types.go
```diff
@@ -8,18 +8,24 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// DataContext indicates data set chosen by the Data Scientist to be used by his application,
+// DataContext indicates data set chosen by the Data Scientist to be read from or written to by his application,
 // and includes information about the data format and technologies used by the application
 // to access the data.
 type DataContext struct {
-	// DataSetID is a unique identifier of the dataset chosen from the data catalog for processing by the data user application.
+	// DataSetID is a unique identifier of the dataset either chosen from the enterprise data catalog for processing by the data user application,
```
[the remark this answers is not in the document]
shlomitk1: OK

manager/apis/app/v1alpha1/m4dapplication_types.go
```diff
@@ -8,18 +8,24 @@ import (
...
+	// or the source location of a new data set being imported into the environment.  This parameter should contain datacatalogservice, catalog, datasetID.
```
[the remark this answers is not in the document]
shlomitk1: OK

manager/controllers/app/moduleinstance.go
```diff
@@ -68,87 +83,115 @@ func (m *ModuleManager) GetCopyDestination(item modules.DataInfo, destinationInt
...
-	actionsOnRead := item.Actions[app.Read]
-	if !actionsOnRead.Allowed {
-		return instances, errors.New(actionsOnRead.Message)
+func (m *ModuleManager) selectCopyModule(item modules.DataInfo, appContext *app.M4DApplication, readSelector *modules.Selector) (*modules.Selector, error) {
+	// logic for deciding whether copy module is required
+	var interfaces []*app.InterfaceDetails
+	var copyRequired bool
+	var actionsOnCopy []pb.EnforcementAction
+
+	if readSelector != nil {
```
simanadler, line 121: Please add comments.  I assume this means that if there's no workload we are assuming a copy flow.  right?

manager/controllers/app/moduleinstance.go
```diff
@@ -68,87 +83,115 @@ func (m *ModuleManager) GetCopyDestination(item modules.DataInfo, destinationInt
...
+		copyRequired, interfaces, actionsOnCopy = m.getCopyRequirements(item, readSelector)
+	} else if item.Context.Requirements.Copy.Required {
```
simanadler, line 123: Please add a comment - User explicitly requested that the data be copied rather than read.

manager/controllers/app/moduleinstance.go
```diff
@@ -201,9 +247,10 @@ func GetSupportedReadSources(module *app.M4DModule) []*app.InterfaceDetails {
 	return list
 }
 
-// check whether copy is required
+// check whether IMPLICIT copy is required
```
simanadler, line 250: Please remove this comment.  This function isn't only for implicit copy

manager/apis/app/v1alpha1/m4dapplication_types.go
```diff
@@ -8,6 +8,39 @@ import (
...
+// DataRequirements structure contains a list of requirements (interface, need to catalog the dataset, etc.)
+type DataRequirements struct {
+	// Interface indicates the protocol and format expected by the data user
+	// +required
+	Interface InterfaceDetails `json:"interface"`
+
+	// CopyRequrements include the requirements for copying the data
+	// +optional
+	Copy CopyRequirements `json:"copy,omitempty"`
+}
```
simanadler, line 42: I personally think we should explicitly state the flow - read, copy, or write (future) - rather than inferring it.  Since I'm in the minority though, I'll go with the flow.  Pun intended.  :)

manager/controllers/app/moduleinstance.go
shlomitk1: Added comments. We are not assuming copy, it should be explicitly requested if no workload exists.

manager/controllers/app/moduleinstance.go
shlomitk1: User can request copy even if the data is read by the workload (performance reasons, etc.)

manager/controllers/app/moduleinstance.go
shlomitk1: It is. Changed the function name.

manager/apis/app/v1alpha1/m4dapplication_types.go
shlomitk1: We will add write when coming to implement the write path. From what I understood (@roee88 - please correct me if I'm wrong), the most common case is that there is always a workload that needs to access the data, but sometimes it is added later in the process. So copy can be prepared in advance, and when the workload is added, a read will be instantiated on top of the existing blueprint (the blueprint will be updated to include read as well). Ingest is not special in this sense - data needs to be prepared and registered, but another module can be added later to access the data (audit, data quality, ...)

roee88 (approved):
@simanadler - does it really make sense to ingest to a cluster data that workloads that run in that cluster won't be able to read?

PR #251, opened 2021-01-04
Credentials management: a minor refactoring

shlomitk1:
Signed-off-by: SHLOMITK@il.ibm.com 

This PR is related to https://github.com/IBM/the-mesh-for-data/issues/221

In some use-cases (e.g. ingest external data) the credentials are already registered. This PR changes the Credentials field to be a path to the credentials after they have been registered. This change makes the code more suitable for both cases (credentials registration by the user or by the manager).


simanadler (dismissed):
Yes, this makes sense.  Thanks!
However, @roee88 is working on the inline catalog and I understand from him that datasets there will have a link to the credentials, rather than the credentials themselves.  GetCredentials in data_catalog_interface.go returns a Credentials structure and not a string for the link.  How would we address this issue?

manager/controllers/app/data_catalog_interface.go
```diff
@@ -60,9 +60,7 @@ func GetCredentials(datasetID string, req *modules.DataInfo, input *app.M4DAppli
 		DatasetId: datasetID,
 		AppId:     utils.CreateAppIdentifier(input)})
 	if err != nil {
-		return err
+		return nil, err
 	}
-	req.Credentials = dataCredentials.DeepCopy()
-
-	return nil
+	return dataCredentials.Creds.DeepCopy(), nil
```
simanadler, line 65: @roee88 is working on the inline catalog and I understand from him that datasets there will have a link to the credentials, rather than the credentials themselves. GetCredentials returns a Credentials structure and not a string for the link. Will the link be returned in the GetConnectionDetails or will we need a change here?

shlomitk1: Dataset credentials can be registered in many different ways: a kubernetes secret, in a vault instance (which?), etc. We need a generalized approach how to access credentials, as well as credentials taxonomy which we have now encoded in a received structure. So far, the idea was that each credential manager connector accesses the credentials (via user vault instance, for example) and passes them to the manager to be stored inside vault. 
I don't like the way it works now (a credentials manager accesses some vault instance to read the credentials, then passes them to the manager through an insecure grpc connection to be stored again in another vault instance). Ideally, each connector will use the secret provider instance to register the credentials and will return the link. So the answer is yes, we will need a change here. This requires a change to the connectors (credential manager), and a corresponding change in protobuf.

simanadler: @roee88 Is this something that is included in the work you're doing on the connectors?

roee88: I'm not sure that I follow. There is no intention to let the connectors store the credentials anywhere. The **connectors** return the dataset credentials (not link) regardless of how the remote system works or what the manager does to distribute these to modules. Do you want to raise this in a community meeting?

PR #254, opened 2021-01-06
fleet: added manager+agent playground

hunchback:
used for evaluation of rancher fleet


roee88:
The main problem with adding fleet integration is that we use the status of remote resources as-is, which is unsupported by fleet. I'm hoping that we could change it with https://github.com/IBM/the-mesh-for-data/discussions/266 (see last item in https://github.com/IBM/the-mesh-for-data/discussions/266#discussioncomment-272974).

PR #267, opened 2021-01-11
Single blueprint namespace for plotter created blueprints

### LINKED ISSUE #257: Use a single namespace for blueprint deployment
Using a single namespace for all blueprints makes many aspects of the mesh for data easier:
- easier to find created blueprints
- easier to remove blueprints (without some finalizers being blocked by namespace deletion)
- more restrictive RBAC. The mesh does not need full access but can be restricted to certain namespaces.

The initial proposal is to use one namespace for blueprints different from m4d-system. (e.g. m4d-blueprints)

froesef:
Closes #257

### DIFF
--- a/manager/Makefile
+++ b/manager/Makefile
@@ -75,6 +75,7 @@ undeploy-crd:
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
 deploy: manifests $(TOOLBIN)/kustomize $(TOOLBIN)/kubectl config/default/kustomization.yaml
 	$(TOOLBIN)/kubectl create namespace m4d-system || true
+	$(TOOLBIN)/kubectl create namespace m4d-blueprints || true
 	cd config/manager && $(ABSTOOLBIN)/kustomize edit set image controller=${IMG}
 	$(TOOLBIN)/kustomize build config/default | $(TOOLBIN)/kubectl apply -f -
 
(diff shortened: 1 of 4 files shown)


manager/Makefile
```diff
@@ -115,6 +118,7 @@ deploy_multi_it: manifests $(TOOLBIN)/kustomize $(TOOLBIN)/kubectl config/defaul
 	make build-ca-certs
 	make deploy-custom-ca-secret
 	$(TOOLBIN)/kubectl create namespace ${KUBE_NAMESPACE} || true
+	$(TOOLBIN)/kubectl create namespace m4d-blueprints || true
```
roee88, line 121: Would it make sense to put this is a Namespace resource in the deployment? Otherwise you also require users to create this namespace

manager/controllers/app/plotter_controller.go
```diff
@@ -192,16 +191,14 @@ func (r *PlotterReconciler) reconcile(plotter *app.Plotter) (ctrl.Result, []erro
...
-			randomNamespace := "m4d-" + random
 			blueprint := &app.Blueprint{
 				TypeMeta: metav1.TypeMeta{
 					Kind:       "Blueprint",
 					APIVersion: "app.m4d.ibm.com/v1alpha1",
 				},
 				ObjectMeta: metav1.ObjectMeta{
 					Name:        plotter.Name,
-					Namespace:   randomNamespace,
+					Namespace:   "m4d-blueprints",
```
roee88, line 201: constant?

roee88 (commented):
Does it depend on Shlomit's PR to be merged first?

manager/Makefile
froesef: I put the namespace creation into the deploy target as well. I think it should be created at the same place as the m4d-system namespace that's why I put it here. If we want to create the namespace as part of a yaml we should also do this for the m4d-system namespace. But I'd wait for that until the installation refactoring is done.

roee88: I might be wrong but I don't think it's the same case as m4d-system (which is where you will install a chart to). This namespace is an installation side effect. No one would manually create it and neither should our scripts.

roee88 (approved):

Issue #258, opened 2021-01-06
Push custom razee code to satcon-client-go library


froesef:
There is certain functionality that is implemented directly in the razee.go file as some features are not yet implemented in the satcon-client-go library. The goal is to move the razee specific functions (e.g. setSubscription) to the external library to improve the library and reduce our code base.

PRs can be found here https://github.com/IBM/satcon-client-go/pulls

Functions to be exported:
- [x] setSubscription mutation
- [x] clusterByName query
- [x] groupByName query
- [x] signIn mutation
- [x] AuthClient for RazeeLocalRoundTripper
- [x] AuthClient for usage with API key (PR under review)

froesef:
All changes have been pushed upstream and approved.

PR #272, opened 2021-01-12
Use new satcon-client-go library methods

froesef:
Implements some parts of #258


pkg/multicluster/razee/razee.go
```diff
@@ -280,54 +18,21 @@ func (t *RazeeGithubApiRoundTripper) RoundTrip(request *http.Request) (*http.Res
...
 }
 
-//nolint:golint,unused
-type RazeeLocalRoundTripper struct {
-	url      string
-	login    string
-	password string
-	token    string
-}
+const TokenValidityDuration = 40 * time.Minute
```
roee88, line 21: Did you check if the token has expiration (`exp`) within it? If so we can remove this hardcoding in a future PR.

froesef: The token is unfortunately only a token. 40 minutes is an estimation. I checked the code and in Razee it defines `240m` as validity period. We can set it to 200 min to let it be valid for more but I would not stress it and set it to 240.

froesef: Let's handle the details of this in the external PR to the satcon library https://github.com/IBM/satcon-client-go/pull/24

tomersolomon1 (approved):

PR #279, opened 2021-01-14
Use a human readable name for release names of deployed modules

### LINKED ISSUE #278: Human readable name for releases of blueprint steps
Currently the release name of a blueprint step is a hash of the step name. This is error prone if blueprint steps have the same name and not easy to read for users.

The proposal is to have a human readable name that is based on the blueprint name and step name. It should also be limited to 63 characters as some K8s objects that may be created in modules have this limit.

froesef:
Fixes #278


shlomitk1 (approved):

roee88:
I didn't see an issue for adding labels, did I miss it?

PR #282, opened 2021-01-17
OPA rules for READ and WRITE actions

ritwikc:
OPA rules created for READ and WRITE actions.

### DIFF
--- a/connectors/opa/lib/evaluate_opa.go
+++ b/connectors/opa/lib/evaluate_opa.go
@@ -66,7 +66,7 @@ func EvaluatePoliciesOnInput(inputMap map[string]interface{}, opaServerURL strin
 	log.Println("opaServerURL")
 	log.Println(opaServerURL)
 
-	res := performHTTPReq(standardClient, opaServerURL+"v1/data/user_policies", httpMethod, inputJSON, contentType)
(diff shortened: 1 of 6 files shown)


ronenkat:
@ritwikc can you reference the issue this PR is addressing? is there a document describing the intent?
Please tie the policies to a sample, they are too specific to be a default.
Can this PR be split into infrastructure that is missing and sample policies? does it make sense?

connectors/opa/lib/evaluate_opa.go
```diff
@@ -66,7 +66,7 @@ func EvaluatePoliciesOnInput(inputMap map[string]interface{}, opaServerURL strin
 	log.Println("opaServerURL")
 	log.Println(opaServerURL)
 
-	res := performHTTPReq(standardClient, opaServerURL+"v1/data/user_policies", httpMethod, inputJSON, contentType)
+	res := performHTTPReq(standardClient, opaServerURL+"v1/data/read_policy", httpMethod, inputJSON, contentType)
```
shlomitk1, line 69: It's confusing to have both read and write policies defined in "read_policy".

ritwikc:
This PR is replaced by #298

Issue #280, opened 2021-01-17
make install-tools: github.com/googleapis/gnostic@latest does not contain OpenAPIv2


hunchback:
CI breaking due to the above, and issue can be reproduced locally by running:

```
git fetch --all
git checkout upstream/master
make uninstall-tools
make install-tools
```

error message:

```
go mod tidy
go: downloading k8s.io/apimachinery v0.20.1
go: downloading github.com/go-logr/logr v0.2.0
go: downloading golang.org/x/text v0.3.4
go: downloading k8s.io/klog/v2 v2.4.0
go: downloading github.com/golang/protobuf v1.4.3
go: downloading sigs.k8s.io/structured-merge-diff/v4 v4.0.2
go: downloading golang.org/x/tools v0.0.0-20210106214847-113979e3529a
go: downloading github.com/googleapis/gnostic v0.4.1
go: finding module for package github.com/googleapis/gnostic/OpenAPIv2
go: downloading github.com/googleapis/gnostic v0.5.3
go: downloading k8s.io/utils v0.0.0-20201110183641-67b214c5f920
go: downloading istio.io/api v0.0.0-20210107192247-a205c627e4b9
go: downloading k8s.io/kube-openapi v0.0.0-20201113171705-d219536bb9fd
go: downloading istio.io/gogo-genproto v0.0.0-20210107204948-697d6f912366
go: downloading github.com/kr/pretty v0.2.0
github.com/ibm/the-mesh-for-data/pkg/helm imports
        helm.sh/helm/v3/pkg/action imports
        k8s.io/cli-runtime/pkg/resource imports
        github.com/googleapis/gnostic/OpenAPIv2: module github.com/googleapis/gnostic@latest found (v0.5.3), but does not contain package github.com/googleapis/gnostic/OpenAPIv2
make: *** [hack/make-rules/tools.mk:114: install-tools] Error 1
```

Issue #238, opened 2020-12-24
Define and enforce a taxonomy


roee88:
An end to end solution requires a taxonomy to be defined and used across different components. The Mesh for Data includes core components (mainly defined in `manager` and `pkg`) and plugins (`connectors` and `modules`). Currently, partial and hardcoded definitions are defined across both core components and plugins.

Design document is available in: https://docs.google.com/document/d/1sfdcNI6PTIYWsSIznpNqJNQCdVaVFchYB2vljzUrlew/edit?usp=sharing

simanadler:
I'm assuming we will define our taxonomy as json schema files.  Thus, we will need a json schema validator to use within our code.  For gaoling code, is this an acceptable library or is there something else you recommend?  https://github.com/xeipuuv/gojsonschema. @roee88

roytman:
Done

Issue #163, opened 2020-11-24
Replace Egeria lab with a different catalog


roee88:
Egeria lab is not suitable as an entry point catalog. The installation is difficult (as reported by several users #162) and the basic experience turned out to be lacking. 

I would like to consider using a different catalog to showcase the M4D project. The options are to pick an existing open source data catalog or to write our own basic catalog that will also demonstrate our minimal requirements from a data catalog.

I'm currently examining both options with concrete plans to:
1. Experiment with  https://marquezproject.github.io/marquez/ and https://www.amundsen.io/amundsen/
2. Design a basic catalog and estimate how much efforts it would be to implement it

Input from people who did similar evaluations would be helpful 

cc @rohithdv

simanadler:
> I'm currently examining both options with concrete plans to:
> 
>     1. Experiment with  https://marquezproject.github.io/marquez/ and https://www.amundsen.io/amundsen/
> 
@roee88 I believe that @etykha and @rohithdv did a search for open source data catalogs and I believe they experimented with both of these options.

roee88:
@rohithdv please share

rohithdv:
Yes we had explored Amundsen earlier along with Ety and Rishi. I could fish out the following docs from our chats earlier. 

Initial comparison table between Amundsen and Egeria (created by Ety),
https://ibm.ent.box.com/notes/692966717227

Lyft/Amundsen
Quick Start guide: https://github.com/lyft/amundsen/blob/master/docs/installation.md#bootstrap-a-default-version-of-amundsen-using-docker

Document created by Rishi Saket on using Amundsen: 

[Amundsen Exploration.zip](https://github.com/IBM/the-mesh-for-data/files/5597287/Amundsen.Exploration.zip)

@roee88 @simanadler

rohithdv:
Basically the summary of our findings was Amundsen is more powerful than Egeria with respect to support for connectors to different datasources. But we went with Egeria as it was IBM supported project.

Amundsen is also featured in number of blogs like [this one which I found interesting](https://eugeneyan.com/writing/data-discovery-platforms/?utm_campaign=Data_Elixir&utm_source=Data_Elixir_310)

As for Marquez, we have not explored it earlier. It looks promising at a first look. Thanks.

roee88:
[...]

PR #290, opened 2021-01-19
Builtin catalog and credentials manager

roee88:
Related issues (not resolved, it's just a step towards it): https://github.com/IBM/the-mesh-for-data/issues/163 https://github.com/IBM/the-mesh-for-data/issues/238

This PR introduces the first version of a built-in catalog and credentials manager. It allows managing data assets as Kubernetes `Asset` CRDs and credentials as Kubernetes `Secret` resources. See the `example` folder for example resources.

What is included:
- `manifests/taxonomy.yaml`: OpenAPI spec definition for authentication methods (e.g., apikey, password) and for connection types (e.g., s3, kafka). It's not a taxonomy solution but rather checking if OpenAPI can be used as a technical tool as part of a solution. The role of Katalog is to be a catalog and credentials that is auto generated from a reference taxonomy. The work on taxonomies is in very early stages, see https://github.com/IBM/the-mesh-for-data/issues/238 and the design document within.
- `manifests/spec.yaml`: the source for the `Asset` CRD leveraging the above spec.
- Connector code that maps from `Asset` and `Secret` to the hardcoded structures of the current connectors API that is defined in `pkg/connectors/proto`

Please avoid reviewing generated files (`docs/`, `*.gen.go`, `*.gen.yaml`)

** https://github.com/IBM/the-mesh-for-data/issues/280 is required before this can be tested**


connectors/katalog/example/asset.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: katalog.m4d.ibm.com/v1alpha1
+kind: Asset
+metadata:
+  name: data-csv
+spec:
+  secretRef: 
+    name: data-csv-creds
+  details:
```
simanadler, line 8: Would endpoint be a better name than details?

connectors/katalog/example/asset.yaml
```diff
@@ -0,0 +1,29 @@
...
+    dataFormat: csv
+    connection:
+      type: s3
+      s3:
+        endpoint: "http://minio:9000"
+        bucket: demo
+        objectKey: data.csv
+  security:
```
simanadler, line 16: Why is this called security?

connectors/katalog/example/asset.yaml
```diff
@@ -0,0 +1,29 @@
...
+    geography: us-south
```
simanadler, line 17: These are data set level metadata? If so, perhaps we should call them something like datasetMetadata and componentMetadata?

simanadler (commented):
Looks good.  Just a couple of minor comments

connectors/katalog/example/asset.yaml
roee88: I actually created my own structure but then decided to have the initial PR be a copy of what we have today in the proto (so the word details is from there) and evolve it later though a design doc. The exception is the word security that was originally called dataset metadata, I changed to avoid confusion with the k8s resource metadata but I will switch back .

connectors/katalog/example/asset.yaml
roee88: Will do

simanadler (approved):

PR #292, opened 2021-01-19
helm push fix: x509: certificate relies on legacy Common Name field

hunchback:
fix bug in helm push to local registry, fixing `make run-integration-tests` stage


roee88 (approved):
Awesome

Issue #65, opened 2020-10-15
Enforcement Action restructure


etykha:
Enforcement Action restructure to more hierarchical format, to reduce number of enforcement actions in the taxonomy and enable different levels of specification for desired governance action. 
This issue has an assumption that enforcement action implementation (e.g. masking algorithm) should be consistent both with policy-manager enforcement implementation behind a policy if such exists, like in WKC or any other policy manager that has also PEP, and between our modules used in the same application.
Full discussion of this issue and suggested solution here: https://ibm.ent.box.com/notes/730332532445

roee88:
Closing, will be addressed as part of the taxonomy work.

PR #302, opened 2021-01-24
fix: code generated from proto breaks compilation

### LINKED ISSUE #291: Code generation from protos is broken
`make -C pkg/connectors build` generates go code from protos.

We relied on the old code generation by protoc-gen-go v1.3.5 because we embedded these structures into CRDs. This is not a supported pattern and any new version of go protobuf generate code that we can't handle.  

Due to dependency issues we moved to protoc-gen-go v1.4.x and need to remove the embedding of generated code in our CRDs.

roee88:
Fixes #291

Changes:
- Re-generated protobuf with `make -C pkg/connectors build`
- Actions in M4DModule are now SupportedActions instead of EnforcementActions  (m4dmodule_types.go); in the spirit of https://github.com/IBM/the-mesh-for-data/issues/65. 
- Using RawExtension instead of proto generated structures in CRDs:
  - Added pkg/serde with RawExtension (de)serialization
  - Transformations passed to module arguments are now RawExtensions (blueprint_types.go)
  - Connection in app.DataStore is now RawExtension (datacatalog_types.go)
  - Extracting from catalog connector only required fields (data_catalog_interface.go)


shlomitk1 (approved):

simanadler:
Can you explain why the move to raw extensions?

roee88:
> Can you explain why the move to raw extensions?

It is described in the issue. Newer versions of protoc-gen-go generate code that can't be used inside CRDs (the structures are not allowed to be copied). We had this issue for a while now and from time to time someone accidently bumped the version and caused our code to break. We recently updated protoc-gen-go because we had to (dependency issues) and this issue came back.

The use of RawExtension is a way to include arbitrary objects in CRDs. See the issue for a link to its use in opa.

PR #307, opened 2021-01-27
charts: add opa-connector to m4d helm

hunchback:
this PR adds the opa-connector as default for the m4d helm chart, the location of the opa-connector docker images can be controlled via values.yaml:

```
image:
  opaConnector: <image location goes here>
```

other than that there is not change to the m4d helm chart.

### DIFF
--- a/charts/m4d/templates/opa_connector.yaml
+++ b/charts/m4d/templates/opa_connector.yaml
@@ -1,0 +1,45 @@
+# Copyright 2020 IBM Corp.
+# SPDX-License-Identifier: Apache-2.0
+
+{{- if .Values.global.opaConnector }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: opa-connector
+  labels:
+    app: opa-connector
+spec:
+  selector:
+    matchLabels:
+      app: opa-connector
+  replicas: 1
+  template:
+    metadata:
+      annotations:
+        sidecar.istio.io/inject: "true"
+      labels:
+        app: opa-connector
+        m4d.ibm.com/componentType: connector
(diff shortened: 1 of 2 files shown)


roee88:
@hunchback can you please add a flag to enable/disable it? (default enabled)

ronenkat:
@hunchback how will this work with tags? 
Its not just opa connector related, but a more general issue.

roee88:
@ronenkat I was thinking that at a later stage we would use something similar to Istio with `hub` and `tag` values.

hunchback:
> @hunchback can you please add a flag to enable/disable it? (default enabled)

yup

hunchback:
> @hunchback how will this work with tags?
> Its not just opa connector related, but a more general issue.

this is controller via the image.opaConnector ... the default value uses the `latest` tag but by providing a custom value you can use any tagged docker image

charts/m4d/templates/opa_connector.yaml
```diff
@@ -0,0 +1,43 @@
...
+      annotations:
+        sidecar.istio.io/inject: "true"
+      labels:
+        app: opa-connector
+        m4d.ibm.com/componentType: connector
+    spec:
+      containers:
+      - name: opa-connector
+        image: {{ .Values.image.opaConnector }}
+        imagePullPolicy: Always
```
froesef, line 26: Can we also have a variable for this? Especially for production environments `IfNotPresent` might be a better option. That way it's configurable.

hunchback: sure

froesef (approved):

PR #318, opened 2021-02-01
Remove dev requirements for local registry setup

### LINKED ISSUE #310: Remove dev requirements for local registry setup
As part of testing a local kind registry is set up. Currently:

Certificates are generated and configured because helm can't push to insecure registries helm/helm#6324. Our current script works in Ubuntu and Mac but not in other linux distros and only when setting GODEBUG=x509ignoreCN=0 environment variable.
The developer needs to modify /etc/hosts in his own machine

The proposed alternative is to always use http://localhost:5000 as the registry URL just like in the instructions in https://kind.sigs.k8s.io/docs/user/local-registry/ and https://helm.sh/docs/topics/registries/. This removes all of the dev requirements described above.

The main issue is that the manager pod will try to pull from localhost:5000 and fail because in the docker network the registry actually runs in kind-registry:5000. We can bypass it by patching the manager with an init container that runs socat to redirect localhost:5000 to kind-registry:5000 as part of the integration-tests kustomization layer (and/or Helm).

roee88:
Fixes #310 

- [X] Changed local registry address to localhost:5000 
- [X] Changed integration-tests Kustomize layer to run a `alpine/socat` sidecar to allow manager container to access to the registry over localhost
- [X] docs: Removed `Pre requisite` section from `/contribute/code/flow.md`
- [X] Fixed formatting issues in create_kind.sh using [shell-format](https://github.com/mvdan/sh#related-projects) (it was hard to modify it without). This seems to follow the google style guide. If vim `=G` is preferred then I can use that for now until we have a standard defined although I don't see common support for it in other IDEs.
- [x] Mirror alpine/socat to quay.io

### DIFF
--- a/hack/tools/kind-config.yaml
+++ b/hack/tools/kind-config.yaml
@@ -9,4 +9,4 @@ nodes:
 containerdConfigPatches:
 - |-
   [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:5000"]
-    endpoint = ["https://kind-registry:5000"]
+    endpoint = ["http://kind-registry:5000"]
--- a/manager/config/integration-tests/manager_patch.yaml
+++ b/manager/config/integration-tests/manager_patch.yaml
@@ -31,13 +31,3 @@ spec:
           value: "false"
         - name: GODEBUG
           value: "x509ignoreCN=0"
-        volumeMounts:
-        - mountPath: /etc/ssl/certs/ca-certificates.crt
-          subPath: ca-certificates.crt
-          name: ca-cert
-          readOnly: true
-      volumes:
-      - name: ca-cert
-        secret:
-          defaultMode: 420
-          secretName: registry-ca-cert
--- a/manager/config/integration-tests/manager_patch.yaml
+++ b/manager/config/integration-tests/manager_patch.yaml
@@ -13,7 +13,7 @@ spec:
     spec:
       containers:
       - name: integration-tests
-        image: alpine/socat:1.0.5
+        image: alpine/socat:latest
         command:
         - socat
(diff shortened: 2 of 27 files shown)


manager/config/integration-tests/manager_patch.yaml
```diff
@@ -12,26 +12,22 @@ spec:
   template:
     spec:
       containers:
+      - name: integration-tests
+        image: alpine/socat:1.0.5
+        command:
+        - socat
+        - TCP4-LISTEN:5000,fork
+        - TCP4:kind-registry:5000
```
shlomitk1, line 20: What is the purpose of an additional container?

hack/tools/kind-config.yaml
```diff
@@ -9,4 +9,4 @@ nodes:
 containerdConfigPatches:
 - |-
   [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:5000"]
-    endpoint = ["https://kind-registry:5000"]
\ No newline at end of file
+    endpoint = ["http://kind-registry:5000"]
```
shlomitk1, line 12: I am not sure I understand. So the kind-registry is still created but localhost:5000 is pointing to kind-registry:5000?

manager/config/integration-tests/manager_patch.yaml
roee88: As described in the issue and the PR head, it makes the registry reachable via localhost.

shlomitk1: ok, got it

shlomitk1 (approved):

froesef (approved):
Very nice! Finally we don't need the internal certificates any more!

Issue #317, opened 2021-02-01
helm: use values to support the different deployment modes 


roee88:
The manager includes a lot of Kustomization layers

For different deployment cases:

- [ ] single cluster ([default](https://github.com/IBM/the-mesh-for-data/tree/master/manager/config/default))
- [ ] multicluster coordinator (https://github.com/IBM/the-mesh-for-data/tree/master/manager/config/multi-controller)
- [ ] multicluster remote (https://github.com/IBM/the-mesh-for-data/tree/master/manager/config/multi-follower)

For testing:

- [ ] integration-tests (https://github.com/IBM/the-mesh-for-data/tree/master/manager/config/integration-tests)
- [ ] multi-integration-tests
- [ ] just the movement-controller (https://github.com/IBM/the-mesh-for-data/tree/master/manager/config/movement-controller)

Also stand alone layers for optional features:

- [ ] control-plane-security (https://github.com/IBM/the-mesh-for-data/tree/master/manager/config/control-plane-security)

All of the above should be supported by the Helm chart. That is, proper values can be passed to create a deployment that is equivalent to the use case.  

The first step is to implement passing of command line arguments to manager (@hunchback)

PR #328, opened 2021-02-03
charts: add to m4d extraArgs/extraEnvs

hunchback:
this PR addresses a subset of issue #317 

to enable future customization of the the manager to support various deployment modes I have added to values.yaml two expansion/customization options:
- extraArgs - add extra arguments to be passed to the manager container
- extraEnvs - pass extra environment variables for the manager container

the basic approach has been taken from what is generally accepted for helm charts such as is used for example in the elastic/elasticsearch helm chart https://github.com/elastic/helm-charts/blob/master/elasticsearch/values.yaml#L35 (many more similar examples exist)

to verify that there has been no degradation due to the change I ran:

```
aidan@AIDANSHRIBMAN-SDFUE6N:~/projects/the-mesh-for-data/charts$ helm install m4d m4d         --namespace m4d-system
NAME: m4d
LAST DEPLOYED: Wed Feb  3 10:32:57 2021
NAMESPACE: m4d-system
STATUS: deployed
REVISION: 1
TEST SUITE: None
```

### DIFF
--- a/charts/m4d/templates/manager.yaml
+++ b/charts/m4d/templates/manager.yaml
@@ -407,8 +407,9 @@ spec:
     spec:
       containers:
       - args:
-        - --enable-leader-election
-        - --enable-all-controllers
+      {{- if .Values.extraArgs }}
+{{ toYaml .Values.extraArgs | indent 8 }}
+      {{- end }}
         env:
         - name: VAULT_TOKEN
           valueFrom:
@@ -421,6 +422,9 @@ spec:
           value: {{ .Values.image.mover }}
         - name: IMAGE_PULL_POLICY
           value: Always
+      {{- if .Values.extraEnvs }}
+{{ toYaml .Values.extraEnvs | indent 8 }}
+      {{- end }}
         image: {{ .Values.image.manager }}
         imagePullPolicy: Always
         name: manager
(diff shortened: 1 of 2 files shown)


charts/m4d/templates/manager.yaml
```diff
@@ -405,13 +405,19 @@ spec:
       - args:
         - --enable-leader-election
         - --enable-all-controllers
```
roee88, line 407: I think that this overrides the other flags.

roee88 (changes requested):
I suggest to either template it better not always include --enable-all-controllers or to change the manager main.go to take the parameters differently (e.g., remove enable-all and make it implicit if no other enable flag os set or something else that makes sense).

hunchback:
> I suggest to either template it better not always include --enable-all-controllers or to change the manager main.go to take the parameters differently (e.g., remove enable-all and make it implicit if no other enable flag os set or something else that makes sense).

will move these into the values.yaml so that they can be overridden

charts/m4d/templates/manager.yaml
hunchback: ok, will move them to values.yaml

ronenkat:
@hunchback looks good, only need to add @roee88's input.

roee88 (approved):

PR #334, opened 2021-02-03
Application labels for plotter, blueprint and modules

### LINKED ISSUE #305: Add labels to deployed modules
Currently the release name includes namespace and name of M4DApplication but adding it as proper labels would help both for debugging and for isolation.

froesef:
Fixes #305

### DIFF
--- a/manager/apis/app/v1alpha1/m4dapplication_types.go
+++ b/manager/apis/app/v1alpha1/m4dapplication_types.go
@@ -201,3 +201,9 @@ type M4DApplicationList struct {
 func init() {
 	SchemeBuilder.Register(&M4DApplication{}, &M4DApplicationList{})
 }
+
+const (
+	ApplicationClusterLabel = "app.m4d.ibm.com/app_cluster"
+	ApplicationNamespaceLabel = "app.m4d.ibm.com/app_namespace"
+	ApplicationNameLabel = "app.m4d.ibm.com/app_name"
+)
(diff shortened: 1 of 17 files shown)


manager/apis/app/v1alpha1/m4dapplication_types.go
```diff
@@ -201,3 +201,9 @@ type M4DApplicationList struct {
 func init() {
 	SchemeBuilder.Register(&M4DApplication{}, &M4DApplicationList{})
 }
+
+const (
+	ApplicationClusterLabel = "app.m4d.ibm.com/app_cluster"
```
roee88, line 206: For consistency with our other labels this should be `m4d.ibm.com/appCluster`. Same for the others.

roee88 (changes requested):
Don't use explicit labels in the modules. Instead expect getting labels in the value in iterate over them. See example in https://github.com/IBM/the-mesh-for-data-flight-module/blob/9992d2226d374ffd7a2eb2753eb50d8e1c669fa4/helm/afm/templates/deployment.yaml#L21

roee88 (approved):

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -403,7 +403,11 @@ func AnalyzeError(app *app.M4DApplication, log logr.Logger, assetID string, err
 }
 
 func ownerLabels(id types.NamespacedName) map[string]string {
-	return map[string]string{OwnerLabelKey: id.Namespace + "." + id.Name}
+	return map[string]string{
+		OwnerLabelKey: id.Namespace + "." + id.Name,
```
shlomitk1, line 407: Adding other labels makes OwnerLabelKey redundant, but ok - I'll handle it later.

shlomitk1 (approved):

Issue #335, opened 2021-02-04
Configure Vault to support K8S auth method in multi-cluster setup


revit13:
This is part of the first milestone described in https://github.com/IBM/the-mesh-for-data/issues/140

Given there is one Vault instance in the coordinator cluster as described in https://github.com/IBM/the-mesh-for-data/issues/140; the Vault instance should be enabled with K8S auth method for every cluster.

More Information on the Vault setup can be found in the following links: 
- https://github.com/hashicorp/vault-plugin-auth-kubernetes/issues/19.
- https://learn.hashicorp.com/tutorials/vault/kubernetes-external-vault?in=vault/kubernetes
- https://www.vaultproject.io/docs/auth/kubernetes

This work will include:

- Adding Ingress to  Kind coordinator cluster  as described in https://kind.sigs.k8s.io/docs/user/ingress/.
- Add documentation to the project webpage describing the requirement of having Ingress to the vault instance in the coordinator cluster in multi-cluster setup.
- Enable K8S auth method for every cluster. 
- Enable secret engine of type kv where the credentials will reside.
- Add policy to allow access permissions to all secrets in the path above.
 - For each cluster add a role to bind all services in the  modules namespace to the policy above. The request for the role uses the auth path configured for each cluster.
- Add documentation to project webpage to describe the process of enabling new cluster to access the Vault in the coordinator cluster.

roytman:
@revit13, what's the status of the issue? can it be closed ?

Noamstrauss:
any updated? @revit13

revit13:
@roytman  @Noamstrauss  the status is as follows: The steps to configure Vault in multi-cluster deployment are documented in the [site ](https://fybrik.io/v1.3/tasks/multicluster/) under `Configure Vault for multi-cluster deployment` section but there is no automatic process for the configuration and thus the steps need to be done manually. That said, [this script](https://github.com/fybrik/fybrik/blob/master/hack/tools/configure_vault_kind_multi.sh) contains helper functions for configuring Vault. it is currently working on two KinD clusters created by https://github.com/fybrik/fybrik/blob/master/hack/setup-local-multi-cluster.sh but can serve as a basis to configure Vault on other clusters.

PR #341, opened 2021-02-07
[RFC] Configure Vault to support K8S auth method in multi-cluster setup

revit13:
I appreciate feedback regarding  this implementation of https://github.com/IBM/the-mesh-for-data/issues/335.
It uses a new configure_vault_kind.sh script under hack/tools/ directory which intended to be ran after the control-plane (including Vault) was initiated (using hack/install.sh for example). I took most of the script functionality from  secret-provider Vault utility functions.
The script enables K8S auth method in a new path called kind for a single cluster setup and an additional  K8S auth method in a new path called control for a multi-cluster setup thus it does not interfere with the current secret-provider implementation (which enables K8S auth method in a path called kubernetes)
An alternative approach would be to have the m4d-controller configure Vault for the clusters using parameters from config-maps.

Note: this PR does not include the documentation changes required by this issue. If this approach is OK then I will do that in a separate follow up patch.

### DIFF
--- a/hack/tools/configure_vault_kind.sh
+++ b/hack/tools/configure_vault_kind.sh
@@ -1,0 +1,72 @@
+#!/usr/bin/env bash
+# Copyright 2020 IBM Corp.
+# SPDX-License-Identifier: Apache-2.0
+
+set -x
+
+op=$1
+
+source ./common.sh
+source ./vault_utils.sh
+
+: ${KUBE_NAMESPACE:=m4d-system}
+: ${INGRESS_ADDRESS:=localhost:80}
+: ${KIND_CLUSTER_KUBE_HOST:=https://kind-control-plane:6443}
+: ${CONTROL_CLUSTER_KUBE_HOST:=https://control-control-plane:6443}
+: ${MODULE_NAMESPACE:="m4d-blueprints"}
+: ${SECRET_PATH:=m4d/dataset-creds}
+: ${ROLE:=module}
+
+
+# $1 - cluster name
+# $2 - vault root token
+# $3 - vault address
+# $4 - kube host of the cluster
+enable_k8s_auth_for_cluster() {
+        kubectl config use-context kind-"$1"
+        kubectl create ns $KUBE_NAMESPACE || true
+        kubectl apply -f vault_auth_sa.yaml -n "$KUBE_NAMESPACE"
+        enable_k8s_auth "$2" "$1" vault-auth "$KUBE_NAMESPACE" "$3" "$4"
+}
+
+# $1 - vault root token
+# $2 - vault address
+configure_vault() {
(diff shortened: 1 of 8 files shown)


roee88 (approved):
LGTM. Notes for future:

1. You may want to talk to @froesef on how Razee controllers communicate over the two kind clusters (i.e., using the same network without requiring an ingress). Seems like you have the same requirements.*
2. Once we have user instructions I think that having a series of vault commands would fit better than curl (Tomer had it earlier I believe) 
3. Consider if for ci/testing running it as a Kubernetes job rather than port forwarding in the background would be cleaner.

\* Obviously a real deployment will need an ingress but it would be one using subdomain rather than path (because vault requires it) and you can't do it in kind.

shlomitk1 (approved):

revit13:
I've changed to vault commands instead of curl.

ronenkat (approved):

hack/tools/configure_vault_kind.sh
```diff
@@ -0,0 +1,70 @@
...
+        ;;
+    multi)
+        header_text "Configure Vault on kind multi-cluster"
+        kubectl config use-context kind-control --namespace=$KUBE_NAMESPACE
+        export VAULT_ADDR=$INGRESS_ADDRESS
+        export VAULT_TOKEN=$(kubectl get secrets vault-unseal-keys -n $KUBE_NAMESPACE -o jsonpath={.data.vault-root} | base64 --decode)
+        bin/vault login "$VAULT_TOKEN"
+        enable_k8s_auth_for_cluster control "$CONTROL_CLUSTER_KUBE_HOST"
+        enable_k8s_auth_for_cluster kind "$KIND_CLUSTER_KUBE_HOST"
+	configure_vault
```
froesef, line 52: minor thing. The indentation seems to be different from the other rows here.

hack/tools/vault_utils.sh
```diff
@@ -0,0 +1,69 @@
...
+# Copyright 2020 IBM Corp.
+# SPDX-License-Identifier: Apache-2.0
+
+: ${PORT_TO_FORWARD:=8200}
+
+# $1 role name
+# $2 policy name
+# $3 path to auth method
+# $4 bound_service_account_namespaces
+create_role() {
```
froesef, line 11: This role creation will be moved to the application controller at some point when every module will be deployed with it's own service account right?

hack/tools/configure_vault_kind.sh
revit13: Thanks I'll fix it.

hack/tools/vault_utils.sh
revit13: Right.

PR #345, opened 2021-02-08
Consolidate implicit copy modules. One for stream and one for batch

### LINKED ISSUE #333: Merge implicit copy modules
Currently there are 3 implicit copy modules. There should only be two:
- one for batch
- one for streams

The goal is to support multiple connection types within the the same chart. E.g. S3 and jdbc for source/target.

froesef:
Closes #333


roee88 (approved):
LGTM. There will be blank lines in the template result but it shouldn't bother anyone. Before merging, I think that we should create a M4DModule for these and place it here. Am I missing something? @froesef

Issue #314, opened 2021-02-01
Add new data catalog to Helm charts 


eletonia:
To deploy Mesh for Data on the Operate First cluster, we need to package all non-cluster-wide components in Helm charts. This includes the new data catalog which replaces Egeria. @roee88

Issue #348, opened 2021-02-09
Improve Helm charts


roee88:
- [x] #351 CRDs generated directly to `charts/m4d-crd` (using controller-gen)
- [x] #368 RBAC not generated from annotations and instead maintained directly in `charts/m4d` (allows different RBAC for coordinator/remote as described in https://github.com/IBM/the-mesh-for-data/issues/304)
- [x] #364 Manager resources splitted to different files and use Helm recommend templating
- [x] https://github.com/IBM/the-mesh-for-data/pull/390 Generate webhook resources into the chart
- [x] configmap m4d-config is currently not part of the chart, should be added
- [x] configmap cluster-metadata is currently not part of the chart, should be added (only if razee is enabled?)
- [x] Profiles: coordinator, remote, ~~integration-tests~~ (https://github.com/IBM/the-mesh-for-data/issues/317)
- [x] #374 Integrate multi cluster setup
- [x] Flag to skip cluster scoped resouces (manager-auth-proxy-*, manager-cr, manager-crb, webhook-configs) or change the crd chart to base chart that includes all cluster scoped resources. @eletonia 
- [ ] #450 Add Istio resources when Istio is available
- [ ] Remove secret provider deployment
- [x] Rework OPA server deployment
- [ ] Remove kustomize config from manager 
- [ ] Change root Makefile to deploy with the charts
- [ ] Workflow for chart validation (See https://github.com/mesh-for-data/charts#2-add-github-workflow-for-testing-the-chart)
- [ ] Update developers docs 
- [ ] Set version 0.1.0 in charts
- [ ] Fine grained RBAC
- [ ] A direct way to disable motion controllers / to deploy just motion controllers (currently possible using manager.enable and manager.overrideArgs). Do we want a `motion` section next to `coordinator` and `worker`?
- [ ] Vault address without "m4d-system" in values.yaml

roee88:
> Flag for applying all, cluster scoped only, or namespace scoped only

@eletonia will this meet the requirements for deploying to operate first?

eletonia:
@roee88 Yes we should be able to use Helm parameters and override them with [ArgoCD parameters](https://argoproj.github.io/argo-cd/user-guide/helm/#helm-parameters) to make ArgoCD only deploy the namespace scoped resources

roee88:
[...]

PR #349, opened 2021-02-09
charts: add katalog connector

roee88:
Fixes https://github.com/IBM/the-mesh-for-data/issues/314

- Added katalog CRD and RBAC resources (generated directly into charts)
- Added katalog connector resources (maintained directly in charts)
- Helm values now allow overriding global or per component values (inspired by the Helm defaults, Istio, cert-manager)

To be continued following https://github.com/IBM/the-mesh-for-data/issues/348

@eletonia for some reason I can't add you as reviewer, can you please join https://github.com/orgs/IBM/teams/m4d-team/members?


eletonia:
@roee88 I just requested to join the m4d-team organization

charts/m4d/values.yaml
```diff
@@ -1,24 +1,135 @@
...
+
+opaService:
+  # Set to false to skip deploying OPA service
+  enabled: true
+  # Image name or a hub/image[:tag]
+  image: "openpolicyagent/opa:latest"
+  # Overrides global.imagePullPolicy
+  imagePullPolicy: ""
+
+katalogConnector:
```
ronenkat, line 63: is the term katalogConnector is correct ? should it be katalogService instead? i belive it is not just the connector but also the CRD is derived from these flags.

ronenkat (commented):
there are multiple topics in the PR. Would be good to seperate in the future.

charts/m4d/values.yaml
roee88: I would like to keep xxxConnector for consistency. You comment implies that we should also add `katalog` for the CRDs/RBAC, which I considered but decided that it will just complicate UX because currently it has no use outside m4d when the connector is enabled.

ronenkat (approved):

eletonia (approved):
All files look good to me!

PR #351, opened 2021-02-09
charts: manager CRDs 

roee88:
Part of https://github.com/IBM/the-mesh-for-data/issues/348

Goal is to eventually remove the use of kustomize and manage everything in a high quality Helm chart. 

# Previous behavior

`make -C manager manifests` generates CRDs to `m4d/config/crds` and then `make -C manager charts-crd` uses kustomize to build it and copy the output to `charts/m4d-crd/templates/crds.yaml`

# Alternatives

## 1 - Behavior implemented in this PR

`make -C manager manifests` will output the CRDs to `charts/m4d-crd/templates` using controller-gen (currently in addition to generation into `config/manager/crds/base`).

```make
	$(TOOLBIN)/controller-gen $(CRD_OPTIONS) output:crd:artifacts:config=$(ROOT_DIR)/charts/m4d-crd/templates/ paths=./apis/...
```

Pros:
- No need to remember invoking another make target (`make charts-crd` is removed).

Cons:
- Once we will want advanced things like API conversion webhooks the generator will override our changes.

## 2

`make -C manager manifests`  runs `controller-gen schemapatch` to update the schemas of CRDs that reside in `charts/m4d-crd/templates`. This is similar to what cert-manager does.

```make
	$(TOOLBIN)/controller-gen schemapatch:manifests=$(ROOT_DIR)/charts/m4d-crd/templates/ output:dir=$(ROOT_DIR)/charts/m4d-crd/templates/ paths=./apis/... 
```

Pros: 
- No need to remember invoking another make target (`make charts-crd` is removed).
- Can later add things like conversion webhooks directly in Helm and the generator won't override them

Cons:
- When adding a new CRD type one needs to manually add it to `charts/m4d-crd/templates` (no need for schema information though). This means that markers that are unrelated to schema are ignored (I think just printer columns). Those initial CRD files can still be generated with controller-gen though. 

## 3

Keep using Kustomize for CRDs. `make -C manager manifests` generates CRDs to `m4d/config/crds` and then we immediately build a layer that uses it and copy the result to `charts/m4d-crd/templates/manager-crds.yaml`


froesef:
What changes do we have in the helm chart that are not generated by controller-gen? I think we should try to generate as much as possible from the code. Manual editing of the crd files or similar may lead to inconsistencies. Maybe there are different tags that we can use in the go code to achieve similar things. Also I think we should drop the `CRD_OPTIONS` variable and put the parameter explicitly into the `controller-gen` call. That way it's clear what is generated at what point.

roee88:
@froesef let's continue in the issue. I added https://github.com/IBM/the-mesh-for-data/issues/348#issuecomment-776652017.

Closing this until a choice is made.

manager/Makefile
```diff
@@ -42,6 +42,7 @@ generate: $(TOOLBIN)/controller-gen
 .PHONY: manifests
 manifests: $(TOOLBIN)/controller-gen
 	$(TOOLBIN)/controller-gen $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(TOOLBIN)/controller-gen crd:trivialVersions=true output:crd:artifacts:config=$(ROOT_DIR)/charts/m4d-crd/templates/ paths=./apis/...
```
froesef, line 45: Can you please also remove the `CRD_OPTIONS` above and remove the variable in general?

froesef (approved):

manager/Makefile
roee88: Sure (I didn't touch it because eventually it will be removed altogether but you're right)

PR #352, opened 2021-02-09
Register asset: pass a reference to user credentials to catalog connector

shlomitk1:
This PR passes a reference to user credentials to register an asset.
It also fixes an uninitialized map problem.


pkg/connectors/protos/register_asset_request.proto
```diff
@@ -14,4 +14,5 @@ message RegisterAssetRequest {
 	Credentials creds = 1;
 	DatasetDetails dataset_details = 2; 	
 	string destination_catalog_id = 3;
+	string app_id = 4;   // reference to user credentials
```
roee88, line 17: Remove the comment, it's what you use it for, not what it is.

roee88 (approved):

Issue #336, opened 2021-02-04
M4d-controller changes to support  the ability of modules to send requests to Vault 


revit13:
This is part of the first milestone described in https://github.com/IBM/the-mesh-for-data/issues/140.
It described the changes needed in the M4d-controller to support the ability of modules to send requests to Vault.

It includes:
- Adding new field to ClusterMetadata structure which is called VaultAuthPath to hold auth path needed for the modules in a cluster to use to authenticate to vault.
- Adding a new field to the Blueprint DataStore field, similar to the CredentialLocation field, to hold information needed for the modules to connect directly to Vault. (it will contain the auth path retrieved from the new field in ClusterMetadata structure described above.
- Preserve the ability of the modules to access Vault using the existing implementation of the secret-provider.

PR #361, opened 2021-02-11
Manager changes to support the ability of modules to send requests to…

revit13:
This patch implements adds support in the manager to allow modules to get datasets credentials directly from Vault instead of the secret provider as described in https://github.com/IBM/the-mesh-for-data/issues/336.
It does not change the existing credentialLocation field that is passed to the modules to access the datasets credentials using the secret-provider.

The patch was tested in a single cluster setup with the following modules:
- Arrow flight module which includes the changes to access Vault directly to retrieve the datasets credentials.
- Copy module which retrieves the datasets credentials using secretProvider.


manager/config/integration-tests/deployment_configmap.yaml
```diff
@@ -24,13 +24,14 @@ data:
   #EXTENSIONS_POLICY_MANAGER_NAME: "OPA"
 
   #internal vault related variables
-  VAULT_ADDRESS: "http://vault:8200/"
+  VAULT_ADDRESS: "http://vault.m4d-system:8200/"
```
shlomitk1, line 27: I don't see here any cluster reference. Does this address work when connecting from another cluster?

manager/controllers/app/moduleinstance.go
```diff
@@ -277,6 +289,9 @@ func (m *ModuleManager) SelectModuleInstances(item modules.DataInfo, appContext
 			m.Log.Info("Allocation failed: " + err.Error())
 			return instances, err
 		}
+                copyCluster, vaultAuthMethod, err := copySelector.SelectCluster(item, m.Clusters)
```
shlomitk1, line 292: 1. Why moved here? The err value is not checked.
2. vaultAuthMethod should not be returned by SelectCluster. Once you have a cluster, you can iterate over clusters to get its property.

manager/controllers/app/moduleinstance.go
```diff
@@ -311,6 +325,8 @@ func (m *ModuleManager) SelectModuleInstances(item modules.DataInfo, appContext
 		if err != nil {
 			return instances, err
 		}
+		readCluster, vaultAuthMethod, err := readSelector.SelectCluster(item, m.Clusters)
```
shlomitk1, line 328: Same here. Use readCluster to get other properties.

manager/controllers/app/moduleinstance.go
revit13: It is moved before the module arguments are set. I'll add a check and iteration.

manager/config/integration-tests/deployment_configmap.yaml
revit13: In deployment mode of multi-cluster setup this address should be the address of the Ingress to Vault. So it should be set based on how the clusters were configured - the Ingress/Route address for multi-cluster (we do not need Ingress in single cluster) and for single cluster vault.m4d-system is enough. I will need to clarify that in the the documentation.

manager/controllers/app/moduleinstance.go
revit13: OK, tnx

shlomitk1 (approved):

manager/config/crd/bases/app.m4d.ibm.com_blueprints.yaml
```diff
@@ -68,10 +68,32 @@ spec:
                                   format:
                                     description: Format represents data format (e.g. parquet) as received from catalog connectors
                                     type: string
+                                  vault:
```
ronenkat, line 71: Should Vault or all the fields should repeat for each action in the blueprint? could some part be factored out?

ronenkat (commented):
We should see if there is a more efficient way to encode how we use Vault.
Should we use "Vault" in the manager code? or should we use the term CredentialStore (or equivalent)?
I am not sure we should embed the implementation details into the CRD and manager, but the concepts, using terms like CredentialStore or Credential Manager.
@roee88 @froesef opinions?

manager/config/crd/bases/app.m4d.ibm.com_blueprints.yaml
froesef: The `authPath` should be different for every cluster. The `address` will probably be the same for each cluster. I'm not sure how we will configure `role`. But the `role` and `secretPath` might be the changing factor per blueprint.

froesef:
> We should see if there is a more efficient way to encode how we use Vault.
> Should we use "Vault" in the manager code? or should we use the term CredentialStore (or equivalent)?
> I am not sure we should embed the implementation details into the CRD and manager, but the concepts, using terms like CredentialStore or Credential Manager.
> @roee88 @froesef opinions?

I think using the term "vault" in the `Blueprint` is fine. The `Blueprint` CRD is pretty much an internal CRD anyways that will not be used by end users but more by developers debugging things. Thus I see it more as an implementation detail itself. Calling it credentials manager might be even misleading as the concepts that we're using here with `authPath`, `secretPath` etc are very much tailored towards vault.

manager/config/crd/bases/app.m4d.ibm.com_blueprints.yaml
roee88: I can see a far future where the address is different (e.g., when using on-prem vaults). Anyway, for what we have defined we can move address, authPath, and role one level up. Not sure if that's that important though. It should be decided by what sounds more convenient for module developers @cdoron

roee88 (approved):
As discussed, in the future move VaultAuthPath out of cluster-metadata

pkg/multicluster/razee/razee_cluster_manager.go
```diff
@@ -316,35 +319,57 @@ func channelName(cluster string, name string) string {
...
-	localAuth := &LocalAuthClient{
-		url:      url,
-		login:    login,
-		password: password,
-		token:    "",
+func NewRazeeManager(url string, login string, password string) (multicluster.ClusterManager, error) {
+	localAuth, err := local.NewClient(url, login, password)
+	if err != nil {
+		return nil, err
 	}
```
revit13, line 326: @froesef I'm not sure why this code is marked as changed as its already in master.

froesef (approved):

manager/config/crd/bases/app.m4d.ibm.com_blueprints.yaml
the-mesh-for-data: @roee88 : I think we could keep the current configuration as is, especially if this may also be the way to go for the "far future". We could always re-examine this approach if needed.

ronenkat (approved):

PR #363, opened 2021-02-11
Fix bug when running "make test" on os x

### LINKED ISSUE #216: test: vault connector errors on Mac

froesef:
Somehow during the merge of the mod files this bug snuck in again. 
@hunchback actually solved it some months ago in #217.

Fixes #216


roee88 (approved):
I assume that it persist after go mod tidy (since the original fix did and it's the same). What do you think od adding a check to the CI to ensure that it builds both on Linux and Mac? (Like they did in kubernetes/release)

froesef:
> I assume that it persist after go mod tidy (since the original fix did and it's the same). What do you think of adding a check to the CI to ensure that it builds both on Linux and Mac? (Like they did in kubernetes/release)

Yes. That might make sense. Although that could increase the CI runtime quite a bit. (depending if we only run `make verify` and `make test` or if we also run all the integration tests. (Which would basically test if kind/docker etc also runs on os x. So this might be obsolete.)

roee88:
@froesef I didn't mean to run anything on mac, just go build with GOOS=darwin. See https://github.com/kubernetes/release/pull/1863.

Issue #304, opened 2021-01-26
Minimize RBAC requirements


roee88:
Currently the manger requires cluster admin RBAC on all resource types (`*` on all). With the move to a single namespace for blueprints we can minimize this requirement. This issue is for making the blueprint controller namespace scoped and removing all unnecessary RBAC.

roee88:
Also need by default to have the vault secrets plugin bound to just m4d-system, with instructions to manually add other namespaces / make it cluster scoped.

roee88:
Depends on https://github.com/mesh-for-data/mesh-for-data/issues/671

roee88:
clusterScoped=false flag achieves this now

PR #368, opened 2021-02-14
rbac: removed kubebuilder:rbac markers

roee88:
Part of #348 

Source of truth is the RBAC files in config and once the move to Helm is complete then it charts. Does not include a fix for #304.


froesef:
I think you should also remove the `rbac:roleName=manager-role` part of the manifests target in the manager Makefile to make sure that the current roles in config are not overwritten.

froesef (approved):

PR #369, opened 2021-02-14
Gui update to support ingest flow

simanadler:
Addresses GUI associated with [Issue 221](https://github.com/IBM/the-mesh-for-data/issues/221)

Changes made:
* Compatible with updated version of M4DApplication that added ingest flow capabilities (already in master)
* Workload selector is now optional, as is required for the ingest flow


shlomitk1 (approved):
We'll also need to add: 
- cluster name for workload selection
- destination catalog for asset registration
- generated asset to show in the results

PR #375, opened 2021-02-17
Remove razee org id parameter

### LINKED ISSUE #356: multicluster: Get rid of RAZEE_ORG_ID parameter
The parameter is quite hard to figure out when not using the REST API. This makes it very hard for a user to find out and configure. (At least in local and iam use cases) 
There is a `me` method that can be used after login that also returns the organization id of a user. It can be used to retrieve this id whenever needed. This will simplify the configuration for users.

froesef:
Fixes #356

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,6 @@ require (
 	cloud.google.com/go v0.66.0 // indirect
 	emperror.dev/errors v0.7.0
 	github.com/IBM/dataset-lifecycle-framework/src/dataset-operator v0.0.0-20210201112050-4d77ac693ac1
-	github.com/IBM/go-sdk-core v1.1.0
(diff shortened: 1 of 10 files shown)


go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	emperror.dev/errors v0.7.0
 	github.com/IBM/dataset-lifecycle-framework/src/dataset-operator v0.0.0-20210201112050-4d77ac693ac1
 	github.com/IBM/go-sdk-core v1.1.0
```
roee88, line 9: Unrelated, but wanted to ask a while ago - why do you import an old version and not v4 like satcon-go does? (I see they recently added v5 though).

froesef: Good point. We actually don't need it. We should use the abstracted functionality from the library.

roee88 (approved):

Issue #337, opened 2021-02-04
Show the ability of modules to send request to Vault 


revit13:
This is the first milestone of the solution chosen for grading remote modules with access to dataset credentials as described in  https://github.com/IBM/the-mesh-for-data/issues/140.

#### High level description of the solution:
- Single Vault instance with support for multiple kubernetes clusters.
- Modules would obtain secrets from Vault by:
-- Use Vault client to get the credentials via HTTPS request
-- Develop Vault plugin in go language to retrieve credentials (third party systems, K8S secretes) to get the dataset credentials. 

This milestone targets the ability of modules to send requests directly from Vault as described in  https://github.com/IBM/the-mesh-for-data/issues/140.

#### Out of scope for this milestone
- The implementation of a new plugin and thus the way dataset credentials are inserted to Vault will not change. 
- Using secure connection to retrieve the credentials.

The ability of the modules to support retrieval of  dataset credentials using secret-provider should remain unchanged in this milestone.

# Sub issues

- [ ] #335  Configure Vault to support K8S auth method in multi-cluster setup
- [X] #336 M4d-controller changes to support the ability of modules to send requests to Vault 
**Documentation :**
- [ ] #421 Add documentation in README .md file for Vault configuration (temporary documentation)
- [X] #435  Document the new Vault fields sent to the modules

PR #377, opened 2021-02-18
Add website documentation to describe configuring Vault to support K8…

revit13:
…S auth method (#335)

Part of #337 - Show the ability of modules to send request to Vault


website/content/docs/setup/vault-configuration/index.md
```diff
@@ -0,0 +1,124 @@
+---
+title: Vault Configuration
```
ronenkat, line 2: We should reference Vault as the implementation details, not the concept.
How about using the term Credential Manager or credential store?
The text should reflect it as well.

website/content/docs/setup/vault-configuration/index.md
```diff
@@ -0,0 +1,124 @@
...
+---
+
+The {{< name >}} uses [HashiCorp Vault](https://www.vaultproject.io/) to provide [Modules](../../overview/modules) running in the remote clusters with the dataset credentials to execute their tasks.
+
+This page contains two sections:
+
+1. [This section](#configure-vault-kubernetes-auth-method-in-each-cluster) contains the required steps to enable the modules to authenticate to Vault in order for them to retrieve the dataset credentials.
+
+2. [This section](#add-vault-secret-plugin) describes how to add a new custom [Vault secret plugin](https://www.vaultproject.io/docs/plugin) to retrieve the dataset credentials from where they are stored.
+
```
ronenkat, line 14: where do you describe the architecture? concepts? etc... ? I can add that part if you like.

website/content/docs/setup/vault-configuration/index.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Configure Vault Kubernetes auth method in each cluster
+
```
ronenkat, line 17: need to describe the prerequisite for k8s auth, such as Vault having connectivity to the remote cluster.

website/content/docs/setup/vault-configuration/index.md
```diff
@@ -0,0 +1,124 @@
...
+Module authentication is done by configuring Vault to use [Kubernetes auth method](https://www.vaultproject.io/docs/auth/kubernetes) in each cluster. Using this method the modules can authenticate to Vault by providing their service account token. Behind the scenes Vault authenticates the token by submitting TokenReview request to the API server of the kubernetes cluster where the module is running. 
+
+Some of the steps described below are not specific to the {{< name >}} project but rather are Vault specific and can be found in Vault related online tutorials.
+
+It is assumed that Vault is deployed in the coordinator cluster in `m4d-system` namespace prior to executing the steps.
+
+1. Setup Ingress to Vault in the coordinator cluster for multi-cluster setup:
+In multi-cluster setup a kubernetes Ingress should be configured in the coordinator cluster to enable communication to Vault from remote clusters.
+The ingress should be deployed in the `m4d-system` namespace where Vault is deployed.
```
ronenkat, line 26: text for ingress, or reference to a script

website/content/docs/setup/vault-configuration/index.md
```diff
@@ -0,0 +1,124 @@
...
+
+2. Add Vault address to `m4d-config`:
+The address of Vault should be added to the kubernetes ConfigMap called `m4d-config` in `m4d-system` namespace in the coordinator cluster as follows:
+```bash
+VAULT_ADDRESS: <Vault address>
+```
+In a multi-cluster setup the Vault address should be the Ingress address described above while in single cluster it should set to: `http://vault.m4d-system:8200/`.
+
+**Pre-steps for configuring Vault:**
```
ronenkat, line 35: please add on which systems/version this was tested on.

ronenkat (commented):
@revit13 thank you.
is this a draft? as parts are missing? or do you want to merge and will have a follow-up PR to complement the info?

website/content/docs/setup/vault-configuration/index.md
revit13: Sure sounds good. Will change it.

website/content/docs/setup/vault-configuration/index.md
revit13: Thanks, if you could add it it would be great. This part  (architecture, concepts, etc... ) is still missing from this page.

website/content/docs/setup/vault-configuration/index.md
revit13: OK, will add it.

website/content/docs/setup/vault-configuration/index.md
revit13: OK, will add that.

website/content/docs/setup/vault-configuration/index.md
revit13: OK.

revit13:
@ronenkat This PR is actually a starting point for documenting the new functionality. This page is missing the context of where this credential store belongs to (as you mentioned - architecture, concepts etc). I added a link from the architecture page but that's not sufficient. I think it would make sense to wait until the architecture part will be complete before merging.
Also, there are two additional missing parts in the documentation that I can think of now: one is for the person who writes the modules - she should be familiar with the new fields passed to the module in order to use them to connect to Vault with a Vault client.  The documentation for the new fields is auto generated as part of the Blueprint but I think it can be helpful to add a wording about that in the Module page. If that makes sense I'll add that. Also, there is the part about the plugin that I am now working on and when I complete it I will have a better picture with regard to the instructions for the plugin writer. This part also involves instructions for Vault setup to use the new plugin.

website/content/docs/setup/vault-configuration/index.md
ronenkat: I will add it as a follow-up PR.

roee88:
@revit13 can we change this PR to draft and continue after we move to the new website?

revit13:
Sure will do that.

PR #378, opened 2021-02-18
Integrate razee configuration into helm chart

### LINKED ISSUE #374: Integrate multicluster configuration into helm chart
The multi cluster setup needs some secrets and mounts for the manager. This is currently done partly by kustomize configurations and by the local razee setup scripts. 
To simplify this it should be part of the new helm chart and support multiple razee authentication mechanisms.

froesef:
Fixes #374

### DIFF
--- a/charts/m4d/values.yaml
+++ b/charts/m4d/values.yaml
@@ -166,6 +166,16 @@ katalogConnector:
 
   affinity: {}
 
+razee: {}
+#  Razee local deployment requires the url, user and password parameters
+#  Razee deployment with oauth API key authentication requires the apiKey parameter
+#  Razee deployment with IBM Cloud Satellite Config requires the iamKey parameter
+#  url: ""
+#  user: ""
+#  password: ""
+#  apiKey: ""
+#  iamKey: ""
+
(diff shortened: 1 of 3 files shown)


charts/m4d/values.yaml
```diff
@@ -166,6 +166,16 @@ katalogConnector:
...
+razee: {}
+#  Razee local deployment requires the url, user and password parameters
+#  Razee deployment with oauth API key authentication requires the apiKey parameter
+#  Razee deployment with IBM Cloud Satellite Config requires the iamKey parameter
+#  url: ""
+#  user: ""
+#  password: ""
+#  apiKey: ""
+#  iamKey: ""
+
```
roee88, line 178: I don't think that this format will play nice with helm doc generators, but we can deal with that later. What's apiKey? I didn't see it used in our codebase

roee88 (approved):

charts/m4d/values.yaml
froesef: `apiKey` will be added as soon as the upstream PR to the library goes through.

PR #380, opened 2021-02-18
fix: update .travis.yml to bypass permission denied error

### LINKED ISSUE #379: Travis: Error: open /etc/docker/certs.d/ghcr.io: permission denied
Travis builds started to fail with `Error: open /etc/docker/certs.d/ghcr.io: permission denied`.
@cdoron fixed the same issue in the flight repo https://github.com/IBM/the-mesh-for-data-flight-module/pull/52/files

roee88:
Fix #379

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,7 +80,7 @@ jobs:
     deploy:
      skip_cleanup: true
      provider: script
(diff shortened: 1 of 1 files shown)


froesef (approved):
This seems like a hack to me but it seems to be used in many other projects that encountered the problem.

Issue #366, opened 2021-02-14
Deleting files from a persistent bucket


shlomitk1:
We have two use-cases for copying data:
- implicit copy: a temporary bucket is allocated. It is deleted (along with all objects) upon deletion of m4dapplication.
- ingesting data from another catalog: a persistent bucket is created. It is not deleted upon deletion of m4dapplication.

The current behavior is that upon deletion a batchtransfer deletes the objects inside a bucket, which causes problems in the second use-case and is not required for the first use-case.

froesef:
The `BatchTransfer` has a parameter `spec.noFinalizer` that can be set to true. That way the finalizer is not called and the data is not deleted.

PR #381, opened 2021-02-21
not deleting datasets upon copy deletion

shlomitk1:

Fixes https://github.com/IBM/the-mesh-for-data/issues/366


froesef (approved):
This will change the default functionality to never remove the files created once the implicit copy object is removed. Fine with me as it seems like we'll delete the whole bucket anyways.

PR #387, opened 2021-02-23
Add configmap m4d-config to the helm chart. 

revit13:
This patch adds the config-map m4d-config to the helm chart as part of https://github.com/IBM/the-mesh-for-data/issues/348.
It contains mostly the vault related values as a starting point.
The hard-coded values are intentional and are set for variables that expected to be removed such as the secret-provider.
I use internal and external terms for vault variables as is in the implementation today but that might also be changed due to changes in the code.

### DIFF
--- a/charts/m4d/templates/m4d-config.yaml
+++ b/charts/m4d/templates/m4d-config.yaml
@@ -1,0 +1,20 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: m4d-config
+data:
+  SECRET_PROVIDER_URL: "http://secret-provider.m4d-system:5555/get-secret"
+  SECRET_PROVIDER_ROLE: "demo"
+
+  {{- with .Values.vault }}
+  VAULT_ADDRESS: {{ .internal.vault_address }}
+  VAULT_DATASET_HOME: {{ .internal.dataset_home }}
+  VAULT_USER_HOME: {{ .internal.user_home }}
+  VAULT_DATASET_MOUNT: {{ .internal.dataset_mount }}
+  VAULT_USER_MOUNT: {{ .internal.user_mount }}
+  VAULT_TTL: {{ .internal.ttl }}
+  VAULT_AUTH: "kubernetes"
+  USER_VAULT_ADDRESS: {{ .external.user_vault_address }}
+  USER_VAULT_PATH: {{ .external.user_vault_path }}
+  {{- end }}
+
--- a/charts/m4d/templates/m4d-config.yaml
+++ b/charts/m4d/templates/m4d-config.yaml
@@ -3,18 +3,31 @@ kind: ConfigMap
 metadata:
   name: m4d-config
 data:
+  {{- with .Values.config }}
   SECRET_PROVIDER_URL: "http://secret-provider.m4d-system:5555/get-secret"
   SECRET_PROVIDER_ROLE: "demo"
-
-  {{- with .Values.vault }}
-  VAULT_ADDRESS: {{ .internal.vault_address }}
-  VAULT_DATASET_HOME: {{ .internal.dataset_home }}
-  VAULT_USER_HOME: {{ .internal.user_home }}
-  VAULT_DATASET_MOUNT: {{ .internal.dataset_mount }}
-  VAULT_USER_MOUNT: {{ .internal.user_mount }}
-  VAULT_TTL: {{ .internal.ttl }}
-  VAULT_AUTH: "kubernetes"
-  USER_VAULT_ADDRESS: {{ .external.user_vault_address }}
-  USER_VAULT_PATH: {{ .external.user_vault_path }}
+  VAULT_ADDRESS: {{ .vault.address }}
+  VAULT_DATASET_HOME: {{ .vault.datasetHome }}
+  VAULT_DATASET_MOUNT: {{ .vault.datasetMount }}
+  VAULT_USER_HOME: {{ .vault.userHome }}
+  VAULT_USER_MOUNT: {{ .vault.userMount }}
+  VAULT_TTL: {{ .vault.ttl }}
+  USER_VAULT_ADDRESS: {{ .vault.userVaultAddress }}
+  USER_VAULT_PATH: {{ .vault.userVaultPath }}
+  MODULES_VAULT_ROLE: {{ .vault.modulesRole }}
+  CATALOG_CONNECTOR_URL: {{ .catalog.connectorUrl }}
+  CATALOG_PROVIDER_NAME: {{ .catalog.name }}
+  EGERIA_SERVER_URL: {{ .catalog.url }}
+  MAIN_POLICY_MANAGER_CONNECTOR_URL: {{ .policyManager.connectorUrl }}
+  MAIN_POLICY_MANAGER_NAME: {{ .policyManager.name }}
+  OPA_SERVER_URL: {{ .policyManager.url }}
+  {{- if .policyManager.extensionManager.enabled }}
+  USE_EXTENSIONPOLICY_MANAGER: "true"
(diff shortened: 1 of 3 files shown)


charts/m4d/templates/m4d-config.yaml
```diff
@@ -0,0 +1,20 @@
...
+  {{- with .Values.vault }}
+  VAULT_ADDRESS: {{ .internal.vault_address }}
+  VAULT_DATASET_HOME: {{ .internal.dataset_home }}
+  VAULT_USER_HOME: {{ .internal.user_home }}
+  VAULT_DATASET_MOUNT: {{ .internal.dataset_mount }}
+  VAULT_USER_MOUNT: {{ .internal.user_mount }}
+  VAULT_TTL: {{ .internal.ttl }}
+  VAULT_AUTH: "kubernetes"
+  USER_VAULT_ADDRESS: {{ .external.user_vault_address }}
+  USER_VAULT_PATH: {{ .external.user_vault_path }}
```
shlomitk1, line 18: Do we still need the "external" parts?

shlomitk1: What are the plans for adding the rest of the config, e.g. catalog connector service, policy manager options?

revit13: @shlomitk1 I am not sure about external. I can revisit this once the dateset credential work will progress if that makes sense. The rest of the vars in the config should be added as well as mentioned it's a starting point for writing the config map...

roee88: I would remove internal/external. I was aiming for something like `config.vault.address`. Which of the parameters in values.yaml will survive after you finish the credentials work?

roee88: We also need the rest of the parameters like Shlomit mentioned. The configmap is unusable without them. You can keep them the same as in config/prod for this PR.

revit13: OK sure will add them.

revit13: Regarding Vault vars: I left the VAULT_DATASET_HOME and VAULT_DATASET_MOUNT as is as I would like to keep the current implementation in the manager that enables the secret engine in VAULT_DATASET_MOUNT path and adds policy to read the secrets. This policy will be used when configuring each cluster by adding a role for the modules to use this policy.
user credentials (VAULT_USER_HOME VAULT_USER_MOUNT) will be store as secrets so I removed them. 
I also removed VAULT_AUTH as its cluster specific.
I am not sure about USER_VAULT_PATH and USER_VAULT_ADDRESS - it is to external vault instance-- do we want to keep that?

shlomitk1: Keep VAULT_USER_HOME and VAULT_USER_MOUNT for now, as manager will store the secret in the vault (wkc connector does not work with secrets directly)

revit13: If the patch is OK I suggest to rename OPA_SERVER_URL to MAIN_POLICY_MANAGER_URL and EGERIA_SERVER_URL to CATALOG_PROVIDER_URL in m4d code.

charts/m4d/templates/m4d-config.yaml
```diff
@@ -0,0 +1,33 @@
...
+data:
+  {{- with .Values.config }}
+  SECRET_PROVIDER_URL: "http://secret-provider.m4d-system:5555/get-secret"
+  SECRET_PROVIDER_ROLE: "demo"
+  VAULT_ADDRESS: {{ .vault.address }}
+  VAULT_DATASET_HOME: {{ .vault.datasetHome }}
+  VAULT_DATASET_MOUNT: {{ .vault.datasetMount }}
+  VAULT_USER_HOME: {{ .vault.userHome }}
+  VAULT_USER_MOUNT: {{ .vault.userMount }}
+  VAULT_TTL: {{ .vault.ttl }}
```
shlomitk1, line 14: There was also VAULT_AUTH equal to "kubernetes". Either remove the code that is using it, or return this var.

revit13: The code that uses it relates to the secret-provider. I will remove it in a follow up patch after all the modules will call directly to vault to get the dataset creds.

roee88: The configmap includes all, but variables that are soon to be removed don't have to be templated.

revit13: OK sure. Hopefully it will be removed soon. If there will be a need I will add it.

revit13: Ok on second though I added VAULT_AUTH.

shlomitk1 (approved):

charts/m4d/values.yaml
```diff
@@ -188,4 +188,57 @@ clusterScoped: true
...
-integrationTests: false
\ No newline at end of file
+integrationTests: false
+
+# Values set in m4d-config ConfigMap.
+config:
+  # Vault related values set in m4d-config ConfigMap.
+  vault:
+    # Set to the Vault address.
+    address: "http://vault.m4d-system:8200/"
```
roee88, line 198: side note - we will need a way to remove m4d-system from the default value at some point.

charts/m4d/templates/m4d-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+  USER_VAULT_PATH: {{ .vault.userVaultPath }}
+  VAULT_MODULES_ROLE: {{ .vault.modulesRole }}
+  CATALOG_CONNECTOR_URL: {{ .catalog.connectorUrl }}
+  CATALOG_PROVIDER_NAME: {{ .catalog.name }}
+  EGERIA_SERVER_URL: {{ .catalog.url }}
+  MAIN_POLICY_MANAGER_CONNECTOR_URL: {{ .policyManager.connectorUrl }}
+  MAIN_POLICY_MANAGER_NAME: {{ .policyManager.name }}
+  OPA_SERVER_URL: {{ .policyManager.url }}
+  {{- if .policyManager.extensionManager.enabled }}
+  USE_EXTENSIONPOLICY_MANAGER: "true"
```
roee88, line 26: Why true

charts/m4d/values.yaml
```diff
@@ -188,4 +188,57 @@ clusterScoped: true
...
+    # Set to the Vault address that stores the credentials to data source for data assets.
+    userVaultAddress: "http://vault:8200/"
+    # Set to the path where credentials to data source for data assets are stored.
+    userVaultPath: "external"
+    # Set to the name of the role that modules use to access dataset credentials.
+    modulesRole: "module"
+
+  catalog:
+    # Set to the catalog URL.
+    url: "https://lab-core.egeria-catalog:9443"
```
roee88, line 218: `url` is specific configuration of the egeria connector, not of "catalog".  We also don't deploy egeria connector as part of the chart so it shouldn't be default.

charts/m4d/values.yaml
```diff
@@ -188,4 +188,57 @@ clusterScoped: true
...
+    # Set to the catalog connector URL.
+    connectorUrl: "egr-connector:50084"
+    # Set to the catalog provider name.
+    name: "EGERIA"
+
+  policyManager:
+    # Set to policyManager URL.
+    url: "opa:8181"
```
roee88, line 226: same comment about opa connector specific configuration

roee88 (approved):
Thanks. If you can please just use extensionManager.enabled before we merge this.

charts/m4d/values.yaml
shlomitk1: @roee88 I don't think it should be here at all - service url, policy name, ... Instead, the connectors should use their own configmaps.

roee88: I agree, but that's what we currently have and this PR is focused on the chart. I will need to iterate on that in follow-up PRs.

PR #393, opened 2021-02-24
Add configmap cluster-metadata to the helm chart. (#348)

revit13:


roee88:
@froesef @tomersolomon1 is it really optional? Or di we always need it even in single cluster

revit13:
@roee88 I think its not optional but I am not sure what the default values should be set to as they are configuration specific.

revit13:
I changed the config map following a talk with Roee.

charts/m4d/templates/cluster-metadata-config.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: cluster-metadata
+data:
+  {{- if .Values.clusterMetadata.clusterName }}
```
froesef, line 6: If name, region and zone are not defined the cluster managers will fail when trying to access these fields. 
The if statement suggests that it may be a property which value can be nil.

We should think about maybe setting default values for these properties. Or find a mechanism to fail helm if they are not set. (Otherwise the deployment will fail at some point and the user has to figure out where the error comes from by looking at pod logs)

roee88: `required` (see https://helm.sh/docs/howto/charts_tips_and_tricks/)

froesef: I was looking at the code and it seems like for local cluster mode it also reads the config map. It could also be changed to be hardcoded for local cluster mode but I prefer what you suggested to have a default for single cluster deployment.

charts/m4d/values.yaml
```diff
@@ -188,4 +188,16 @@ clusterScoped: true
...
+# Cluster Metadata values defined in cluster-metadata ConfigMap
+clusterMetadata:
+  # Set to the name of the cluster.
+  clusterName: hobbiton
+  # Set to cluster region.
+  region: eriador
+  # Set to cluster zone.
+  zone: theshire
+  # Set to the cluster Vault auth method path.
+  vaultAuthPath: kubernetes
```
froesef, line 202: Is this the auth path that is expected or should this be `/v1/auth/kubernetes/login` ?

revit13: @froesef you are right this field is passed as is to the modules values.yaml. But I prefer to leave it as is in cluster-metedata.yaml file and convert it to v1/auth/<auth path>/login in moduleinstance.go. I'll add a patch to fix that.

froesef (approved):

roee88 (approved):
@froesef can we have a chat on name,region,zone in the razee context? I want to try eliminating cluster matadata both for local and multi-cloud if possible.

PR #394, opened 2021-02-24
Simpler Egeria install and support for Egeria 2.6

rohithdv:
This PR deals with changes related to support for Egeria 2.6 
(1) Simpler installation and configuration of Egeria  (based on changes done in https://github.com/odpi/egeria/issues/4454) 
(2) Egeria Connector related changes to support Egeria 2.6. Some of these new changes in Egeria 2.6 include the following: 
       (A) output of rest APIs like the  get asset rest API has changed. There are now some new fields added and some older fields deleted.
      (B) Also the format of some rest APIs have changed.
      (C) The default user info has also changed. 

To install egeria, go to `third_party/egeria/ `and execute `make deploy`

Port forward: 
kubectl port-forward -n <namespace> svc/egeria-platform 9443:9443
Access Swagger UI : https://localhost:9443/swagger-ui/index.html?configUrl=/v3/api-docs/swagger-config

kubectl port-forward -n <namespace> svc/egeria-presentation 8091:8091
Access Egeria UI : https://localhost:8091/org/ (username: garygeeke, password: admin) 

To install asset in egeria, do as in the following example: 
        cd `thirdy_party/egeria/usage`
	`./create_new_asset.sh example_transactions.csv.json "finance”


connectors/egeria/src/main/java/com/ibm/egeria/EgeriaClient.java
```diff
@@ -55,7 +55,7 @@
     private String assetGuid;
     // at this point we make it a constant, this will change and should be passed as
     // credentials for Egeria
-    private final String userid = "calliequartile";
+    private final String userid = "garygeeke";
```
shlomitk1, line 58: I think the username should not be hard-coded but rather be passed through env variable (a secret attached to the deployment)

third_party/egeria/usage/read_asset.sh
```diff
@@ -8,22 +8,23 @@ echo "Read asset info"
 
 echo "ASSET_ID: $GUID"
 
-USER=calliequartile
+USER=garygeeke
```
shlomitk1, line 11: Same comment about hard-coded username. Should not be in the git repo.

connectors/egeria/src/main/resources/json/asset/asset_metadata.json
```diff
@@ -1,82 +1,141 @@
...
+            ],
+            "elementTypeDescription": "A schema type for a table oriented data structure.",
+            "elementTypeId": "248975ec-8019-4b8a-9caf-084c8b724233",
+            "elementTypeName": "TabularSchemaType",
+            "elementTypeVersion": 1,
+            "elementVersion": 1,
+            "headerVersion": 0,
+            "status": "ACTIVE"
+        }
+    }
```
shlomitk1, line 140: do we really need all these properties?

connectors/egeria/src/main/java/com/ibm/egeria/EgeriaClient.java
rohithdv: fixed. Now we have an env variable "EGERIA_DEFAULT_USERNAME" which is set in in egr_connector.yaml. Thanks.

third_party/egeria/usage/read_asset.sh
rohithdv: This file is not used during deployment. It is mainly for initialisation purposes on local machine. So it is not possible to read this value from deployment based env variable in my view. Please suggest if there are any other ways to initialise this variable. Thanks.

connectors/egeria/src/main/resources/json/asset/asset_metadata.json
rohithdv: This json is a superset of all possible fields that can be returned by Egeria during the Get Asset API. This is used mainly to load the relevant classes at compile time. At runtime any subset of these fields can be returned by Egeria and we can  read only the relevant fields. But if we reduce the fields in this file and they are returned by Egeria at runtime, then we can end up in a run time error. That is the purpose of having all the possible fields in this json. It sets up some sort of template response for the REST API response from Egeria.

shlomitk1 (approved):

third_party/egeria/deploy.sh
```diff
@@ -4,24 +4,22 @@ set -e
 set -x
 
 : ${WITHOUT_OPENSHIFT=false}
+: ${DONOT_DELETE_PERSISTENT_STORAGE=false}
 
-# CHART=odpi-egeria-lab
-# RELEASE=lab
-NAMESPACE=egeria-catalog2
+NAMESPACE=irltest1
```
ronenkat, line 9: what are we using "irltest1" for? if this is the target namespace where we install Egeria it should not be named IRLTEST1

rohithdv: Sorry this was an overlooking error. Corrected it. Please check. Thanks.

ronenkat (approved):

PR #396, opened 2021-02-26
Add GitHub Actions as a CI

### LINKED ISSUE #386: Move CI to GitHub Actions

froesef:
Fixes #386


.github/workflows/build.yml
```diff
@@ -0,0 +1,180 @@
...
+      uses: actions/download-artifact@v2
+      with:
+        name: images
+    - name: Load images
+      if: ${{ github.event_name != 'pull_request' && env.ACTIVE_CI == 'gh_actions' }}
+      run: docker load -i images.tar
+    - run: docker images
+    - name: Publish images
+      if: ${{ github.event_name != 'pull_request' && env.ACTIVE_CI == 'gh_actions' }}
+      run: sudo chmod o+x /etc/docker && make docker-retag-and-push-public && make helm-push-public
```
roee88, line 131: [FUTURE] 
I think that this eventually be a separate workflow that runs on master and tags? Like build tag and push (leverage layer caching for performance), all in a workflow that runs in push to master or a `v*` tag.

roee88, line 131: helm-push-public relies on TRAVIS_TAG environment variable

roee88, line 131: No need to run `sudo chmod o+x /etc/docker` in gh-actions as far as I know

.github/workflows/build.yml
```diff
@@ -0,0 +1,180 @@
...
+      uses: actions/download-artifact@v2
+      with:
+        name: website
+    - name: Publish website
+      if: ${{ github.event_name != 'pull_request' && env.ACTIVE_CI == 'gh_actions' }}
+      uses: peaceiris/actions-gh-pages@v3
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        publish_dir: website/${{ env.WEBSITE_LOCAL_DIR }}
+        publish_branch: gh-pages
```
roee88, line 180: [FUTURE]

I'll make a PR to move this to separate workflows with the new website.

Makefile
```diff
@@ -147,6 +147,14 @@ ifneq (${TRAVIS_TAG},)
...
 endif
 
+.PHONY: save-images
+save-images:
+	docker save -o images.tar ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/manager:${DOCKER_TAGNAME} \
+		${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/secret-provider:${DOCKER_TAGNAME} \
+		${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/egr-connector:${DOCKER_TAGNAME} \
+		${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/dummy-mover:${DOCKER_TAGNAME} \
+		${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/opa-connector:${DOCKER_TAGNAME} \
+		${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/vault-connector:${DOCKER_TAGNAME}
```
roee88, line 157: [NEAR FUTURE] 

We will need to add the katalog connector to the CI

roee88 (commented):
I marked with FUTURE comments for future work, but the other comments apply to this PR.

.github/workflows/build.yml
```diff
@@ -0,0 +1,180 @@
...
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.m2/repository
+          ~/go/pkg/mod
+          ${{ github.workspace }}/hack/tools/bin
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: ${{ runner.os }}-go
+    - name: Verify
+      run: make verify
```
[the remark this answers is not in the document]
froesef: I agree. Do you think git hooks would be an option for e.g. `go fmt` ?

.github/workflows/build.yml
froesef: Github actions has `github.ref` but it works a bit different as it has the branch name or tag name. So we will have to redo this part. I'd propose to tackle this in a separate PR. There are also actions for publishing images. Maybe one of them already does something that we want to happen on a release vs main branch build.

froesef: Yes. Separate workflows might be better here. I wanted to keep one for now to avoid duplication for now. If we find some templating mechanism that makes sure that both workflows always run the tests the same way I think we should split them.

.github/workflows/build.yml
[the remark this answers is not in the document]
roee88: A commit hook? You would need to trust people to set it up after cloning the repo. For CI we can check if go fmt did anything and fail otherwise (like what https://github.com/grandcolline/golang-github-actions does). As a developer I just set my IDE to do.

.github/workflows/build.yml
roee88: We can do a quick and dirty workaround to set TRAVIS_TAG like what's described in https://github.community/t/how-to-get-just-the-tag-name/16241/10 without doing major refactoring just to get this working because we want to soon create a tagged release of the project (I think the middle of this month is a target but unsure). Unless you think we can solve it in a separate PR in time.

.github/workflows/build.yml
```diff
@@ -0,0 +1,182 @@
...
+      - github_actions_tests # Added for testing CI. TODO remove before merging
+  pull_request:
+    branches:
+      - master
+
+env:
+  DOCKER_HOSTNAME: ghcr.io
+  DOCKER_USERNAME: the-mesh-for-data
+  ACTIVE_CI: travisci # This parameter defines which CI should push to the docker repository
+  GO_VERSION: 1.13
```
roee88, line 16: You can probably add java-version too

roee88 (approved):

Issue #275, opened 2021-01-14
When a blueprint is updated, no clean-up is done


shlomitk1:
Consider the following use-case.
The user has selected a dataset which requires an implicit copy. A blueprint has been constructed including read and copy.
The user is modifying a dataset, and it no longer requires a copy in the modified blueprint. However, the previously installed copy chart is still there. Since today all blueprints are running in the same namespace, no clean-up will ever be done (before this change, deleting the blueprint resulted in the namespace deletion).

roee88:
Even with the multiple namespaces cleanup was needed (when a reconcile changes the blueprint to not include a module that was previously required).

PR #409, opened 2021-03-08
blueprint cleanup

shlomitk1:

Fixes https://github.com/IBM/the-mesh-for-data/issues/275 by keeping the installed releases in the status and removing the releases that are not required for the reconciled blueprint.


manager/controllers/app/blueprint_controller_test.go
```diff
@@ -78,10 +79,25 @@ var _ = Describe("Blueprint Controller", func() {
...
+			By("Expecting to reconcile successfully")
+			Eventually(func() int {
+				f := &app.Blueprint{}
+				if err := k8sClient.Get(context.Background(), key, f); err != nil {
+					return 0
+				}
+				if f.Status.Releases == nil {
+					return 0
+				}
+				return len(f.Status.Releases)
```
froesef, line 98: You're checking here that one release was removed. Could you extend the check that it tests that the correct release has been removed?

shlomitk1: Extended the check to ensure that the correct release has been left, and the version matches the observed generation.

froesef (approved):

PR #412, opened 2021-03-08
DRAFT: Initial Taxonomy Implementation

simanadler:
Initial work on taxonomy and OPA request's use of it.
1) Created taxonomies
2) Created an OPA request structure based on the taxonomy
3) Created an example OPA request

Note: None of this is used in the code yet.  This is just a first step to be used as part of the design discussion.


pkg/taxonomy/data_catalog_schema.json
```diff
@@ -0,0 +1,82 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "$id": "https://github.com/IBM/the-mesh-for-data/pkg/taxonomy/data_catalog_schema.json",
+    "title": "Data Catalog Taxonomy",
+    "description": "Portion of the taxonomy owned by the data catalog.",
+    "definitions": {
+        "data_sensitivity": {
+            "type": "string",
+            "description": "Level of sensitivity of the data.",
+            "enum": ["confidential", "personal", "special personal"]
```
ronenkat, line 10: should this be set at the M4D system level? or should this be user/deployment specific?
can we "configure" this in the "install"/"customize" phase?
I would look at this list as an example enum

pkg/taxonomy/data_catalog_schema.json
```diff
@@ -0,0 +1,82 @@
...
+            "enum": ["union", "country", "state", "city", "region"]
+        },
+        "geography": {
+            "type": "string",
+            "description": "Name of the geography"
+        },
+        "data_format": {
+            "type": "string",
+            "description": "Format in which the data is stored.",
+            "enum": ["parquet", "csv", "arrow", "json", "table", "avro", "binary"]
```
ronenkat, line 24: I don't think data_format should be a strict enum...
adding a new data format should not impact the taxonomy

pkg/taxonomy/module_schema.json
```diff
@@ -0,0 +1,72 @@
...
+                },
+                "name": {
+                    "type": "string"
+                }
+            }
+        },
+        "protocol": {
+            "type": "string",
+            "description": "Protocol via which data is accessed.",
+            "enum": ["s3", "kafka", "jdbc-db2", "m4d-arrow-flight"]
```
ronenkat, line 34: I don't think protocol should be an enum.
The list of supported protocols by M4D could be build at run-time by collecting the list of supported protocol from the installed modules, and that list can generate checkers that M4D would use to validate M4DApplication input.

roee88:
Cool, but I have too many comments for reviewing it here. This includes general guidelines on how to write the spec for making it directly usable in our built-in catalog and the scope of what's part of a taxonomy and what's not.

As you said it's for the discussion so I suggest to setup a meeting to discuss like we did so far before taking this here and then have concrete smaller scoped PRs. Would that work for you?

Ronen, things like protocol must be part of the taxonomy, even if you eventually auto-generate it somehow. The taxonomy specification is a (partially) mutable piece that is the lower level mechanism for coordinating between all components.

pkg/taxonomy/data_catalog_schema.json
shlomitk1: If it is not a strict enum, we should remove the enforced validation from the controller code.

simanadler:
> A great step toward organizing our terms and taxonomy.
> some feedback and thoughts.
> The PR is not self contained and does not reference background material. I think more context is needed, even if one looks at #238.

@ronenkat Did you see the design document? https://docs.google.com/document/d/1sfdcNI6PTIYWsSIznpNqJNQCdVaVFchYB2vljzUrlew/edit#heading=h.img5h5xmx8ok

> I wonder what we should leave to be defined at the install or customization phase, and what should we include as "system"/default.

As described in the design document, different parts of the taxonomy are "owned" by different systems/components.  That's the reason that I separated the taxonomy into 3 different files: policy manager, data catalog, modules.  M4D may support different instances of all of these.  The systems and modules are define when M4D is deployed, so the taxonomy should be deployed together with them (future - pulling automatically the taxonomy values from the relevant "owning" system.). Whether the taxonomy needs to be updated after deployment ... there certainly could be situations where that would be necessary - ex: if there are changes (additions/deletions) to the values in the owning system.

> for enums that we put as samples, we should clearly indicate that it sample values, and the user is expected to override it in a deployment --- we will need to allow him the override process.

The enum values need to come from the "owning" system.

simanadler:
> Cool, but I have too many comments for reviewing it here. This includes general guidelines on how to write the spec for making it directly usable in our built-in catalog and the scope of what's part of a taxonomy and what's not.
> 
> As you said it's for the discussion so I suggest to setup a meeting to discuss like we did so far before taking this here and then have concrete smaller scoped PRs. Would that work for you?

@roee88 Sure!
> 
> Ronen, things like protocol must be part of the taxonomy, even if you eventually auto-generate it somehow. The taxonomy specification is a (partially) mutable piece that is the lower level mechanism for coordinating between all components.

Exactly

pkg/taxonomy/module_schema.json
simanadler: See my comment below about owning systems

pkg/taxonomy/data_catalog_schema.json
simanadler: That's exactly what the taxonomy is for.

pkg/taxonomy/data_catalog_schema.json
simanadler: See comment below

Issue #414, opened 2021-03-11
Pluging setup: Add to vault helm installation different values.yaml files for the Vault secret plugins


revit13:
This issue is part of https://github.com/IBM/the-mesh-for-data/issues/406

**Background**:
To retrieve the dataset credentials from where they are stored (catalog/ K8S secrets) Vault secrets plugins are used.
For  example, when dataset credentials are stored in kubernetes secrets the plugin https://github.com/mesh-for-data/vault-plugin-secrets-kubernetes-reader can be used.

**Issue Description**:
This issue extends Vault helm installation to use a values.yaml file with the appropriate Vault secret plugin details for the mesh for data configuration . For example, when the mesh for data configuration stores dataset credentials in kubernetes secrets the values.yaml will include the required values to install https://github.com/mesh-for-data/vault-plugin-secrets-kubernetes-reader plugin in the helm installation.

PR #417, opened 2021-03-11
Add to vault helm installation different values.yaml files

revit13:
This PR addresses issue #414 .

To retrieve the dataset credentials from where they are stored Vault secrets plugins are used.
This PR uses kubernetes secrets plugin: https://github.com/mesh-for-data/vault-plugin-secrets-kubernetes-reader for the default Vault installation by extending Vault helm installation to use a values.yaml file with the appropriate Vault secret plugin details.


charts/Makefile
```diff
@@ -17,14 +17,17 @@ cert-manager:
...
-vault:
-	@echo "Installing vault ..."
-	helm repo add hashicorp https://helm.releases.hashicorp.com
-	helm install vault hashicorp/vault --version 0.8.0 \
-                --set "global.openshift=false" \
-                --set "server.dev.enabled=true" \
-		--wait --timeout 120s
+.PHONY: install-vault
+install-vault:
+	make -C vault install
```
roee88, line 22: I see that in the root makefile we use $(MAKE)

charts/Makefile
```diff
@@ -17,14 +17,17 @@ cert-manager:
 		--namespace cert-manager \
 		--wait --timeout 120s
 
-.PHONY: vault
```
roee88, line 20: This target is in use, replace its usage too.

charts/vault/Makefile
```diff
@@ -0,0 +1,32 @@
+SHELL := /bin/bash # Use bash syntax
```
roee88, line 1: What command requires it?

charts/vault/Makefile
```diff
@@ -0,0 +1,32 @@
...
+ROOT_DIR:=../..
+include $(ROOT_DIR)/Makefile.env
+include $(ROOT_DIR)/hack/make-rules/docker.mk
+
+KUBE_NAMESPACE ?= m4d-system
+WITHOUT_OPENSHIFT ?= true
```
roee88, line 7: Consider changing to proactive OPENSHIFT flag defaulted to false and just always set its value to .global.openshift

charts/vault/Makefile
```diff
@@ -0,0 +1,32 @@
...
+install:
+	kubectl create namespace $(KUBE_NAMESPACE) 2>/dev/null || true
+	helm repo add hashicorp https://helm.releases.hashicorp.com
+        ifeq ($(WITHOUT_OPENSHIFT), true)
+	$(info installing Vault without openshift)
+	helm install vault hashicorp/vault --values values/plugin-secrets-kubernetes-reader.yaml
+        else
+	helm install vault hashicorp/vault --values values/plugin-secrets-kubernetes-reader.yaml --values values/openshift.yaml
+        endif
+	kubectl create secret generic vault-unseal-keys --from-literal=vault-root=root -n $(KUBE_NAMESPACE) || true
```
roee88, line 19: What's this secret?

charts/vault/Makefile
```diff
@@ -0,0 +1,32 @@
...
+
+.PHONY: install
+install:
+	kubectl create namespace $(KUBE_NAMESPACE) 2>/dev/null || true
+	helm repo add hashicorp https://helm.releases.hashicorp.com
+        ifeq ($(WITHOUT_OPENSHIFT), true)
+	$(info installing Vault without openshift)
+	helm install vault hashicorp/vault --values values/plugin-secrets-kubernetes-reader.yaml
+        else
+	helm install vault hashicorp/vault --values values/plugin-secrets-kubernetes-reader.yaml --values values/openshift.yaml
```
roee88, line 17: Not setting dev mode? I don't think that the intention of this makefile is for production deployment

charts/vault/Makefile
```diff
@@ -0,0 +1,32 @@
...
+.PHONY: uninstall
+uninstall:
+	kubectl delete secret vault-unseal-keys -n $(KUBE_NAMESPACE)
+	kubectl delete -f clusterrole.yaml -n $(KUBE_NAMESPACE)
+	kubectl delete -f clusterrolebinding.yaml -n $(KUBE_NAMESPACE) || true
+	helm uninstall vault -n $(KUBE_NAMESPACE)
+
+.PHONY: wait-for-vault
+wait-for-vault:
+	@while [[ -z `kubectl get pods vault-0 -o json | jq 'select(.status.phase=="Running") | true'` ]]; do echo "Waiting for vault pod" ; sleep 2; done
```
roee88, line 32: Why not install with --wait and timeout instead?

charts/vault/README.md
```diff
@@ -0,0 +1,11 @@
+
+## Vault Helm Installation
+
+The dataset credentials are retreived from where they are stored (catalog/kubernetes secrets) using [Vault plugins](https://www.vaultproject.io/docs/internals/plugins).
+
+An example for such plugin is [vault-plugin-secrets-kubernetes-reader](https://github.com/mesh-for-data/vault-plugin-secrets-kubernetes-reader) which reads dataset credentials that are stored in kubernetes secrets.
+
+The Vault Helm Installation can support plugins for different configurations depending on the values provided to the chart.
+
+The values/ directory contains yaml files for the different plugins available.
```
roee88, line 10: Is it possible to install multiple plugins together using what you have here?

charts/vault/values/plugin-secrets-kubernetes-reader.yaml
```diff
@@ -0,0 +1,34 @@
+server:
```
roee88, line 1: Not related to plugin, like other fields here. 

Consider making this strictly for the things required to setup the plugin so it can be merged later with production deployment values. This is the file in this PR that can be useful beyond just our testing.

charts/vault/values/plugin-secrets-kubernetes-reader.yaml
```diff
@@ -0,0 +1,34 @@
...
+  dev:
+    enabled: true
+  extraArgs: '-dev-root-token-id=root -dev-plugin-dir=/usr/local/libexec/vault'
```
roee88, line 4: What does it do?

charts/vault/values/plugin-secrets-kubernetes-reader.yaml
```diff
@@ -0,0 +1,34 @@
...
+    - name: plugins
+      emptyDir: {}
+  volumeMounts:
+    - mountPath: /usr/local/libexec/vault
+      name: plugins
+      readOnly: false
+  extraInitContainers:
+    - name: kubesecrets-plugin
+      image: ghcr.io/the-mesh-for-data/vault-plugin-secrets-kubernetes-reader:latest
+      imagePullPolicy: Always
```
roee88, line 15: If this is omitted then it's also omitted from the deployment? I think that would be a better default.

roee88 (changes requested):
Folders under charts should be Helm charts. This isn't a helm chart.

charts/vault/Makefile
revit13: The while in the wait-for-vault command. I will remove it based on your next suggestion.

charts/vault/Makefile
revit13: Sure will do that. tnx

charts/vault/Makefile
revit13: its in the values.yaml file. I would set it outside but there are another flags that are related to dev mode: -dev-plugin-dir and VAULT_DEV_LISTEN_ADDRESS.

charts/vault/Makefile
revit13: Its the vault token. its been used in the project to authenticate to vault (in the manager for example)... the installation in third_party/vault seems to create it automatically, the following is from the vault-0 pod created with banzaicloud repo:
  - args:
    - --pre-flight-checks
    - "true"
    - --mode
    - k8s
    - --k8s-secret-namespace
    - m4d-system
    - --k8s-secret-name
    - vault-unseal-keys
    - --k8s-secret-labels
    - app.kubernetes.io/name=vault,vault_cr=vault

charts/vault/Makefile
revit13: Will do that. tnx

charts/vault/values/plugin-secrets-kubernetes-reader.yaml
revit13: dev-root-token-id sets the vault token to root and dev-plugin-dir registers all plugin under the plugin directory when running in dev mode. in non dev mode there are other steps to register the plugin.

charts/Makefile
revit13: OK will change that. tnx

charts/vault/values/plugin-secrets-kubernetes-reader.yaml
revit13: Please explain not sure I understand... tnx

charts/vault/Makefile
roee88: Okay, let's talk about that for a follow-up, doesn't seem right. Unsealing is a separate process (not required in dev mode btw). Login should use a proper method even from the manager.

charts/vault/values/plugin-secrets-kubernetes-reader.yaml
roee88: Can you please share a link to the vault docs?

charts/vault/values/plugin-secrets-kubernetes-reader.yaml
roee88: Suggestion is to omit `imagePullPolicy: Always` and keep default behavior (IfNotExists unless tag is latest)

charts/vault/values/plugin-secrets-kubernetes-reader.yaml
revit13: Sure: https://www.vaultproject.io/docs/commands/server#dev-options

revit13:
@roee88 Does third_party/vault directory is a better place to put this content (replacing the current Vault installation under this directory)? Thanks

charts/vault/Makefile
roee88: VAULT_DEV_LISTEN_ADDRESS is a workaround for a bug in 0.9.0.
No longer needed in 0.9.1. See [changelog](https://github.com/hashicorp/vault-helm/blob/master/CHANGELOG.md#091-february-2nd-2021).

roee88: Did you consider using proper plugin dir configuration (not for dev)?

roee88:
> @roee88 Does third_party/vault directory is a better place to put this content (replacing the current Vault installation under this directory)? Thanks

Yes. The Makefile in charts/ would eventually be part of root Makefile too so the flow to have in mind is from root setting up cert-manager, vault, m4d-crd and m4d.

ronenkat (changes requested):
PR is missing description and has no associated issue.
Please explain why this PR is needed, and what is the expected behaviour after it.

revit13:
@ronenkat Sorry the issue number was in the commit message. It contains the description of the problem (#414).

charts/vault/Makefile
revit13: I will try to use that... tnx

revit13: @roee88 I checked that and in order to configure the plugin that way, Vault server should not be in dev mode as it needs the plugin_directory to be in a configuration file, please see:  https://www.vaultproject.io/docs/configuration and 
https://www.vaultproject.io/docs/configuration#plugin_directory
Should I switch to non dev mode? Thanks

charts/vault/README.md
revit13: When using a single yaml for several plugins that works OK. if I use separate yaml files for each plugin and the use --values for each of them I could not install them properly. I might be doing something wrong so I continue to investigate.

revit13:
Something went wrong while resolving conflicts... I'll close this PR and open a new one.

PR #423, opened 2021-03-14
read-module endpoint

tomersolomon1:
Signed-off-by: tomersolomon1 <tomer.solomon1@gmail.com>

Implementing the following design - https://docs.google.com/document/d/10qL_sUBqbSmPww1GQ76ON8KjQjFVByU3aXarmCx9uHA/edit

Which is based on this discussion: https://github.com/IBM/the-mesh-for-data/discussions/266

This PR adds the URLs of the read-module endpoints to the status of the m4d application.


tomersolomon1:
I also want to do some refactoring - I noticed we used both the terms `AssetID` and `DataSetID`. I find it confusing, and think it's better to use only one term. Is there a reason for using these two terms?

manager/controllers/app/blueprint_controller.go
```diff
@@ -140,7 +135,7 @@ func (r *BlueprintReconciler) deleteExternalResources(blueprint *app.Blueprint)
 
 func (r *BlueprintReconciler) hasExternalResources(blueprint *app.Blueprint) bool {
 	for _, step := range blueprint.Spec.Flow.Steps {
-		releaseName := getReleaseName(blueprint.Name, step)
+		releaseName := utils.GetReleaseName(blueprint.Labels[app.ApplicationNameLabel], blueprint.Labels[app.ApplicationNameLabel], step)
```
shlomitk1, line 138: Duplicates - one should be Namespace

manager/controllers/app/blueprint_controller.go
```diff
@@ -116,15 +116,10 @@ func (r *BlueprintReconciler) reconcileFinalizers(blueprint *app.Blueprint) (ctr
...
-func getReleaseName(blueprintName string, step app.FlowStep) string {
-	fullName := blueprintName + "-" + step.Name
-	return utils.HelmConformName(fullName)
-}
-
 func (r *BlueprintReconciler) deleteExternalResources(blueprint *app.Blueprint) error {
 	errs := make([]string, 0)
 	for _, step := range blueprint.Spec.Flow.Steps {
-		releaseName := getReleaseName(blueprint.Name, step)
+		releaseName := utils.GetReleaseName(blueprint.Labels[app.ApplicationNameLabel], blueprint.Labels[app.ApplicationNameLabel], step)
```
shlomitk1, line 122: Duplicate Name label

manager/controllers/app/blueprint_controller.go
```diff
@@ -257,8 +252,8 @@ func (r *BlueprintReconciler) reconcile(ctx context.Context, log logr.Logger, bl
 		if err != nil {
 			return ctrl.Result{}, errors.WithMessage(err, "Blueprint step arguments are invalid")
 		}
-
-		releaseName := getReleaseName(blueprint.Name, step)
+		log.Info("Step name: " + step.Name)
+		releaseName := utils.GetReleaseName(blueprint.Labels[app.ApplicationNameLabel], blueprint.Labels[app.ApplicationNameLabel], step)
```
shlomitk1, line 256: See previous comment

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -232,6 +232,34 @@ func (r *M4DApplicationReconciler) deleteExternalResources(applicationContext *a
...
+				for _, arg := range step.Arguments.Read {
+					applicationContext.Status.ReadEndpointsMap[arg.AssetID] = app.EndpointSpec{
+						Hostname: fqdn,
+						Port:     originalEndpointSpec.Port,
+						Scheme:   originalEndpointSpec.Scheme,
+					}
+				}
+			}
+		}
+		// We found a blueprint with a read modules, we can go home
```
shlomitk1, line 256: Typo in a comment, the second part may be omitted.

manager/apis/app/v1alpha1/m4dapplication_types.go
```diff
@@ -42,10 +42,8 @@ type DataRequirements struct {
 	Copy CopyRequirements `json:"copy,omitempty"`
 }
 
-// DataContext indicates data set chosen by the Data Scientist to be used by his application,
-// and includes information about the data format and technologies used by the application
-// to access the data.
-type DataContext struct {
+// AssetDetails provides all details required to fully identify an asset - the datasetID and the catalog details
+type AssetDetails struct {
```
shlomitk1, line 46: Catalog service does not identify the asset, merely suggests what service to refer to for the lookup. 
I prefer you leave the current spec because the catalog service is not used yet, and adding an hierarchy only complicates the spec. There is an additional cost to changing the spec: in all samples we have, documentation, plus changing the code of GUI backend/frontend s.t. the GUI functionality is preserved.

charts/m4d-crd/templates/app.m4d.ibm.com_m4dapplications.yaml
```diff
@@ -218,7 +218,24 @@ spec:
...
                     type: string
                 type: object
-              description: ProvisionedStorage maps a dataset (identified by DataSetID) to the new provisioned bucket. It allows M4DApplication controller to manage buckets in case the spec has been modified, an error has occurred, or a delete event has been received. ProvisionedStorage has the information required to register the dataset once the owned plotter resource is ready
+              description: ProvisionedStorage maps a dataset (identified by AssetID) to the new provisioned bucket. It allows M4DApplication controller to manage buckets in case the spec has been modified, an error has occurred, or a delete event has been received. ProvisionedStorage has the information required to register the dataset once the owned plotter resource is ready
+              type: object
+            readEndpointsMap:
+              additionalProperties:
+                description: The spec of the endpoint exposed by the module
+                properties:
+                  Scheme:
```
ronenkat, line 227: is this schema or protocol?

charts/m4d-crd/templates/app.m4d.ibm.com_m4dapplications.yaml
```diff
@@ -218,7 +218,24 @@ spec:
                     description: Reference to a secret where the credentials are stored
                     type: string
                 type: object
-              description: ProvisionedStorage maps a dataset (identified by DataSetID) to the new provisioned bucket. It allows M4DApplication controller to manage buckets in case the spec has been modified, an error has occurred, or a delete event has been received. ProvisionedStorage has the information required to register the dataset once the owned plotter resource is ready
+              description: ProvisionedStorage maps a dataset (identified by AssetID) to the new provisioned bucket. It allows M4DApplication controller to manage buckets in case the spec has been modified, an error has occurred, or a delete event has been received. ProvisionedStorage has the information required to register the dataset once the owned plotter resource is ready
+              type: object
+            readEndpointsMap:
```
ronenkat, line 223: how do you tie an endpoint with an asset?
Will you be removing "dataAccessInstructions:" ?

charts/m4d-crd/templates/app.m4d.ibm.com_m4dapplications.yaml
tomersolomon1: URI scheme, see https://en.wikipedia.org/wiki/List_of_URI_schemes

charts/m4d-crd/templates/app.m4d.ibm.com_m4dapplications.yaml
tomersolomon1: I'm not sure I understand your question. Anyhow, I prefer to discuss the go structs and not the yamls which are a bit verbose.

shlomitk1 (approved):

manager/apis/app/v1alpha1/m4dmodule_types.go
```diff
@@ -87,6 +87,25 @@ type SupportedAction struct {
 	Level connectors.EnforcementAction_EnforcementActionLevel `json:"level,omitempty"`
 }
 
+// The spec of the endpoint exposed by the module
+type EndpointSpec struct {
+	// +required
+	Hostname string `json:"endpointName,omitempty"`
+	// +required
+	Port int32 `json:"port" protobuf:"varint,3,opt,name=port"`
```
roee88, line 95: Remove protobuf stuff

manager/apis/app/v1alpha1/m4dmodule_types.go
```diff
@@ -87,6 +87,25 @@ type SupportedAction struct {
 	Level connectors.EnforcementAction_EnforcementActionLevel `json:"level,omitempty"`
 }
 
+// The spec of the endpoint exposed by the module
+type EndpointSpec struct {
+	// +required
+	Hostname string `json:"endpointName,omitempty"`
```
roee88, line 93: style: match json tags to variable name (but start with lowercase).
That is: endpointName -> hostname

manager/apis/app/v1alpha1/m4dmodule_types.go
```diff
@@ -87,6 +87,25 @@ type SupportedAction struct {
...
+	// +required
+	Port int32 `json:"port" protobuf:"varint,3,opt,name=port"`
+
+	// For example: http, https, grpc, grpc+tls, jdbc:oracle:thin:@ etc
+	// +required
+	Scheme string `json:"Scheme,omitempty"`
```
roee88, line 99: Scheme->scheme

roee88, line 99: remove omitempty from required field(s)

manager/apis/app/v1alpha1/m4dmodule_types.go
```diff
@@ -87,6 +87,25 @@ type SupportedAction struct {
...
+}
+
+type ModuleAPI struct {
+	// +required
+	InterfaceDetails `json:",inline"`
+	// +optional
+	Endpoint EndpointSpec `json:"endpointSpec,omitempty"`
```
roee88, line 106: endpointSpec -> endpoint (in json tag)

pkg/multicluster/local/local_cluster_manager.go
```diff
@@ -84,14 +83,8 @@ func (cm *ClusterManager) UpdateBlueprint(cluster string, blueprint *v1alpha1.Bl
...
-	resource := &v1alpha1.Blueprint{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      blueprint.Name,
-			Namespace: blueprint.Namespace,
-		},
-	}
-	if _, err := ctrl.CreateOrUpdate(context.Background(), cm.Client, resource, func() error {
-		resource.Spec = blueprint.Spec
+	// We use the identity transformation: f(x) = x
+	if _, err := ctrl.CreateOrUpdate(context.Background(), cm.Client, blueprint, func() error {
```
roee88, line 87: I believe that's not how CreateOrUpdate works. Internally it calls Get that actually modifies your blueprint.Spec to the state in the cluster. The original code was written correctly.

roee88 (changes requested):
Partial review (sorry)

manager/controllers/app/blueprint_controller.go
```diff
@@ -116,15 +116,10 @@ func (r *BlueprintReconciler) reconcileFinalizers(blueprint *app.Blueprint) (ctr
...
-func getReleaseName(blueprintName string, step app.FlowStep) string {
-	fullName := blueprintName + "-" + step.Name
-	return utils.HelmConformName(fullName)
-}
-
 func (r *BlueprintReconciler) deleteExternalResources(blueprint *app.Blueprint) error {
 	errs := make([]string, 0)
 	for _, step := range blueprint.Spec.Flow.Steps {
-		releaseName := getReleaseName(blueprint.Name, step)
+		releaseName := utils.GetReleaseName(blueprint.Labels[app.ApplicationNameLabel], blueprint.Labels[app.ApplicationNamespaceLabel], step)
```
froesef, line 122: When the application controller creates the plotter the name will be `appName + "-" + appNamespace`. The plotter will create a blueprint per cluster and will name it the same. Thus a blueprint name should already have app name and app namespace in the name. Why do we need to pass this here separately?

manager/controllers/utils/utils.go
```diff
@@ -118,6 +118,22 @@ func CreateDataSetIdentifier(datasetID string) string {
...
 }
 
+// Generating release name from step
+func GetReleaseName(applicationName string, namespace string, step app.FlowStep) string {
+	return GetReleaseNameByStepName(applicationName, namespace, step.Name)
+}
+
+// Generate release name from step name
+func GetReleaseNameByStepName(applicationName string, namespace string, stepName string) string {
+	fullName := applicationName + "-" + namespace + "-" + Hash(applicationName, 5) + "-" + Hash(namespace, 5) + "-" + stepName
```
froesef, line 128: Why do we hash `applicationName` and `namespace` here if we already have the full value of them in the name?

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -232,6 +232,35 @@ func (r *M4DApplicationReconciler) deleteExternalResources(applicationContext *a
...
+	var foundReadEndpoints = false
+	for _, blueprintSpec := range blueprintsMap {
+		for _, step := range blueprintSpec.Flow.Steps {
+			if step.Arguments.Read != nil {
+				// We found a read module
+				foundReadEndpoints = true
+				releaseName := utils.GetReleaseName(applicationContext.ObjectMeta.Name, applicationContext.ObjectMeta.Namespace, step)
+				moduleName := step.Template
+				originalEndpointSpec := moduleMap[moduleName].Spec.Capabilities.API.Endpoint
+				fqdn := utils.GenerateModuleEndpointFQDN(releaseName, originalEndpointSpec, BlueprintNamespace)
```
froesef, line 247: Where is the endpointSpec passed to the module? The way I understand the read module is that it will use the release name as a service name. And the release name contains application name + namespace + step name. I'm not sure what service will be exposed at the fqdn that you're building here.

manager/apis/app/v1alpha1/m4dmodule_types.go
froesef: Why is the host name part of the module specification? Modules are templates that are used in order to create helm installs of copies/services. Thus this is not a hostname of a service but an added string to a service name (that you're building as fqdn below)

froesef:
I might get this issue wrong. But I thought the point was (as described in the google doc 2b) to allow the user to specify an endpoint where a service should be exposed. (otherwise it's similar to what already exists with the python notebook code sample that gets generated). And also to have a stable endpoint no matter of what read module is used. I thought a service with an external name will be created for the user that automatically adjust the alias so that the user does not have to worry about changes from the module side.
So the specification of an endpoint should be in the application spec and not the module spec.

manager/apis/app/v1alpha1/m4dmodule_types.go
tomersolomon1: Since I'm intending to reuse this struct also for holding the "real" hostname (in the status), I think it doesn't matter.. If you have a better name I'm open to suggestions :)

manager/controllers/app/blueprint_controller.go
tomersolomon1: appName1 = "a", NS1 = "a-a"
appName2 = "a-a", NS2 = "a"

We have appName1-NS1 == appName2-NS2

I wanted to avoid such cases using hash. But we can decide to handle it together in the future.

manager/controllers/utils/utils.go
tomersolomon1: See above comment

pkg/multicluster/local/local_cluster_manager.go
tomersolomon1: You're right, I'll fix

manager/controllers/app/blueprint_controller.go
shlomitk1: You can't have two blueprints with the same name.

tomersolomon1: Changed it to be:
```golang
// Generate release name from step name
func GetReleaseNameByStepName(applicationName string, namespace string, stepName string) string {
	fullName := applicationName + "-" + namespace + "-" + stepName
	return HelmConformName(fullName)
}

// Generate fqdn for a module
func GenerateModuleEndpointFQDN(releaseName string, blueprintNamespace string) string {
	return releaseName + "." + blueprintNamespace + ".svc.cluster.local"
}
```

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -232,6 +232,35 @@ func (r *M4DApplicationReconciler) deleteExternalResources(applicationContext *a
 	return nil
 }
 
+// getReadModulesEndpoints populates the ReadEndpointsMap map in the status of the m4dapplication
+// Current implementation assumes there is only one cluster with read modules (which is the same cluster the user's workload)
+func (r *M4DApplicationReconciler) getReadModulesEndpoints(applicationContext *app.M4DApplication, blueprintsMap map[string]app.BlueprintSpec, moduleMap map[string]*app.M4DModule) {
```
froesef, line 237: Minor thing. The function name is misleading. It does not get the read module endpoints. It sets them. Also there is no need to bind this to `M4DApplicationReconciler`.
This could be e.g. `setReadModuleEndpoints`. And then the call below would be:
```
setReadModuleEndpoints(applicationContext, blueprintPerClusterMap, moduleMap)
```

froesef (approved):

manager/controllers/app/m4dapplication_controller.go
tomersolomon1: Fixed

pkg/multicluster/local/local_cluster_manager.go
```diff
@@ -4,11 +4,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
roee88, line 7: please move it back. Ping me for setting up your IDE to avoid this (I know, docs for that should be added but we don't have it yet).

manager/controllers/app/blueprint_mgr.go
```diff
@@ -83,7 +83,7 @@ func (r *M4DApplicationReconciler) GenerateBlueprint(instances []modules.ModuleI
 
 		// Create a flow step
 		var step app.FlowStep
-		step.Name = modulename + "-" + utils.Hash(moduleInstance.AssetID, 10) // Need unique name for each step so include ids for dataset
+		step.Name = utils.CreateStepName(modulename, moduleInstance.AssetID) // Need unique name for each step so include ids for dataset
```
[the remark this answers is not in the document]
shlomitk1: No. For read modules there can be multiple assets handled by a single step: step.Arguments.Read[i].AssetID
When multiple instances are united into a single step, moduleInstance.AssetID will represent just one of these values, but it is enough to create a unique step name. The uniqueness is more relevant for the instances that are not united, to distinguish between copyX and copyY.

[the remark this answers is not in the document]
roee88: >  , but it is enough to create a unique step name. 

Only if you only consume the asset in one way, which wasn't supposed to be a restriction but it's not that important. I do find it very confusing though.

manager/apis/app/v1alpha1/m4dmodule_types.go
```diff
@@ -74,6 +74,25 @@ type SupportedAction struct {
 	Level connectors.EnforcementAction_EnforcementActionLevel `json:"level,omitempty"`
 }
 
+// The spec of the endpoint exposed by the module
+type EndpointSpec struct {
+	// +optional
+	Hostname string `json:"hostname"`
```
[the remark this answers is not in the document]
tomersolomon1: I intended to use it, you and Florian thought it's redundant.. Anyhow this field is used currently by the status of the m4d-application. I'll add comment.

manager/controllers/app/blueprint_mgr.go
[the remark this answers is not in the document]
tomersolomon1: I'm not I understand, what's the bottom line?

pkg/multicluster/local/local_cluster_manager.go
tomersolomon1: I don't understand. Remove it?

manager/controllers/app/blueprint_mgr.go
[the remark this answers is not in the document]
roee88: No change required in this PR here

pkg/multicluster/local/local_cluster_manager.go
roee88: This is minor. You moved the import to here from line 11. We keep import order consistent to avoid unrelated changes to imports when someone edits a file.

Because you're a comitter I suggested that you configure your IDE to do that:

- In vscode gopls (enabled by default in recent versions) automatically fix import order. 
- In goland it's Ctrl+Alt+S and navigate to Tools | File Watchers, click the Add button and select goimports.

manager/apis/app/v1alpha1/m4dmodule_types.go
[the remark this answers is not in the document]
roee88: Then I guess I was wrong or misunderstood, happens. Anyway, we don't dictate that the service would be named just as the release name. It has 10 more letters to use so it can append for example "-service" to the release name in the Helm template. But I'm okay with forcing it to be just the release name until it will hit us again, you should add it to the module docs though.

roee88 (approved):
LGTM, fix the minor comments that are still left before we merge

manager/apis/app/v1alpha1/m4dmodule_types.go
```diff
@@ -74,6 +74,26 @@ type SupportedAction struct {
 	Level connectors.EnforcementAction_EnforcementActionLevel `json:"level,omitempty"`
 }
 
+// EndpointSpec is used both by the module creator and by the status of the m4dapplication
+type EndpointSpec struct {
+	// Can be left empty when declaring a module, currently this field is only used by the status of the m4dapplication
```
roee88, line 79:
```suggestion
Always equals the release name. Can be omitted.
```

manager/apis/app/v1alpha1/m4dmodule_types.go
```diff
@@ -74,6 +74,26 @@ type SupportedAction struct {
...
+	// For example: http, https, grpc, grpc+tls, jdbc:oracle:thin:@ etc
+	// +required
+	Scheme string `json:"scheme"`
+}
+
+type ModuleAPI struct {
+	// +required
+	InterfaceDetails `json:",inline"`
+	// +optional
+	Endpoint EndpointSpec `json:"endpoint,omitempty"`
```
roee88, line 94:
```suggestion
	// +required
	Endpoint EndpointSpec `json:"endpoint"`
```

roee88 (commented):
Need to add documentation to module development docs.

ronenkat (approved):

manager/controllers/app/blueprint_mgr.go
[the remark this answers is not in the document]
shlomitk1: @roee88 https://github.com/IBM/the-mesh-for-data/pull/545 concatenates all asset ids into a single one.

Issue #422, opened 2021-03-14
Dependency on oc introduced


elsalant:
hack/install.sh seems now to have a dependency on OpenShift:
oc delete securitycontextconstraints --ignore-not-found egeria-restricted
./deploy.sh: line 17: oc: command not found
make: *** [Makefile:7: deploy] Error 127

ronenkat:
@rohithdv can you look at how to exchange the use of 'oc' with 'kubectl' in https://github.com/IBM/the-mesh-for-data/blob/master/third_party/egeria/deploy.sh lines 17, 41-43 ?

cdoron:
@ronenkat : I think that this problem appeared when the WITHOUT_OPENSHIFT variable changed from "true" to "false" (line 6)

ronenkat:
@elsalant please check if the problem persist when changing line 6 to ": ${WITHOUT_OPENSHIFT=true}".
Thank you @cdoron

Mohammad-nassar10:
I also got the same error. I think you can run "WITHOUT_OPENSHIFT=false ./hack/install.sh" instead. It worked for me.

elsalant:
@cdoron Well, it gets past the problem with the oc command, but it now times out on 
+ helm install egeria egeria-base --namespace=egeria-catalog
Error: failed post-install: timed out waiting for the condition

Trying this as a dry-run gives: 
salant@fogprotect:~/the-mesh-for-data/third_party/egeria$ helm install egeria egeria-base --namespace=$NAMESPACE --dry-run
Error: failed to download "egeria-base" (hint: running `helm repo update` may help)
 Running helm repo update does not help.  Neither does undeploying egeria and redeploying
(This may not be connected with the WITHOUT_OPENSHIFT=true issue though....)

rohithdv:
> @rohithdv can you look at how to exchange the use of 'oc' with 'kubectl' in https://github.com/IBM/the-mesh-for-data/blob/master/third_party/egeria/deploy.sh lines 17, 41-43 ?

I tried to change oc with kubectl. However I think the command is specific to oc. I get the following error: 

<img width="827" alt="image" src="https://user-images.githubusercontent.com/59818576/111190758-6beea280-85dd-11eb-9507-7d6fab0dbc2c.png">

@ronenkat @roee88 @elsalant

roee88:
> @cdoron Well, it gets past the problem with the oc command, but it now times out on
> 
[...]

PR #424, opened 2021-03-15
Change WITHOUT_OPENSHIFT flag to true in egeria 

rohithdv:
This PR partially fixes PR #422 . 


shlomitk1 (approved):

third_party/egeria/deploy.sh
```diff
@@ -3,7 +3,7 @@
 set -e
 set -x
 
-: ${WITHOUT_OPENSHIFT=false}
+: ${WITHOUT_OPENSHIFT=true}
```
ronenkat, line 6: I believe it should be :=true and not =true

rohithdv: Fixed. Thanks.

roee88: If that's true then why not change the other variable too?

rohithdv: fixed for the DONOT_DELETE_PERSISTENT_STORAGE flag also. I think = or := is equivalent. But it is better to have a consistent notation.

PR #426, opened 2021-03-15
Add Makefile commands for the katalog connector deployment

revit13:
This PR adds Makefile commands for the katalog connector to enable its deployment. The katalog connector commands are invoked separately from the other connectors commands because the set of commands in the other connector needs to be revisited.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ docker-mirror-read:
 .PHONY: build
 build:
 	$(MAKE) -C pkg/policy-compiler build
+	$(MAKE) -C connectors/katalog build
 	$(MAKE) -C manager manager
 
 .PHONY: test
@@ -75,6 +76,7 @@ deploy:
 	$(MAKE) -C secret-provider deploy
 	$(MAKE) -C manager deploy
 	$(MAKE) -C connectors deploy
+	$(MAKE) -C connectors/katalog deploy
 
 .PHONY: undeploy
 undeploy:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -75,19 +76,22 @@ deploy:
 	$(MAKE) -C secret-provider deploy
 	$(MAKE) -C manager deploy
 	$(MAKE) -C connectors deploy
+	$(MAKE) -C connectors/katalog deploy
```
roee88, line 79: No such target. Exclude deployment from this PR because it will be part of the chart.

Makefile
```diff
@@ -75,19 +76,22 @@ deploy:
...
 
 .PHONY: undeploy
 undeploy:
 	$(MAKE) -C secret-provider undeploy
 	$(MAKE) -C manager undeploy
 	$(MAKE) -C manager undeploy-crd
 	$(MAKE) -C connectors undeploy
+	$(MAKE) -C connectors/katalog undeploy
```
roee88, line 87: Same

Makefile
```diff
@@ -124,7 +130,8 @@ DOCKER_PUBLIC_NAMES := \
 	egr-connector \
 	dummy-mover \
 	opa-connector \
-	vault-connector
+	vault-connector \
+	katalog-connector
```
roee88, line 134: Remove to match what you described in thr PR.

roee88:
I think it would be best to wait for https://github.com/IBM/the-mesh-for-data/pull/430 as it invalidates some of the changes here. Also, don't forget to add the image to
DOCKER_PUBLIC_NAMES and save_images (and anything else that I missed).

roee88:
Seems to be handled by other PRs

PR #429, opened 2021-03-15
Improved Helm chart

roee88:
Install with `make -C charts cert-manager vault vault-setup m4d`. 

Requires https://github.com/IBM/the-mesh-for-data/pull/426 before katalog-connector can be deployed successfully so you can deploy with egeria instead `helm uninstall m4d -n m4d-system && helm install m4d m4d --set coordinator.catalog=egeria -n m4d-system` .

# Changes

1. m4d-crd: type and version fields in Chart.yaml and an empty values.yaml as required by linters
2. webhook-configs.yaml generated to files and not to templates to make it included only conditionally
3. Grouped secondary components to their own folders: opa connector, katalog connector, egeria connector, opa server
4. Removed some deprecated/temporary fields from values.yaml (extension policy manger, some vault related fields) 
5. Refactor values.yaml:
    - Top level `cluster` for the cluster metadata
    - Top level `coordinator` to enable and configure coordinator-related parts in manager
    - Top level `worker` to enable and configure worker-related parts in manager 
    - Component specific configuration are part of the component (i.e., no more shared config)
    - Every component can be enabled/disabled but the default is an `auto` option to enable if configuration requires the component
6. manager `args` is now `overrideArgs`

# TODO (followups)

- [ ] Disable deployment of cluster scoped resources: manager-auth-proxy-*, manager-cr, manager-crb, webhook-configs
- [ ] Set version 0.1.0 in charts
- [ ] Fine grained RBAC
- [ ] A direct way to disable motion controllers / to deploy just motion controllers (currently possible using manager.enable and manager.overrideArgs). Do we want a `motion` section next to `coordinator` and `worker`?
- [ ] Vault address without "m4d-system" in values.yaml
- [ ] Remove secret provider deployment
- [ ] Rework OPA server deployment


roee88:
@revit13 I missed the fact that the credentials connector configuration wasn't set in m4d-config in the Helm chart. Was it omitted because it works without it already? Should we add it i the meantime?

revit13:
@roee88 Can you please give an example? Thanks

roee88:
> @roee88 Can you please give an example? Thanks

https://github.com/IBM/the-mesh-for-data/blob/947c852b869c553b25cf756d89bc71db7913a80f/manager/config/prod/deployment_configmap.yaml#L15

revit13:
@roee88 Thanks, yes please add it in the meantime

charts/m4d/templates/egeria-connector/egeria-connector-deployment.yaml
```diff
@@ -0,0 +1,63 @@
...
+      serviceAccountName: {{ .Values.egeriaConnector.serviceAccount.name }}
+      securityContext:
+        {{- toYaml .Values.egeriaConnector.podSecurityContext | nindent 8 }}
+      containers:
+        - name: egeria-connector
+          securityContext:
+            {{- toYaml .Values.egeriaConnector.securityContext | nindent 12 }}
+          image: {{ include "m4d.image" ( tuple $ .Values.egeriaConnector ) }}
+          imagePullPolicy: {{ .Values.egeriaConnector.imagePullPolicy | default .Values.global.imagePullPolicy }}
+          ports:
```
froesef, line 42: Are there some default liveness or readiness probes for egeria?

charts/m4d/templates/webhook-configs.yaml
```diff
@@ -1,89 +1,3 @@
...
-      - apiGroups:
-          - motion.m4d.ibm.com
-        apiVersions:
-          - v1alpha1
-        operations:
-          - CREATE
-          - UPDATE
-        resources:
-          - streamtransfers
+{{- if include "m4d.isEnabled" (tuple .Values.manager.enabled (or .Values.coordinator.enabled .Values.worker.enabled)) }}
```
froesef, line 1: Currently the webhooks are only used in the motion apis. Thus only in the worker. But if we're planning to add webhooks for the other APIs as well we should leave this as is. (And I think we can find some useful checks for M4dApplication as well)

charts/m4d/values.yaml
```diff
@@ -12,19 +24,110 @@ global:
...
+  # Set to false to disable coordinator components in manager.
+  enabled: true
+
+  # Configures the catalog system name to be used by the coordinator manager.
+  # Accepted values are "katalog", "egeria" or any meaningful name if a third party connector is used.
+  catalog: "katalog"
+
+  # Overrides the catalog connector URL.
+  # Defaults to `<catalog>-connector`.
+  catalogConnectorURL: ""
```
froesef, line 55: How does helm 3 deal with defaults? Is an empty string treated the same as nil or no definition? Thus when defining `""` here is the default executed or is the empty string written in as value?

roee88: generally they are treated the same. When you use `default` in the templates any of these values are replaced with the default.

charts/m4d/templates/webhook-configs.yaml
roee88: Yes I wondered whether we want these always enabled to avoid changes here in the future. What would be your preference?

charts/m4d/templates/egeria-connector/egeria-connector-deployment.yaml
roee88: This is the connector, not egeria itself. But you're right we can probably add it. I'll check if this can be done easily for GRPC like it is for HTTP.

roee88: OK it's not that easy, the examples I found require a grpc_health_probe binary.

charts/m4d/templates/webhook-configs.yaml
froesef: I would leave them enabled for all.

froesef (approved):

PR #430, opened 2021-03-15
Refactor docker image build

froesef:
I'm proposing to refactor the `docker-build` and `docker-all` commands. As many images are actually build in different ways they should explicitly specify what is done to build the image (e.g. first build a binary and then build a Dockerfile). This makes it easier for developers to see what's going on.

This also handles parts of #426


Makefile
```diff
@@ -27,7 +27,7 @@ run-integration-tests:
 	$(MAKE) kind
 	$(MAKE) cluster-prepare
 	$(MAKE) docker
-	$(MAKE) -C test/services docker-all
+	$(MAKE) -C test/services docker-build-and-push
```
roee88, line 30: Why is test/services special? (`docker-build-and-push`)

connectors/Makefile
```diff
@@ -2,6 +2,7 @@ ROOT_DIR := ..
 include $(ROOT_DIR)/Makefile.env
 
 CONNECTORS := \
+	katalog \
```
roee88, line 5: Conflicts with the PR by @revit13, please coordinate if this is something that you continue. 

I'm not a fan of this Makefile by the way, I prefer each connector to decide what targets it wants to have and be executed independently (e.g., I don't really want to copy terminate or any of the deployments targets to katalog, I don't want to deploy both egeria and katalog, etc.)

connectors/egeria/Makefile
```diff
@@ -10,6 +10,7 @@ build:
 	mvn -B clean package -DskipTests
 
 docker-build: build
+	docker build . -t ${IMG} -f Dockerfile
```
roee88, line 13: Why do we need `-f Dockerfile`

connectors/katalog/Makefile
```diff
@@ -13,14 +13,24 @@ generate: $(TOOLBIN)/oapi-codegen $(TOOLBIN)/crdoc $(TOOLBIN)/openapi2crd
...
+ifeq ($(NEED_TEST_COVERAGE_STATISTICS),1)
+	go test -v ./... -coverprofile=coverage.out
+	go tool cover -func=coverage.out
+    ifeq ($(NEED_TEST_COVERAGE_STATISTICS_IN_HTML),1)
+		go tool cover -html=coverage.out
+    endif
+	rm coverage.out
+else
+	go test -v ./...
+endif
```
roee88, line 36: Please change to just `go test -v ./...`. When we handle coverage tooling it will be part of the entire project.

roee88 (commented):
Thanks

connectors/egeria/Makefile
froesef: You're right we don't as `-f Dockerfile` is the default. Should we leave it so it's clear which file is taken?

connectors/Makefile
froesef: I agree there are too many targets here. I propose to remove the following targets from `connectors/Makefile`:
- build
- clean
- run
- terminate

These are all targets that are going to be called when working on or testing one connector.

shlomitk1 (approved):

connectors/egeria/Makefile
roee88: Having it makes it less clear to me, as if you do something special although you don't. Maybe it's just me.

revit13 (approved):

connectors/Makefile
roee88: Okay, but this makefile is currently broken so either do it in this PR or remove this change. If done in this PR then please also;
- remove charts target
- add empty deploy/undeploy targets to katalog to keep this makefile working

connectors/opa/README.md
```diff
@@ -5,7 +5,7 @@ For each of the tasks related to the opa connector, execute the corresponding co
 
 * Build the connector: `make build`
```
roee88, line 6: You removed the build target

connectors/Makefile
froesef: I kept the charts targets as they are as they're not based on the CONNECTORS array but explicitly specify the three connectors. I'll leave them until we're sure we have everything we need in helm and then we can remove them.

connectors/opa/README.md
froesef: Not for opa. The removed build target above was in katalog.

connectors/Makefile
roee88: Sure. They will override the file in charts so no one should execute them.

roee88 (approved):

PR #431, opened 2021-03-16
Add a kuberentes secret for Vault token in vault helm installation.

revit13:
Co-authored-by: Roee Shlomo <roee88@gmail.com>

### DIFF
--- a/charts/Makefile
+++ b/charts/Makefile
@@ -20,11 +20,16 @@ cert-manager:
 .PHONY: vault
 vault:
 	@echo "Installing vault ..."
+	kubectl create namespace m4d-system || true
 	helm repo add hashicorp https://helm.releases.hashicorp.com
 	helm install vault hashicorp/vault --version 0.8.0 \
                 --set "global.openshift=false" \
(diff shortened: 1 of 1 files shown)


roee88 (approved):

roee88:
Thanks. Please open an issue to improve this as nothing should login as root.

revit13:
@roee88 I opened an issue. Maybe we should consider installing Vault in non dev mode? it seems that we assume we need the token in k8s secret which missed the point of using dev mode where the token is known...

roee88:
I don't mind that the value is root and is known when you install it in dev mode or otherwise, it's not our problem and that's not what I meant. 

What I meant is that this secret you added here is actually not for vault but rather for the manager to login to vault. A proper method to allow the manager to login to vault should be added instead, not relying on such a secret that always has root in it. Be it a secret dedicated to the manager that the user configures (as part of m4d helm install) or using Kubernetes login.

revit13:
@roee88 OK thanks. I'll change the issue to reflect that.

PR #434, opened 2021-03-16
Tidy up test targets

froesef:
This PR adds the connectors and pkg directly to the `test` target of the main make file.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,9 +16,6 @@ jobs:
   verify:
     name: Verify
     runs-on: ubuntu-latest
-    env:
-      DOCKER_HOSTNAME: "localhost:5000"
-      DOCKER_NAMESPACE: m4d-system
     steps:
     - name: Install Go
       uses: actions/setup-go@v2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,9 +42,6 @@ jobs:
   test:
     name: Unit tests
     runs-on: ubuntu-latest
-    env:
-      DOCKER_HOSTNAME: "localhost:5000"
-      DOCKER_NAMESPACE: m4d-system
     steps:
     - name: Set up JDK
       uses: actions/setup-java@v1
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,8 +9,6 @@ on:
       - master
 
 env:
-  DOCKER_HOSTNAME: ghcr.io
-  DOCKER_USERNAME: the-mesh-for-data
   GO_VERSION: 1.13
   JAVA_VERSION: 11
 
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,7 @@ run-integration-tests:
 	$(MAKE) -C manager deploy_it
 	$(MAKE) -C manager wait_for_manager
 	$(MAKE) helm
+	$(MAKE) -C pkg/helm test
 	$(MAKE) -C manager run-integration-tests
 
 .PHONY: run-deploy-tests
--- a/Makefile
+++ b/Makefile
@@ -17,9 +17,8 @@ build:
 
 .PHONY: test
 test:
-	$(MAKE) -C connectors test
-	$(MAKE) -C pkg test
-	$(MAKE) -C manager test
+	go test -v ./...
+	# The tests for connectors/egeria are dropped because there are none
 
 .PHONY: run-integration-tests
 run-integration-tests: export DOCKER_HOSTNAME?=localhost:5000
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,8 @@ build:
 
 .PHONY: test
 test:
-	$(MAKE) -C pkg/policy-compiler test
+	$(MAKE) -C connectors test
+	$(MAKE) -C pkg test
 	$(MAKE) -C manager test
 
 .PHONY: run-integration-tests
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ build:
 
 .PHONY: test
 test:
+	$(MAKE) -C manager pre-test
 	go test -v ./...
 	# The tests for connectors/egeria are dropped because there are none
 
--- a/manager/Makefile
+++ b/manager/Makefile
@@ -15,6 +15,8 @@ include $(ROOT_DIR)/hack/make-rules/verify.mk
 config/default/kustomization.yaml: config/default/kustomization.yaml.in
 	cp $< $@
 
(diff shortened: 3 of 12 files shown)


pkg/helm/helm_test.go
```diff
@@ -79,6 +79,11 @@ func TestHelmCache(t *testing.T) {
 }
 
 func TestHelmRegistry(t *testing.T) {
+	// Test should only run as integration test if registry is available
+	if host, isSet := os.LookupEnv("DOCKER_HOSTNAME"); !isSet || host != "localhost:5000" {
```
shlomitk1, line 83: Not sure about this. Theoretically, I can run integration tests from any environment.

pkg/helm/helm_test.go
```diff
@@ -109,6 +114,10 @@ func TestHelmRegistry(t *testing.T) {
 }
 
 func TestHelmRelease(t *testing.T) {
+	// Test should only run as integration test if registry is available
+	if host, isSet := os.LookupEnv("DOCKER_HOSTNAME"); !isSet || host != "localhost:5000" {
```
shlomitk1, line 118: Same comment

pkg/vault/vault_interface_test.go
```diff
@@ -21,7 +22,7 @@ func Log(t *testing.T, label string, err error) {
 
 func TestCredentialManagerInterface(t *testing.T) {
 	var err error
-	os.Setenv("RUN_WITHOUT_VAULT", "0")
+	os.Setenv("RUN_WITHOUT_VAULT", "1")
```
shlomitk1, line 25: We want to test real vault connection, not the dummy interface.

pkg/helm/helm_test.go
froesef: True. So how about dropping the host check and only skipping the test if the `DOCKER_HOSTNAME` environment variable is not set?

pkg/vault/vault_interface_test.go
froesef: Yes. Is there a real vault instance we can run against? E.g. in the integration test one could run against vault and in the unit test against the dummy?

pkg/helm/helm_test.go
shlomitk1: Agree.

pkg/vault/vault_interface_test.go
shlomitk1: Both helm and vault packages are tested in integration tests only, not in unit tests, if I am not mistaken.

roee88:
Thanks for this PR, you're touching a painful issue in the codebase.

Some general comments:

1. This project is a tree of makefiles, I keep getting lost and it keeps introducing duplications and errors. I strongly recommend to stop using this pattern or at least limit its use to minimum. For example, *none* of the makefiles in pkg should exist.
2. `go test ./...` runs recursively. You can't run it in pkg and then separately in things under pkg, it's duplicate work. It's actually more than duplicate work, it will run tests that require vault to be setup but those should only run through the test script in pkg/vault. Which brings me to the next item
3. We should try to minimize situations that *require* setting up things outside of the test suites (good example: motion controller tests; bad example: pkg/vault test). For example for vault testing see https://stackoverflow.com/a/57773764.

shlomitk1:
> 
> 
> Otherwise, a clear separation of integration tests is needed.

We don't really want to test vault, only that the vault deployment in the control plane is working as expected. Same goes for helm. 
This is another problem: how to run tests that do require a full setup? Running same tests both in unit tests and as end-to-end tests in an existing cluster has its issues - hence special casing whether EXISTING_CONTROLLER/CLUSTER is set or not.

roee88:
> > Otherwise, a clear separation of integration tests is needed.
> 
> We don't really want to test vault, only that the vault deployment in the control plane is working as expected. Same goes for helm.

I don't understand this statement. The tests in both cases test the client code in those packages. It doesn't and shouldn't test that vault is installed correctly as far as I can tell.

shlomitk1:
> Agree. My question is do we want to do this special casing for every test or just create completely septate integrations test (e.g. in /test).

I vote for completely separate integration tests.

pkg/vault/vault_interface_test.go
froesef: They are currently not tested at all. And for the integration tests they need an exposed port that currently is not exposed in the integration tests.

froesef:
I think what the vault test should test is if our programs can run against a vault interface. (And as @roee88 mentioned there are utilities from vault that are designed to be used in such unit test situations. Thus there is no need for an external vault setup (or integration tests))

The same is actually also valid for out helm and integration tests. There we want to test against a service that exposes a helm api or a Kubernetes api (e.g. when posting Blueprints, etc in the integration tests) But as there are no good in memory alternatives for the two that simulate the appropriate behaviour (e.g. also run pods etc) we're running these as integration tests against a kind environment.

I think we should keep the tests in the packages where their code is if they are specialised e.g. on just motion, app, helm etc. 

If we're really testing a full fledged custom setup where we're writing test values into vault, constructing an M4dApplication, installing modules etc in order to e.g. run a real copy and afterwards check S3 if it was really executed this can be under the `test` folder.

Lastly I agree with @roee88 that we should at some point just have one `go test ./...` in the root.

Makefile
```diff
@@ -17,7 +17,8 @@ build:
 
 .PHONY: test
 test:
-	$(MAKE) -C pkg/policy-compiler test
+	$(MAKE) -C connectors test
+	$(MAKE) -C pkg test
```
roee88, line 21: I suggest to remove this (and the makefile in pkg) from this PR. It will break tests unless ugly workarounds are added.

roee88: Related to my other comment about this running recursively also the vault tests.

.github/workflows/build.yml
```diff
@@ -9,8 +9,6 @@ on:
       - master
 
 env:
-  DOCKER_HOSTNAME: ghcr.io
```
roee88, line 12: You can remove these from the verify job too (I think)

manager/Makefile
```diff
@@ -15,6 +15,8 @@ include $(ROOT_DIR)/hack/make-rules/verify.mk
 config/default/kustomization.yaml: config/default/kustomization.yaml.in
 	cp $< $@
 
+pre-test: generate manifests $(TOOLBIN)/etcd $(TOOLBIN)/kube-apiserver
```
roee88, line 18: Minor: you can use this in `test` below to avoid repeating.

roee88 (approved):

roee88 (approved):

PR #437, opened 2021-03-17
Add to vault helm installation different values.yaml files for Vault …

revit13:
…secret plugins (#414)

off-by: Revital Sur <eres@il.ibm.com>
Co-authored-by: Roee Shlomo <roee88@gmail.com>

This PR addresses issue #414 .
Note: this is a follow up of PR #417 which was closed due to technical reasons.

To retrieve the dataset credentials from where they are stored Vault secrets plugins are used.
This PR uses kubernetes secrets plugin: https://github.com/mesh-for-data/vault-plugin-secrets-kubernetes-reader for the default Vault installation. This is done by extending Vault helm installation to use a values.yaml file with the appropriate Vault secret plugin details.


charts/Makefile
```diff
@@ -1,6 +1,8 @@
 ROOT_DIR := ..
 include $(ROOT_DIR)/Makefile.env
 
+WITHOUT_OPENSHIFT ?= true
```
roee88, line 4: Why without? Make it positive. It will also fix the wrong usage bellow :)

charts/Makefile
```diff
@@ -22,14 +24,18 @@ vault:
 	@echo "Installing vault ..."
 	kubectl create namespace m4d-system || true
 	helm repo add hashicorp https://helm.releases.hashicorp.com
-	helm install vault hashicorp/vault --version 0.8.0 \
-                --set "global.openshift=false" \
+	helm install vault hashicorp/vault --version 0.9.1 \
+                --set "global.openshift=$(WITHOUT_OPENSHIFT)" \
                 --set "server.dev.enabled=true" \
+		--values ../third_party/vault/values/plugin-secrets-kubernetes-reader.yaml \
```
roee88, line 30: Did you check that it creates the spec you expected? It's unclear from the Helm docs if -f (--values) and --set can be used together. I didn't check.

charts/Makefile
revit13: Other places in the code such as third_party/egeria/deploy.sh uses WITHOUT_OPENSHIFT ... should I leave it afterall to make it consistent? tnx

charts/Makefile
revit13: Yes it seems to work ok: the server starts in dev mode and the plugin is enabled.

charts/Makefile
roee88: You pass it as is to `global.openshift` so usage is wrong here (installing with openshift=true when WITHOUT_OPENSHIFT=true.

As far as I can tell WITHOUT_OPENSHIFT is only used in the old vault deploy script and in Egeria deploy script. I think it would be better to change it in the other places instead but whatever you want.

revit13: OK, I'll change it then (and fix the usage) tnx

roee88 (commented):
The content of `third_party/vault` is a bit messy because we don't cleanup the old way yet. Can you please consider grouping the new files you added to `third_party/vault` into a subdirectory? For example:

```bash
third_party/vault/plugin-secrets-kubernetes-reader/
├── clusterrole.yaml
├── clusterrolebinding.yaml
├── values.yaml
└── README.md
```

roee88 (approved):

PR #438, opened 2021-03-17
Fix checking of status indicators that are not of type string

shlomitk1:

Bug fix: numeric indicators such as `status.succeeded > 0` have not been evaluated.
Solution: use NestedFieldNoCopy instead of NestedString.

### DIFF
--- a/manager/controllers/utils/unstructured.go
+++ b/manager/controllers/utils/unstructured.go
@@ -4,6 +4,7 @@
 package utils
 
 import (
+	"encoding/json"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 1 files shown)


froesef (approved):

manager/controllers/utils/unstructured.go
```diff
@@ -47,6 +48,15 @@ func (c UnstructuredAsLabels) Get(label string) string {
...
 	fields := strings.Split(label, ".")
 	// not checking whether the label exists and is valid. Assuming Get is called after Has in labels package evaluation
-	val, _, _ := unstructured.NestedString(obj, fields...)
-	return val
+	// fetch a string value
+	if valStr, _, err := unstructured.NestedString(obj, fields...); err == nil {
+		return valStr
+	}
+	// convert a received interface into a string
+	val, _, _ := unstructured.NestedFieldNoCopy(obj, fields...)
```
roee88, line 56: Is it really okay to ignore the error here? (last `_`)

shlomitk1: It's supposed to be applied after a successful Has function that checks the value, but I'll add the sanity check.

roee88 (approved):

PR #443, opened 2021-03-18
Activate GH Actions as main CI that pushes images

froesef:
Let's do the switch. GH Actions has been stable so far. Actually Travis fails more often latetly.


roee88 (approved):
What would be the trigger to delete travis.yaml? I'm fine with removing it right away, am I missing something?

froesef:
Exactly. Deleting travis.yaml will remove the travis checks. But I noticed that we need to add the credentials as repo secrets. Can you please do that?

PR #447, opened 2021-03-21
Data Catalog Taxonomy

simanadler:
Creates the initial taxonomy containing the data catalog related terms.  These terms are used in other components as well (ex: policy manager, modules, etc) and thus the taxonomy is the mechanism via which common terms are managed and validates.

Currently this taxonomy is manually created.  In the future it should be generated by querying the relevant data catalog - probably as part of the catalog connector.

Includes tests, which also show how to use the json schema validator.


third_party/taxonomy/Makefile
```diff
@@ -0,0 +1,15 @@
...
+deploy: $(TOOLBIN)/kubectl 
+	$(TOOLBIN)/kubectl create configmap taxonomy-configmap --from-file=module_schema.json --from-file=data_catalog_schema.json --from-file=policy_manager_schema.json
+
+.PHONY: undeploy
+undeploy: $(TOOLBIN)/kubectl 
+	$(TOOLBIN)/kubectl delete configmap taxonomy-configmap
+
+.PHONY: test
+test:
+	go test ./... -coverprofile cover.out
```
shlomitk1, line 15: Does this test run in CI?

simanadler: Not yet.  I will add the tests to the CI after I finish the policy and module taxonomies

roee88:
I'm writing a document for writing taxonomy specification (focus on schema objects) to have what we discussed verbally written down. I'll validate the schema here against it when done.

third_party/taxonomy/data_catalog_schema.json
```diff
@@ -0,0 +1,103 @@
...
+            "type": "object",
+            "description": "Data set or other type of resource.",
+            "properties": {
+                "name": {
+                    "type": "string",
+                    "description": "Unique name of the data set"
+                },
+                "category": { "type": "string", "$ref": "#/definitions/data_type"},
+                "location": { "type": "object", "$ref": "#/definitions/geography"},
+                "tags": { "$ref": "#/definitions/metadata"},
```
ronenkat, line 71: why not have category and location as tags? i.e., dataset is a set of {name, tags, columns}
Tags already include data sensitivity and geography.

ronenkat (commented):
looks good.
The taxonomy should move out f 3rd party and be split into config? and samples (or similiar) in the repo, and be built into config/taxonomy during install.

third_party/taxonomy/data_catalog_schema.json
```diff
@@ -0,0 +1,103 @@
...
+            "type": "string",
+            "description": "Format in which the data is stored.",
+            "enum": ["parquet", "csv", "arrow", "json", "table", "avro", "binary"]
+        },
+        "data_type": {
+            "type": "string",
+            "description": "Type of data.",
+            "enum": ["tabular", "image", "text", "voice", "video"]
+        },
+        "metadata": {
```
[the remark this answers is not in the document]
simanadler: I separated the structures from the values

third_party/taxonomy/data_catalog_schema.json
simanadler: done

roee88:
> looks good.
> The taxonomy should move out f 3rd party and be split into config? and samples (or similiar) in the repo, and be built into config/taxonomy during install.

Can we just put it in `config/taxonomy` directly in this PR?

roee88:
BTW the structure for resource need to change in a follow-up PR to match the design (i.e., tags as a map like in the current implementation for performance and easier policy writing).

roee88 (approved):
If you can move to a top level config folder not under manager it would be great. Anyway LGTM :)

manager/config/taxonomy/catalog.values.schema.json
```diff
@@ -0,0 +1,49 @@
...
+        "data_type": {
+            "type": "string",
+            "description": "Type of data.",
+            "enum": ["tabular", "image", "text", "voice", "video"]
+        },
+        "valid_metadata": {
+            "type": "array",
+            "description": "A list of valid key value metadata pairs describing aspects of the data.  Values must be one of the items indicated.",
+            "items": {
+                "properties": {
```
roee88, line 31: Please add `type` and set it to "object".

PR #449, opened 2021-03-22
Remove .travis.yaml

froesef:
Let's remove Travis CI as we moved our image builds and pushes to Github Actions.

This should be merged if everyone agrees that Github Actions is running stable.


froesef:
For this to pass we'll have to deactivate the Travis CI check as mandatory as it won't run any more.

roee88 (approved):

PR #450, opened 2021-03-22
charts: added Istio resources

roee88:
https://github.com/IBM/the-mesh-for-data/issues/348

added Istio resources for control plane security


revit13 (approved):

roee88:
> Not sure if we want to specify that that these yaml files should be applied in m4d-system (at least the Sidecar yaml)...

We don't set namespace, it's using the namespace that the chart is installed to. Currently it must be m4d-system but the charts are a step towards allowing something else.

What I wonder about is whether the files in this PR should be part of the charts at all, or just as part of the instructions to enable control plane security.

roee88:
Closing for now, will handle it better in future PR.

PR #454, opened 2021-03-22
Use helm chart for integration tests

### LINKED ISSUE #452: Deploy integration tests using helm chart
Deploy the integration tests using the new helm chart to ensure that the deployment is working as expected and the components are functioning as expected.

froesef:
Fixes #452


roee88:
Can you please change the new values files to only include the diff from values.yaml?

charts/m4d/templates/_helpers.tpl
```diff
@@ -66,19 +66,24 @@ Create the value of an image field from hub, image and tag
 {{/*
 isEnabled evaluates an enabled flag that might be set to "auto".
 Returns true if one of the following is true:
+The return value when using `include` is always a String.
```
roee88, line 69: Thanks!

Makefile
```diff
@@ -23,17 +23,18 @@ test:
 .PHONY: run-integration-tests
 run-integration-tests: export DOCKER_HOSTNAME?=localhost:5000
 run-integration-tests: export DOCKER_NAMESPACE?=m4d-system
+run-integration-tests: export VALUES_FILE=m4d/integration-tests.values.yaml
 run-integration-tests:
 	$(MAKE) kind
-	$(MAKE) cluster-prepare
+	$(MAKE) -C charts vault
+	$(MAKE) -C charts cert-manager
+	kubectl apply -f https://raw.githubusercontent.com/IBM/dataset-lifecycle-framework/master/release-tools/manifests/dlf.yaml
```
roee88, line 31: This repeats too many times so I think a make target for dlf (now datashim) will be better.

charts/m4d/values.yaml
```diff
@@ -33,11 +33,11 @@ global:
 # Cluster metadata values
 cluster:
   # Set to the name of the cluster.
-  name: hobbiton
-  # Set to cluster region.
-  region: eriador
+  name: thegreendragon
   # Set to cluster zone.
-  zone: theshire
+  zone: hobbiton
```
roee88, line 38: Important fix ;)

manager/Makefile
```diff
@@ -126,7 +126,7 @@ undeploy_control_plane_security: $(TOOLBIN)/kustomize $(TOOLBIN)/kubectl config/
 
 .PHONY: wait_for_manager
 wait_for_manager: $(TOOLBIN)/kubectl
-	$(TOOLBIN)/kubectl wait --for=condition=available -n ${KUBE_NAMESPACE} deployment/m4d-controller-manager --timeout=120s
+	$(TOOLBIN)/kubectl wait --for=condition=available -n ${KUBE_NAMESPACE} deployment/manager --timeout=120s
```
roee88, line 129: I don't think we need this anymore. We install the chart with --wait.

manager/config/default/manager_auth_proxy_patch.yaml
```diff
@@ -6,7 +6,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: controller-manager
+  name: manager
```
roee88, line 9: Just to understand where we're at: are these files still being used? Where? 

Question applies to all changes in manager/config. I'm trying to understand if they are still used in integration tests.

hack/setup-local-multi-cluster.sh
```diff
@@ -5,29 +5,31 @@
...
 make -C manager wait_for_manager
 make -C modules helm-chart-push
 
+export VALUES_FILE=m4d/kind-kind.values.yaml
 kubectl config use-context kind-kind
-make cluster-prepare
+make -C charts vault
+make -C charts cert-manager
+kubectl apply -f https://raw.githubusercontent.com/IBM/dataset-lifecycle-framework/master/release-tools/manifests/dlf.yaml
 make cluster-prepare-wait
```
roee88, line 30: I think here too only waiting for dlf is needed.

manager/config/default/manager_auth_proxy_patch.yaml
froesef: They're not being used. But it was easier for me to change them to when I was verifying that both deployment mechanisms are doing the same expected thing. I'd propose that we remove them at a later point when we're sure that everything can be removed.

hack/setup-local-multi-cluster.sh
froesef: Yes. But aggregating the waiting had a point of being able to deploy many things in parallel and then have one huge wait loop before continuing. This sped up the tests and deployments quite a bit. An alternative is to call all these targets in parallel.

roee88 (approved):
There are redundant steps due to `--wait` but we can improve that later.

PR #466, opened 2021-03-25
Taxonomy: added document with guidelines for writing schemas

roee88:
Guidelines for writing the schemas within a JSON Schema document used for taxonomy.

cross-reference: #447

### DIFF
--- a/config/taxonomy/HOWTO_SPEC.md
+++ b/config/taxonomy/HOWTO_SPEC.md
@@ -19,7 +19,6 @@ You SHOULD declare the schema version in the document:
 The goal is to preserve compatibility with json schema DRAFT 4, OpenAPI 3.0, and Kubernetes CRDs (all are based on json schema DRAFT 4), and code generation tools.
 We hope that newer json schema specs will be supported in the future as we improve our tooling.
 
-
 ## Schema properties
 
 In this section we describe all schema properties that can be used. We categorize the properties to [structural properties](#structural-properties) and [validation properties](#validation-properties).
@@ -231,23 +230,31 @@ UUID | `string` | `uuid` | a UUID
 
 Any time a schema object can be used, a reference object can be used in its place. This allows referencing definitions instead of defining them inline.
 
+Assume that all schema files are mounted in a common folder so you can reference other taxonomy JSON files by their name. 
+
+### Examples
+
+A reference to a schema defined in an external file:
+
 ```json
-// a reference to a schema defined in an external file
 {
-  "$ref": "definitions.json#/Pet"
+  "$ref": "another.json#/definitions/Pet"
 }
 ```
 
+A reference to a schema defined in the same file:
+
 ```json
-// a reference to a schema defined in the same file
 {
   "$ref": "#/definitions/Pet"
 }
 ```
 
+A more complete example:
+
 ```json
-// a more complete example
 {
+    "$schema": "http://json-schema.org/draft-04/schema#",
     "definitions": {
         "Part": {
             "type": "string",
@@ -264,6 +271,7 @@ Any time a schema object can be used, a reference object can be used in its plac
                 }
             }
         }
+    }
 }
 ```
 
--- a/config/taxonomy/HOWTO_SPEC.md
+++ b/config/taxonomy/HOWTO_SPEC.md
@@ -6,9 +6,19 @@ This document lists the requirements for writing such schemas.
 ## What spec to follow?
 
 We use a subset of json schema DRAFT 4 that is described in this document. 
-The goals are to preserve compatibility with json schema DRAFT 4, OpenAPI 3.0, and Kubernetes CRDs (all are based on json schema DRAFT 4), and code generation tools.
 
+You SHOULD declare the schema version in the document:
+
+```json
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    ...
+}
+```
+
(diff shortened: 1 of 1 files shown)


simanadler (commented):
This copies a lot of information from other documents.  You usually prefer to reference other documents, so I'm a bit surprised at the approach here.

roee88:
> This copies a lot of information from other documents. You usually prefer to reference other documents, so I'm a bit surprised at the approach here.

There is no single source that exactly match what I wrote here. It's a subset from the documents listed at the bottom. Is there an issue with a self contained document?

simanadler:
> > This copies a lot of information from other documents. You usually prefer to reference other documents, so I'm a bit surprised at the approach here.
> 
> There is no single source that exactly match what I wrote here. It's a subset from the documents listed at the bottom. Is there an issue with a self contained document?

No, of course not

config/taxonomy/HOWTO_SPEC.md
```diff
@@ -0,0 +1,266 @@
...
+email | `string` | `email` | Internet email address
+hostname | `string` | `hostname` | an Internet host name
+IPv4 | `string` | `ipv4` | IPv4 address according to the "dotted-quad" ABNF syntax
+IPv6 | `string` | `ipv6` | IPv6 address 
+URI  | `string` | `uri` | a valid URI, according to [RFC3986](https://tools.ietf.org/html/rfc3986)
+UUID | `string` | `uuid` | a UUID
+
+<!-- TODO: update openapi2crd to prune format if not one of: "int32", "int64", "float", "double", "byte", "date", "date-time", "password" -->
+
+## Reference objects
```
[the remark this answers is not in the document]
roee88: I think we should at least start without using the id property. The schema files will be loaded to manager and other components as local files (volume mount from configmaps) in the same directory.  They are not published anywhere. All references are relative to the loaded taxonomy folder. The first example is to an external file.

config/taxonomy/HOWTO_SPEC.md
```diff
@@ -0,0 +1,276 @@
...
+    ```
+2. Use `additionalProperties` with object type to define a map
+    ```json
+    // map from string to integer values
+    {
+        "type": "object",
+        "additionalProperties": {
+            "type": "integer"
+        }
+    }
```
simanadler, line 68: Where in this example does it indicate that the key of the map is a string?  If you use this definition and provide {5:2} it works and so does {"whatever":2}.  {"whatever":"xxx"} will of course fail because "xxx" is not an integer.

config/taxonomy/HOWTO_SPEC.md
```diff
@@ -0,0 +1,276 @@
...
+    ```
+3. Use `items` to define the type of array items
+    ```json
+    // array of strings
+    {
+        "type": "array",
+        "items": {
+            "type": "string"
+        }
+    }
```
simanadler, line 78: When creating an array of geography items I had to define it slightly differently in order for the required name property in geography to be enforced.  "items": { "properties": { "geography": {"$ref": "#/definitions/geography"}}}

config/taxonomy/HOWTO_SPEC.md
```diff
@@ -0,0 +1,276 @@
...
+
+You SHOULD declare the schema version in the document:
+
+```json
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    ...
+}
+```
+
```
simanadler, line 18: Perhaps mention that in order to for one file to reference a different schema file, the files should have IDs and that they should assume all schema files are mounted in a common folder.  Ex: "$id": "catalog.structs.schema.json"

simanadler (commented):
See comments

simanadler:
Do you want to add a section (perhaps in a follow up PR) about how to test the schemas?

config/taxonomy/HOWTO_SPEC.md
roee88: AFAIK JSON keys must be strings.

config/taxonomy/HOWTO_SPEC.md
roee88: You definition should be fixed to:
```json
    {
        "type": "array",
        "items": {
            "type": "object",
            "properties": {
              "geography": {
                "$ref": "#/definitions/geography"
              }
            }
        }
    }
```

It's the same as here but defining an object rather than a string as the array item type.

config/taxonomy/HOWTO_SPEC.md
roee88: Like I wrote before, I don't think that we should add `$id` now https://github.com/IBM/the-mesh-for-data/pull/466#discussion_r606699694. BTW in Draft 4 `$id` is actually `id`. Are you sure it's necessary? Does it make sense for non published files? The spec says that it must be a URI. 

> assume all schema files are mounted in a common folder 

Will add.

roee88:
> Do you want to add a section (perhaps in a follow up PR) about how to test the schemas?

Yes we should add that too but we first need to discuss the approach for testing though. IMHO there should be in `pkg/taxonomy` a `ValidateDocument` function (the method you wrote, I don't remember the name) and `ValidateSchema` function that ensures that the schema respects this document (I will write one in the future). Then in testing we should always call ValidateSchema and only call ValidateDocument on the actual structures that we use in manager or elsewhere (i.e., not on passthrough values).

config/taxonomy/HOWTO_SPEC.md
roee88: Added in the `Reference objects` section

simanadler (approved):

PR #492, opened 2021-04-07
governance actions in read scenario

shlomitk1:
Current implementation does not reflect read and ingest use-cases. For read scenario, after receiving a list of transformations to perform on sensitive data given the workload location, if we ask for actions on copying the data to the workload location - we need to reflect the required transformations or we should get a list of actions including the required transformations. But then, this contradicts the ingest scenario when we want to find a suitable location to copy data as-is. Leaving the interpretation for WRITE to copy data as-is, this work-around does not involve WRITE actions in the read scenario use-case.

### DIFF
--- a/manager/controllers/app/m4dapplication_controller_test.go
+++ b/manager/controllers/app/m4dapplication_controller_test.go
@@ -342,45 +342,6 @@ var _ = Describe("M4DApplication Controller", func() {
 			Expect(getErrorMessages(resource)).To(ContainSubstring("read"))
 		})
 
-		// Tests denial of the necessary copy operation
-
-		// Assumptions on response from connectors:
-		// s3 dataset
-		// Read module with source=s3,parquet exists
-		// Copy to s3 is required by the user
-		// Enforcement action for write operation: Deny
-		// Result: an error
-
-		It("Test deny-on-copy", func() {
-			appSignature := types.NamespacedName{Name: "with-copy", Namespace: "default"}
-			resource := InitM4DApplication(appSignature.Name, 1)
-			resource.Spec.Data[0] = apiv1alpha1.DataContext{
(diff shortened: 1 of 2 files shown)


simanadler:
@shlomitk1 I'm not sure I understand the problem.  The policies for the ingest scenario should be: (1) if write, and if dataresidency=Turkey and data=Confidential and destination=Netherlands then **deny** (2) if write, and if dataresidency=Turkey and data=Confidential and destination=Netherlands then **allow**.  The policy for the data usage scenario is the same we have been using all along (3) if read, and if dataresidency=Turkey and if intent = Fraud Detection and processingLocation=Netherlands, then **allow** but mask columns with Confidential data

What is the conflict?  During the ingest scenario the query to the policy compiler is write, and in the data usage scenario it's read

shlomitk1:
> 
> 
> @shlomitk1 I'm not sure I understand the problem. The policies for the ingest scenario should be: (1) if write, and if dataresidency=Turkey and data=Confidential and destination=Netherlands then **deny** (2) if write, and if dataresidency=Turkey and data=Confidential and destination=Netherlands then **allow**. The policy for the data usage scenario is the same we have been using all along (3) if read, and if dataresidency=Turkey and if intent = Fraud Detection and processingLocation=Netherlands, then **allow** but mask columns with Confidential data
> 
> What is the conflict? During the ingest scenario the query to the policy compiler is write, and in the data usage scenario it's read

At some point we had an additional set of actions to be applied on top of the read actions. Use-case: for read-only - redact data, but if you want to copy data - you need also to encrypt the data. That's why we were sending two requests: one for READ, another for COPY/WRITE. This PR abandons this idea and returns the implementation to what you are saying (either READ or WRITE depending on the scenario type).

simanadler (approved):

PR #493, opened 2021-04-07
Add possibility to do custom builds to fork's ghcr.io

froesef:
This will add the possibility to easily do custom build to the GitHub packages of a fork.
E.g. users can manually publish images produced by a branch to their GitHub container registry so that they can use them and try them out before doing a PR or merging to master.

Please tell me if the name "Custom build" is fine for this category or not.

### DIFF
--- a/.github/workflows/debug_build.yml
+++ b/.github/workflows/debug_build.yml
@@ -14,6 +14,7 @@ env:
 
 jobs:
   build-and-publish:
+    if: github.repository != 'IBM/the-mesh-for-data'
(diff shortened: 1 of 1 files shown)


roee88:
Can you add protection to not never run this in the upstream? Like the opposite of https://github.com/xskipper-io/xskipper/blob/363408c2449e74c1f0f3b534cfdd782719b42b9f/.github/workflows/release.yml#L18

roee88 (approved):

PR #496, opened 2021-04-07
Publish docker images and helm charts on tags

froesef:
I decided to keep the logic in the make file. It seemed like less of a hack than the single github steps.

Although. If the make file is pushing latest and tag version tags for a tag build and the master branch is pushing latest. Aren't we pushing latest twice for the same commit? Shouldn't it be an either or as the CI is triggered twice?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - master
+    tags:
+      - '*'
   pull_request:
     branches:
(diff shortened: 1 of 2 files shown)


roee88:
Good point. Technically latest should be the latest released version. Usually master builds have dev or master as the docker tag, or with a bit more investment the next version number with a SNAPSHOT indicator.

roee88 (approved):

PR #498, opened 2021-04-07
Adding default OPA policy

### LINKED ISSUE #490: OPA connector fails when no policies are defined
When no policies are defined a default should be used (e.g., allow) rather then erroring.

Currently this error shows up:
```
Could not select a read module for default/data-csv : rpc error: code = Unknown desc = error in GetOPAOperationDecision (i = 0): error in format of OPA evaluation (incorrect result map)
```

rohithdv:
Fixes #490.

### DIFF
--- a/connectors/opa/lib/evaluate_opa.go
+++ b/connectors/opa/lib/evaluate_opa.go
@@ -5,7 +5,6 @@ package lib
 
 import (
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"log"
 	"net/http"
@@ -39,19 +38,19 @@ func performHTTPReq(standardClient *http.Client, address string, httpMethod stri
 	if err != nil {
 		log.Fatal("Error:", err)
 	}
-	fmt.Println(httpMethod + " succeeded")
+	log.Println(httpMethod + " succeeded")
 
 	return res
 }
 
-func doesOpaHaveUserPoliciesLoaded(responsedata string) (string, bool) {
-	decisionid, _ := jsonparser.GetString([]byte(responsedata), "decision_id")
+func doesOpaHaveUserPoliciesLoaded(responsedata []byte) (string, bool) {
+	decisionid, _ := jsonparser.GetString(responsedata, "decision_id")
 
-	fmt.Printf("decision_id: %s", decisionid)
-	if value, _, _, err := jsonparser.Get([]byte(responsedata), "result"); err == nil {
-		fmt.Printf("result: %s", value)
+	log.Printf("decision_id: %s", decisionid)
+	if value, _, _, err := jsonparser.Get(responsedata, "result"); err == nil {
+		log.Printf("result: %s", value)
 	} else {
-		fmt.Printf("Result Key does not exist implying no policies are loaded in opa")
+		log.Printf("Result Key does not exist implying no policies are loaded in opa")
 		return decisionid, false
 	}
 	return decisionid, true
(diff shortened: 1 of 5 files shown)


shlomitk1:
@rohithdv How does this work? This policy is loaded by default, so even if we have a user policy defined, a deny action will always be true, won't it?

rohithdv:
> @rohithdv How does this work? This policy is loaded by default, so even if we have a user policy defined, a deny action will always be true, won't it?

Yes true. The earlier fix was not correct. I have updated the fix. Please check the latest changes in this PR. Thanks. @shlomitk1

shlomitk1 (approved):

ronenkat (commented):
Is there a way to have the decision not in the code?
please document the default decision.
was this tested end-to-end?

roee88:
> Is there a way to have the decision not in the code?

I personally prefer the approach here. The code must not fail if there are no rules. Thus the default is allow but to change it to deny by default you add a policy that denies if no other rules are triggered (currently not straight forward but with the new opa policies design it is).

 a policy to deny if there are no other rules triggered to make it deny by default.

ronenkat:
> > Is there a way to have the decision not in the code?
> 
> I personally prefer the approach here. The code must not fail if there are no rules. Thus the default is allow but to change it to deny by default you add a policy that denies if no other rules are triggered (currently not straight forward but with the new opa policies design it is).

I would be more comfortable if the default would come from a file, either part of the git repo or a configmap, rather than compiled in the code.

rohithdv:
> > > Is there a way to have the decision not in the code?
> > 
> > 
> > I personally prefer the approach here. The code must not fail if there are no rules. Thus the default is allow but to change it to deny by default you add a policy that denies if no other rules are triggered (currently not straight forward but with the new opa policies design it is).
> 
> I would be more comfortable if the default would come from a file, either part of the git repo or a configmap, rather than compiled in the code.

To enable default policy outside the opa connector (i.e., from a rego file), I was just experimenting with the below approach and I think the following solution may work. 

(1) Load the default_policy.rego (attached below) along with opa during opa installation phase. This is the default policy file which enables any default behaviour (say deny by default or allow by default) by using a "helper" variable called "enfaction"( short for enforcement action). The default policy file also takes care of any precedence conditions that we may define (for eg: deny > transform > allow). This is also shown as an example in the file. 

[defaultpolicy.rego.txt](https://github.com/IBM/the-mesh-for-data/files/6286280/defaultpolicy.rego.txt)

(2) Change the opa connector code (one time change only) to query the "enfaction" target (which is specified in the default_policy.rego) and process the opa response accordingly. 

(3) The users may load any further policies at their will (for example the policies given in user-created-policy-1 or user-created-policy-2 folders or any similar folder) and the corresponding policies will be enforced.  
 
Please let me know your feedback if this approach is ok. Thanks. @ronenkat @roee88.

roee88:
@rohithdv your technical approach is similar to what I described (and did in the katalog PoC). But @ronenkat I'm sorry this is not a discussion for PR. I suggest to move along with the current *fix* (it's a bug being fixed) and move this discussion to the OPA+Taxonomy discussions. We need a fix before the release and the current one is legitimate IMO.

connectors/opa/lib/evaluate_opa.go
```diff
@@ -75,5 +89,15 @@ func EvaluatePoliciesOnInput(inputMap map[string]interface{}, opaServerURL strin
 	log.Println("responsestring data")
 	log.Println(string(data))
 
-	return string(data), nil
+	currentData := string(data)
+	decisionid, flag := doesOpaHaveUserPoliciesLoaded(string(data))
```
ronenkat, line 93: the json parser expects []byte, why cast to string and back to []byte?

ronenkat (commented):
there are fmt.printf and log.println in the code. Should we have both?
are the fmt statements for debugging? and they should be removed?

connectors/opa/lib/evaluate_opa.go
rohithdv: Fixed. Please check. Thanks.

rohithdv:
> there are fmt.printf and log.println in the code. Should we have both?
> are the fmt statements for debugging? and they should be removed?

Now using on log package for printing debug information. Removed fmt package for consistency. Thanks.

shlomitk1 (approved):
Verified.

PR #502, opened 2021-04-09
Filter cluster-scoped resources for Operate First 

eletonia:
This PR adds a script to filter out the cluster-scoped resources in the large yaml files generated by `helm template .` in the `charts/m4d` and `charts/m4d-crd` directories. This is required to deploy these cluster-scoped resources on Operate First.

The `filterClusterScoped.py` script takes `m4d.yaml` and `m4d-crd.yaml` as inputs and creates separate yaml files for cluster-scoped resources in the `samples/operate-first` directory.


roee88:
Quick comment before reviewing, you must set release name and namespace when execution helm template. Otherwise the result YAMLs have invalid names (like the NetworkPolicy name).

roee88:
I've read https://github.com/operate-first/apps/blob/master/cluster-scope/README.md#application-folder-structure and have some comments:
1. Generate the initial yaml with `helm template m4d -n m4d-system charts/m4d` (similar for the crds chart)
2. I think that you missed some cluster scoped resources unless I captured it wrong. You may use `yq` to select cluster scoped only the initial yaml. For example: `helm template m4d -n m4d-system charts/m4d | ./hack/tooks/bin/yq e 'select(.kind=="Namespace" or .kind=="MutatingWebhookConfiguration" or .kind=="ValidatingWebhookConfiguration" or .kind=="ClusterRole" or .kind=="ClusterRoleBinding")' -`
3. When splitting to multiple files write to a folder named as the plural kind name (e.g., clusterrolebindings)
4. When splitting to multiple files write to a file named as the `.metadata.name` from the resource instead of parsing the comment
5. In this or a future PR we can setup a github action that can be triggered manually to send a PR to operate-first repo. We would trigger this PR in M4D tagged releases only. I suggest to wait for a tag release anyway.

eletonia:
Thanks for reviewing @roee88 

I generated the initial yaml with `helm template m4d -n m4d-system charts/m4d | ./hack/tooks/bin/yq e 'select(.kind=="Namespace" or .kind=="MutatingWebhookConfiguration" or .kind=="ValidatingWebhookConfiguration" or .kind=="ClusterRole" or .kind=="ClusterRoleBinding")' -` and now [the script](https://github.com/eletonia/the-mesh-for-data/blob/master/samples/operate-first/filterClusterScoped.py) groups the yamls into folders.

roee88 (approved):
Awesome. Can you work on the following as followup PRs? I'm trying to get to a point where we can send a PR to operate-first/apps on each release automatically:
- Add Makefile with the commands to generate the initial YAML and to run the script
- Add `requirements.txt` for the python dependencies 
- Add `m4d-system` or whatever we need to add below the type folders. For example `clusterrolebindings/m4d-system/...`
- Generate `kustomization.yaml` files in the python script

Issue #501, opened 2021-04-08
Dependencies: bump controller-tools


roee88:
The version of controller-tools that we use today generates deprecated resources (CRDs, webhooks). Bumping to 0.5.0 (currently latest) will solve this while still preserving compatibility with Kubernetes 1.16+.

froesef:
I tried to upgrade controller-tools and many different other libraries. In the end I can generate the v1 CRDs fine but when the status of the plotter is updated (`r.Status().Update(ctx, &plotter)`) somehow parts of the status vanish (the map that contains the status of cluster name to blueprint metadata contains the entry for the cluster but the MetaBlueprint is completely empty. Thus there seem to be problems with maps in the status.

roee88:
Do you have a branch for that in your fork?

PR #504, opened 2021-04-11
change: mover image location

### LINKED ISSUE #500: Charts: decouple mover image parameter from global config
The mover image is versioned independently of m4d. The `mover.image` Helm parameter needs to be fully qualified. Must also be set correctly in tests to point to the c
"dummy" mover.

roee88:
* charts: mover image is fully qualified
* charts: updated comments for mover fields
* mover image changed to ghcr.io/mesh-for-data/mover:latest
* default values in modules do not specify image

@froesef I couldn't find where/why to change anything related to dummy mover. Also, is there a specific tag I should use?

Fix #500

### DIFF
--- a/modules/m4d-implicit-copy-batch/values.yaml
+++ b/modules/m4d-implicit-copy-batch/values.yaml
@@ -12,7 +12,7 @@
 # app_cluster:
 labels: {}
 
-image: "ghcr.io/the-mesh-for-data/mover:latest"
+image: ""
 imagePullPolicy: null
 noFinalizer: "true"
 
--- a/charts/m4d/templates/manager-deployment.yaml
+++ b/charts/m4d/templates/manager-deployment.yaml
@@ -86,7 +86,7 @@ spec:
             - name: ENABLE_WEBHOOKS
               value: "true"
             - name: MOVER_IMAGE
-              value: {{ include "m4d.image" ( tuple $ .Values.mover ) }}
+              value: {{ .Values.mover.image }}
             - name: IMAGE_PULL_POLICY
               value: {{ .Values.mover.imagePullPolicy | default .Values.global.imagePullPolicy }}
             {{- if .Values.manager.extraEnvs }}
--- a/charts/m4d/values.yaml
+++ b/charts/m4d/values.yaml
@@ -107,9 +107,11 @@ worker:
 
 # Mover configuration used in the Mostion API controllers
 mover:
-  # Image name or a hub/image[:tag]
-  image: "mover"
-  # Overrides global.imagePullPolicy
+  # Default image for mover (format: hub/image[:tag])
+  image: "ghcr.io/mesh-for-data/mover:latest"
+  
+  # Default image pull policy for mover.
+  # Overrides global.imagePullPolicy.
   imagePullPolicy: ""
 
 # Manager component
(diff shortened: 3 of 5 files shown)


modules/m4d-implicit-copy-batch/values.yaml
```diff
@@ -12,7 +12,7 @@
 # app_cluster:
 labels: {}
 
-image: "ghcr.io/the-mesh-for-data/mover:latest"
+image: ""
```
froesef, line 15: I think this is a good change. That way the default image is the one that is configured in the MOVER_IMAGE environment variable. Alternatively one can always overwrite this value in the module definition.

shlomitk1: If I understand correctly, the default behavior is that a webhook defaults image to MOVER_IMAGE. What happens when webhooks are not running, e.g. in tests?

roee88: Good catch. 

What would be the preference:
1. Omit `mover` section from m4d chart so there will be no default MOVER_IMAGE. Instead, the `image` value in the module `values.yaml` is the default.
2. Default also in reconcile loop. This ensures that the functional behavior is the same even if no webhooks are running (there is still a benefit in running the webhooks because it will block invalid input on submission and avoid changing the spec state to be invalid). In previous projects I used this code as part of the reconcile function:
```go
	// Validate the resource
	switch instance.(type) {
	case webhook.Validator:
		if err := instance.(webhook.Validator).ValidateCreate(); err != nil {
			reqLogger.Error(err, "resource is invalid")
			return r.ManageError(instance, err)
		}
	}

	// Set default values for the resource field
	switch instance.(type) {
	case webhook.Defaulter:
		beforeDefaults := instance.DeepCopyObject()
		instance.(webhook.Defaulter).Default()
		if !equality.Semantic.DeepEqual(beforeDefaults, instance) {
			reqLogger.Info("updating resource after setting defaults")
			if err := r.Client().Update(context.TODO(), instance); err != nil {
				return r.ManageError(instance, err)
			}
			return ctrl.Result{Requeue: true}, nil
		}
	}
```

shlomitk1: I like the first approach.

roee88: The first approach would require setting the image in each BatchTransfer/StreamTransfer when using directly and not as a module. Do we care? @froesef it's your call what to do.

froesef: I agree with removing the `MOVER_IMAGE` parameter. It's a relic from the times when the motion components were separate and it was easier for testing. What we could also do is define a default in the CRD definition (controller-gen annotations). Should I do a PR for this?

roee88: > What we could also do is define a default in the CRD definition (controller-gen annotations).

You can't. Defaulting requires apiextensions.k8s.io/v1 so #501 must be fixed first. 
Defaulting is also not enabled by default on Kubernetes 1.16 (which is the minimal version we support).
I just had the same discussion with Tomer.

roee88: With that in mind feel free to prepare a separate PR or take over this one.

roee88:
To be replaced by a PR that removes the MOVER_IMAGE paramter

PR #512, opened 2021-04-11
chore: remove third_party/registry

### LINKED ISSUE #511: chore: remove third_party/registry
third_party/registry creates an Image Pull Secret used by the "default" service account in m4d-system to load images from a private registry. It is called from the `cluster-prepare` make target.

Problem:
1. The components deployed by the chart use unique service accounts (e.g., in manager it defaults to "manager") and it's configurable by the user
2. We don't use private images in our codebase. If one wants to use a private image he should create the secret on his own. Passing it to the installation can be done by passing the `imagePullSecrets` to the helm install. 
3. Our CI calls it as part of `cluster-prepare` but it shouldn't (normal installation don't have it)

roee88:
Fix #511


shlomitk1 (approved):
Looks good to me.

PR #522, opened 2021-04-14
Dataset Credentials related changes in Egeria

rohithdv:
This PR addresses changes related to  passing dataset credentials in Egeria. 

The following changes are done: 

(1) Add a field called credentials-secret-ref to fullPath (https://github.com/IBM/the-mesh-for-data/blob/5d8bb22d58ebf99ea9ac1ef7ee9d61fe9792ede4/third_party/egeria/usage/example_transactions.csv.json#L5 ):
The field value should be of the form: <secret namespace>/<secret name> for example:
 "fullPath": "{\"data_store\":\"S3\", \"bucket\":\"mybucket\", \"endpoint\":\"s3.eu-de.cloud-object-storage.appdomain.cloud\", \"object_key\":\"transactions.csv\", \"region\":\"eu\", \"data_location\": \"US\",  \"credentials-secret-ref\": \"default/my-secret\"}",
 
2. VaultSecretPath in CredentialsInfo field(https://github.com/IBM/the-mesh-for-data/blob/master/pkg/connectors/protos/dataset_details.proto#L78) should be of the form in Egeria connector:
"/v1/kubernetes-secrets/<secret name>?namespace=<namespace name>
", for example:
"/v1/kubernetes-secrets/my-secret?namespace=default"
 
For reference, Please see:
https://github.com/IBM/the-mesh-for-data/blob/master/connectors/katalog/pkg/connector/catalog.go#L56
Which uses PathForReadingKubeSecret function to construct the path to return:
https://github.com/IBM/the-mesh-for-data/blob/582ed5581d507c5429e250f836b5ef072a9af607/pkg/vault/vault_paths.go#L24 (edited) 


third_party/egeria/usage/example_transactions.csv.json
```diff
@@ -2,7 +2,7 @@
     "class": "NewCSVFileAssetRequestBody",
     "displayName": "transactions.csv",
     "description": "Synthetic Financial Dataset For Fraud Detection",
-    "fullPath": "{\"data_store\":\"S3\", \"bucket\":\"mybucket\", \"endpoint\":\"s3.eu-de.cloud-object-storage.appdomain.cloud\", \"object_key\":\"transactions.csv\", \"region\":\"eu\", \"data_location\": \"US\"}",
+    "fullPath": "{\"data_store\":\"S3\", \"bucket\":\"mybucket\", \"endpoint\":\"s3.eu-de.cloud-object-storage.appdomain.cloud\", \"object_key\":\"transactions.csv\", \"region\":\"eu\", \"data_location\": \"US\", \"credentials-secret-ref\": 'default/my-secret'}",
```
revit13, line 5: hi @rohithdv  when posting the example_transactions.csv.json to egeria I get error back related to the format of the new field. To simplify things I suggest to change it as follows and it worked ok:

"fullPath": "{\"data_store\":\"S3\", \"bucket\":\"notebook-sample\", \"endpoint\":\"s3.eu-gb.cloud-object-storage.appdomain.cloud\", \"object_key\":\"data.csv\", \"region\":\"eu\", \"data_location\": \"Turkey\",                       \"dataset_creds_secret_name\": \"dataset-creds\", \"dataset_creds_secret_namespace\": \"default\"}"

I appreciate if you could also change the egeria connector code to use dataset_creds_secret_name and  dataset_creds_secret_namespace fields. Thanks

shlomitk1 (commented):
What happens if the json file does not include this field? Until it is checked by taxonomy, can an error be returned?
@ronenkat @roee88 @simanadler - FYI

revit13 (approved):

third_party/egeria/usage/example_transactions.csv.json
```diff
@@ -2,7 +2,7 @@
...
     "columnHeaders": [
```
revit13, line 6: @rohithdv- a minor comment (not a blocker for merging)- I just noticed that all other fields contain underscore in their key names so you could consider changing the new field to be in the same format.

roee88:
> What happens if the json file does not include this field? Until it is checked by taxonomy, can an error be returned?
> @ronenkat @roee88 @simanadler - FYI

Isn't it the same as any other field in fullPath?

third_party/egeria/usage/example_transactions.csv.json
rohithdv: Fixed. Please check. Thanks.

shlomitk1:
> 
> 
> > What happens if the json file does not include this field? Until it is checked by taxonomy, can an error be returned?
> > @ronenkat @roee88 @simanadler - FYI
> 
> Isn't it the same as any other field in fullPath?

For other fields, an empty string is returned if no field specified. In this case an invalid path is constructed.
Actually, I think it will crash on non-existent credentials field in 

> 
> 
> > What happens if the json file does not include this field? Until it is checked by taxonomy, can an error be returned?
> > @ronenkat @roee88 @simanadler - FYI
> 
> Isn't it the same as any other field in fullPath?

No, instead of returning an empty value, it will crash on `String secretName = credSecretRef.split("/")[1];` or return an invalid path.

third_party/egeria/usage/example_transactions.csv.json
revit13: Thanks it works OK.

shlomitk1 (changes requested):
Regarding these lines:
        String secretNamespace = credSecretRef.split("/")[0];
        String secretName = credSecretRef.split("/")[1];
Please add a check that the field exists (would json return nil otherwise?) and a check that 
 credSecretRef.split("/") is an array of size 2.

rohithdv:
> Regarding these lines:
> String secretNamespace = credSecretRef.split("/")[0];
> String secretName = credSecretRef.split("/")[1];
> Please add a check that the field exists (would json return nil otherwise?) and a check that
> credSecretRef.split("/") is an array of size 2.

Fixed. Please check. Thanks. @shlomitk1 @revit13

shlomitk1 (approved):

roee88 (commented):
1. Why all the `replaceAll("[\r\n]", "")`? 
2. Don't you want to error if `credentials_secret_ref` exists but is not in a valid format?

rohithdv:
> * Why all the `replaceAll("[\r\n]", "")`?
> * Don't you want to error if `credentials_secret_ref` exists but is not in a valid format?

For the first comment, we get an error when we build egeria connector with java linters on (mvn -B clean install -DskipTests). The error from spot bugs (one of the linters we use during building stage) is given in the following html document (given as a zipped file). Please check "Security Warnings" section and "Potential CRLF Injection for logs" subsection in the below html document. 


[spotbugsXml.html.zip](https://github.com/IBM/the-mesh-for-data/files/6377480/spotbugsXml.html.zip)

For the second comment, I have made the changes in the [latest commit] (https://github.com/IBM/the-mesh-for-data/pull/522/commits/a51ed66fdc03843dc07cb94b198ca9e05bf22241).

Please let me your comments. Thanks. 
@roee88

PR #524, opened 2021-04-15
Remove mover image config from m4d-config

### LINKED ISSUE #500: Charts: decouple mover image parameter from global config

froesef:
This PR fixes #500 and make the standard image configurable in the m4d module and not any more as a global configuration in the m4d-config configmap.

### DIFF
--- a/charts/m4d/templates/manager-deployment.yaml
+++ b/charts/m4d/templates/manager-deployment.yaml
@@ -85,10 +85,6 @@ spec:
           env:
             - name: ENABLE_WEBHOOKS
               value: "true"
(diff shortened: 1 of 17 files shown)


roee88 (approved):

roee88:
LGTM. Consider setting "ghcr.io/mesh-for-data/mover:latest" as a common const in the code (though it might be removed with controller-tools 0.5.0 so not that important).

shlomitk1 (approved):

PR #456, opened 2021-03-22
[RFC] website: updated quickstart documentation

roee88:
Preview in https://607d43241d8fadb03c5fa9b0--mesh-for-data.netlify.app/get-started/quickstart/

PR #525, opened 2021-04-16
Change the quick start guide missing dependencies

rachit1arora:


roee88:
Hi,

Thank you for the PR. 

We're changing the quickstart in https://github.com/IBM/the-mesh-for-data/pull/456 and the new version does not require these dependencies (it basically depends just on helm3 and kubectl). I suggest to wait for it to be merged. 

However, the dependencies list in the [Development docs](https://github.com/IBM/the-mesh-for-data/blob/master/site/docs/contribute/environment.md#dependencies) does not include all of the dependencies you added. Specifically nc and gcc. Would you mind submitting a PR to update it? Also, can you share where gcc was required?

rachit1arora:
@roee88 : sure, I will raise a PR to updated location 

gcc was required for one of the go dependencies 

```go: downloading github.com/docker/go-metrics v0.0.0-20181218153428-b84716841b82
# runtime/cgo
cgo: exec gcc: exec: "gcc": executable file not found in $PATH

```

Issue #523, opened 2021-04-14
Send a PR to operate-first/apps on each release automatically


eletonia:
Dependent on #600 
This issue lists initial steps to get to a point where we can automatically send a PR to https://github.com/operate-first/apps so the latest Fybrik resources will be deployed to the Operate First cluster. @roee88 @ronenkat 
- [x] Add Makefile with the commands to generate the initial YAML and to run the `filterClusterScoped` script
- [x] Add `requirements.txt` for the python dependencies
- [x] Add `fybrik-system` or whatever we need to add below the type folders. For example `clusterrolebindings/fybrik-system/...`
- [x] Generate `kustomization.yaml` files in the python script
- [ ] Add a GitHub workflow go automate the process upto raising a PR

roee88:
An example workflow that does something similar is https://github.com/IBM/the-mesh-for-data/blob/master/.github/workflows/chart-publish.yaml. 

The changes from it would be:
- instead of the charts repo checkout the operate-first/apps repo
- replace the run commands with the make commands you added and then copy the results info the right folder within the cloned apps dir
- don't use tibdex/github-app-token@v1 but rather use the Fork method of peter-evans/create-pull-request@v3 as described in https://github.com/peter-evans/create-pull-request/blob/master/docs/concepts-guidelines.md#push-pull-request-branches-to-a-fork

I suggest to try this on your own test repositories the mimick ibm/the-mesh-for-data and operate-first/apps.

roee88:
As described in https://github.com/fybrik/fybrik/pull/816#issuecomment-988710351. Still missing:

- [ ] Avoid having copies of resources in https://github.com/fybrik/fybrik/tree/master/samples/operate-first
- [ ] Submit PR to the upstream apps repo, not to our fork. The use of the fork is intended for a place for creating a branch to submit the PR from. See https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#push-pull-request-branches-to-a-fork and the push-to-fork option.
- [ ] Unless we can completely remove `samples/operate-first` we should at least move it to somewhere else because it's not a sample.

eletonia:
@roytman Where should we move [samples/operate-first](https://github.com/fybrik/fybrik/tree/master/samples/operate-first) to in our repository? Also where should documentation for Operate First deployment go?

roytman:
Hi @eletonia , please put the `operate-first` files under the `integration/operate-first` dir in our repository. 
[...]

PR #538, opened 2021-04-20
Operate First - cluster scoped resources 

eletonia:
This PR implements the first 4 tasks in #523:

- Add Makefile with the commands to generate the initial YAML and to run the filterClusterScoped script
-  Add requirements.txt for the python dependencies
-  Add m4d-system or whatever we need to add below the type folders. For example clusterrolebindings/m4d-system/...
-  Generate kustomization.yaml files in the python script


samples/operate-first/filterClusterScoped.py
```diff
@@ -10,16 +10,40 @@ def splitToYamls(yamlFile):
...
+    dirs = [name for name in os.listdir(".") if os.path.isdir(name)]
+    for subdir in dirs:
+        outputPath = os.path.join(subdir,'m4d-system','kustomization.yaml')
+        f = open(outputPath,'w')
+        kustomizationString = "---\napiVersion: kustomize.config.k8s.io/v1beta1\nkind: Kustomization" + "\n\nresources:\n"
+        subdirPath = os.path.join(".", subdir,'m4d-system')
+        for root, subdirs, files in os.walk(subdirPath):
+            for file in files:
+                if file != 'kustomization.yaml':
+                    kustomizationString += "\t- " + file +'\n'
```
roee88, line 30: Replace tab with two spaces

samples/operate-first/Makefile
```diff
@@ -0,0 +1,11 @@
+export RELEASE_NAME=m4d
+export NAMESPACE=m4d-system
+export M4D_CHART_PATH=../../charts/m4d
```
roee88, line 3: Add the following to the top of the Makefile
```
ROOT_DIR := ../..
```

Then use `$(ROOT_DIR)` in this line and the next one

samples/operate-first/Makefile
```diff
@@ -0,0 +1,11 @@
...
+export CRD_CHART_PATH=../../charts/m4d-crd
+
+.PHONY: helm-template
+helm-template: 
+	helm template $(RELEASE_NAME) -n $(NAMESPACE) $(M4D_CHART_PATH) | ../../hack/tools/bin/yq e 'select(.kind=="Namespace" or .kind=="MutatingWebhookConfiguration" or .kind=="ValidatingWebhookConfiguration" or .kind=="ClusterRole" or .kind=="ClusterRoleBinding")' - > m4d.yaml
```
roee88, line 8: Use `$(TOOLBIN)/yq` instead of `../../hack/tools/bin/yq`

samples/operate-first/Makefile
```diff
@@ -0,0 +1,11 @@
...
+	helm template $(RELEASE_NAME) -n $(NAMESPACE) $(CRD_CHART_PATH) > m4d-crd.yaml
```
roee88, line 9: Use `$(TOOLBIN)/helm` instead of helm

samples/operate-first/Makefile
roee88, line 8: Use `$(TOOLBIN)/helm` instead of helm

samples/operate-first/Makefile
```diff
@@ -0,0 +1,11 @@
+export RELEASE_NAME=m4d
+export NAMESPACE=m4d-system
+export M4D_CHART_PATH=../../charts/m4d
+export CRD_CHART_PATH=../../charts/m4d-crd
+
+.PHONY: helm-template
+helm-template:
```
roee88, line 7: Replace with:
```suggestion
helm-template: $(TOOLBIN)/yq $(TOOLBIN)/helm
```

samples/operate-first/filterClusterScoped.py
```diff
@@ -10,16 +10,40 @@ def splitToYamls(yamlFile):
             if yml:
                 filename = yml['metadata']['name']
                 outputFolder = yml['kind'].lower()+'s'
-                outputPath = os.path.join(outputFolder, filename + ".yaml")
+                outputPath = os.path.join(outputFolder, 'm4d-system', filename + ".yaml")
                 if not os.path.exists(os.path.dirname(outputPath)):
                     os.mkdir(os.path.dirname(outputPath))
```
roee88, line 15: Since this is now a nested folder please change to use os.makedirs with exit_ok=True instead of using mkdir. You then also don't need the os.path.exists check

samples/operate-first/filterClusterScoped.py
```diff
@@ -10,16 +10,40 @@ def splitToYamls(yamlFile):
...
                 f = open(outputPath, "w") 
                 documents = yaml.dump(yml, f)
-                f.close()          
+                f.close()
```
roee88, line 18: I suggest using `with` blocks everywhere you work on a file

samples/operate-first/Makefile
```diff
@@ -0,0 +1,11 @@
+export RELEASE_NAME=m4d
+export NAMESPACE=m4d-system
+export M4D_CHART_PATH=../../charts/m4d
+export CRD_CHART_PATH=../../charts/m4d-crd
+
+.PHONY: helm-template
+helm-template: 
+	helm template $(RELEASE_NAME) -n $(NAMESPACE) $(M4D_CHART_PATH) | ../../hack/tools/bin/yq e 'select(.kind=="Namespace" or .kind=="MutatingWebhookConfiguration" or .kind=="ValidatingWebhookConfiguration" or .kind=="ClusterRole" or .kind=="ClusterRoleBinding")' - > m4d.yaml
+	helm template $(RELEASE_NAME) -n $(NAMESPACE) $(CRD_CHART_PATH) > m4d-crd.yaml
+	python3 -m pip install -r requirements.txt
```
roee88, line 10: Please move the two python calls to a different helm target like `generate`.

roee88: The use of python3 binary s environment specific. I usually setup a virtual env and avoid messing with the system python. In all virtual envs and in github actions the binary is just called python (even for 3.x). The header in your python script is also environment specific. My suggestion:
1. Change the header in the python script to `#!/usr/bin/env python` (we don't really use it in the Makefile but just for completion)
2. Change python3 to python
3. On your own machine setup a virtual env when you run it. In the next PR when we create a GitHub action it won't be relevant.

eletonia:
@roee88 Thanks for reviewing! I made the changes

samples/operate-first/Makefile
```diff
@@ -0,0 +1,20 @@
...
+include $(ROOT_DIR)/Makefile.env
+export RELEASE_NAME=m4d
+export NAMESPACE=m4d-system
+export M4D_CHART_PATH=$(ROOT_DIR)/charts/m4d
+export CRD_CHART_PATH=$(ROOT_DIR)/charts/m4d-crd
+
+.PHONY: helm-template
+helm-template: $(TOOLBIN)/yq $(TOOLBIN)/helm
+	$(TOOLBIN)/helm template $(RELEASE_NAME) -n $(NAMESPACE) $(M4D_CHART_PATH) | $(TOOLBIN)/yq e 'select(.kind=="Namespace" or .kind=="MutatingWebhookConfiguration" or .kind=="ValidatingWebhookConfiguration" or .kind=="ClusterRole" or .kind=="ClusterRoleBinding")' - > m4d.yaml
+	$(TOOLBIN)/helm template $(RELEASE_NAME) -n $(NAMESPACE) $(CRD_CHART_PATH) > m4d-crd.yaml
```
roee88, line 11: using the same release name will work because we don't use it in the crd templates, but if we ever change that then it would be better to use a unique release name like `m4d-crd`

samples/operate-first/Makefile
```diff
@@ -0,0 +1,20 @@
...
+	python -m pip install -r requirements.txt
+	python filterClusterScoped.py
```
roee88, line 13: please remove this line, it's already part of generate

samples/operate-first/Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+.PHONY: generate
+generate:
+	python -m pip install -r requirements.txt
+	python filterClusterScoped.py
+
+all: helm-crd-hook helm-verify
```
roee88, line 20: 1. Change `all` to be the first target in this Makefile (i.e. move it to before the `helm-template` target) so it would be the default target
2. You probably meant `all: helm-template generate` because `helm-crd-hook` and `helm-verify` do not exist here
3. add `.PHONY: all` like in the other targets

samples/operate-first/requirements.txt
```diff
@@ -0,0 +1 @@
+pyyaml
```
roee88, line 1: Please fixate it to version `5.4.1` because earlier versions had a security issue (we just dealt with that in the flight module project)

samples/operate-first/Makefile
eletonia: Fixed!

samples/operate-first/requirements.txt
eletonia: Done!

samples/operate-first/Makefile
```diff
@@ -0,0 +1,20 @@
...
+	$(TOOLBIN)/helm template $(RELEASE_NAME) -n $(NAMESPACE) $(M4D_CHART_PATH) | $(TOOLBIN)/yq e 'select(.kind=="Namespace" or .kind=="MutatingWebhookConfiguration" or .kind=="ValidatingWebhookConfiguration" or .kind=="ClusterRole" or .kind=="ClusterRoleBinding")' - > m4d.yaml
+	$(TOOLBIN)/helm template $(CRD_RELEASE_NAME) -n $(NAMESPACE) $(CRD_CHART_PATH) > m4d-crd.yaml
+
+.PHONY: generate
+generate:
+	python -m pip install -r requirements.txt
+	python filterClusterScoped.py
+
+.PHONY: all
+all: helm-template generate
```
roee88, line 20: minor: I meant moving these two lines to before line 9. If `all` is the first target then simply running `make` will invoke it. So it's a best practice to make `all` the first target.

roee88 (approved):
There is one minor comment left. Let me know if you want to address it.

samples/operate-first/Makefile
eletonia: Ah okay, I moved it!

Issue #225, opened 2020-12-22
Develop a mechanism to pass cluster-dependent arguments during a module orchestration 


shlomitk1:
**Problem statement:** 
Manager selects a module based on M4DModule set running on a coordinator cluster. The module arguments include some values that are cluster dependent, such as URL of some instance to connect to in a selected cluster, thus can not be configured in a M4DModule directly. 
A possible solution would be to apply a local instance of the module in a remote cluster that will include the specific values, and use the local instance to override the values during the orchestration.

shlomitk1:
Requirements:
We want to propagate some of the decisions to the local cluster level (Blueprint, Blueprint controller running in the remote cluster). 
Another requirement is to enable an easier module development when there is no forced connection between module arguments and the manager knowledge of the module (we do not want to change module arguments structure).
Proposed solution:
- M4DModule is deployed only in coordinator cluster.
- Optionally, spec of the module will allow attach environment similar to the way it works with pods.
```
spec:
...
  env:
    - name:
       value:
   envfrom:
      - configMapRef:
             name:
       - secretRef:
             name:
```
The first part (env) is used for variable defaults - add a variable with its default value for all possible deployments. Names/Values are propagated through blueprint as an addition to module arguments. It is optional.
The second part (envfrom) provides instructions to the BlueprintController where to look for cluster-specific values. Configmap/secret names are propagated to the BlueprintController, the values take precedence over all module arguments including env part. Optional as well.
Example: a module receives jobNumber, url, username and password arguments.
M4DModule spec will look like:
```
env:
  - name: jobNumber
  - value: 1
envfrom:
   - configMapRef:
      name: module-config
    - secretRef:
      name: module-secret
```
      
A local cluster will deploy a module-config configmap with jobNumber and url variables, module-secret with username and password. If the configmap does not include jobNumber, the default value 1 will be taken.

revit13:
@shlomitk1 how does the BlueprintController will get the instructions on where to look for cluster-specific values? Will the Blueprint contain it? Your suggestion seems to answers the requirements as far as I understand them. Unfortunately, I am not familiar with all Kubernetes features so I do not have an alternative approach for a solution if such exists.

shlomitk1:
> 
> 
[...]

PR #539, opened 2021-04-21
Overriding module arguments

shlomitk1:

Fixes https://github.com/IBM/the-mesh-for-data/issues/225

Description: For each module instance specified in the Blueprint, the Blueprint controller fetches the corresponding module definition (if the module is deployed in that cluster) and overrides value using Chart.Values from the module definition.


roee88:
So this require applying `M4DModule` in the remote cluster and include **all** fields of it? because fields like `chart.name`, `capabilities`, etc. are required in `M4DModule`.

shlomitk1:
> 
> 
> So this require applying `M4DModule` in the remote cluster and include **all** fields of it? because fields like `chart.name`, `capabilities`, etc. are required in `M4DModule`.

Yes. The idea was that two almost identical yamls will be applied in different clusters with the only difference in a cluster specific value.

PR #541, opened 2021-04-21
Extend implicit copy module

### LINKED ISSUE #540: Add csv to csv copy support in implicit-copy
Currently the implicit copy module only copies from X to parquet. 

The mover also has the possibility to copy from and to other formats. The goal is to extend the default copy module 
to include some other commonly used formats that are supported by the mover.

froesef:
Fixes #540


roee88 (commented):
Good improvement. However, what I meant in our last call is to have a mock that is independent on the application controller and its types and interfaces. This way the mock would be shared across the project in a clean interface and without issues of circular dependencies. You don't really need to invent a new interface but rather reuse the existing one created by proto (without any grpc usage, just plain go interface). The datadetails structure and friends are things that I consider as internal implementation details of the application controller and I can easily see a near future where they don't exist. Do you want to give it a try? Or shall we merge as is and discuss it later? (Maybe I can do a follow-up PR to showcase what I mean, if time allows). I'm good either way.

Minor: we can have a generic function that returns a resource from YAML using the universal decoder.

roee88:
Actually the [original mock code](https://github.com/IBM/the-mesh-for-data/blob/b1f74c3f2cd3682f1bc3cad210333413a56d2b8a/manager/controllers/mockup/mockup_catalog_connector.go#L24)  is closer to what I meant. 

_Warning: untested code ahead_

Consider the following interface:
```go
// DataCatalog is an interface of a facade to a data catalog.
type DataCatalog interface {
	pb.DataCatalogServiceServer
	io.Closer
}
```

A real implementation of it using a grpc client:

```go
// Ensure that grpcDataCatalog implements the DataCatalog interface
var _ DataCatalog = (*grpcDataCatalog)(nil)

type grpcDataCatalog struct {
	name              string
	connectionURL     string
	connectionTimeout time.Duration
	connection *grpc.ClientConn
	client     pb.PolicyManagerServiceClient
}

// NewGrpcDataCatalog creates a DataCatalog facade that connects to a GRPC service
// You must call .Close() when you are done using the created instance
func NewGrpcDataCatalog(name string, connectionURL string, connectionTimeout time.Duration) (DataCatalog, error) {
	ctx, cancel := context.WithTimeout(context.Background(), connectionTimeout)
	defer cancel()
	connection, err := grpc.DialContext(ctx, connectionURL, grpc.WithInsecure(), grpc.WithBlock())
	if err != nil {
		return nil, errors.Wrap(err, fmt.Sprintf("NewGrpcDataCatalog failed when connecting to %s", connectionURL))
	}
	return &grpcDataCatalog {
		name:              name,
		connectionURL:     connectionURL,
		connectionTimeout: connectionTimeout,
		client:            pb.NewDataCatalogServiceClient(connection),
		connection:        connection,
	}, nil
}

func (m *grpcDataCatalog) GetDatasetInfo(ctx context.Context, in *pb.CatalogDatasetRequest) (*pb.CatalogDatasetInfo, error) {
	result, err := m.client.GetDatasetInfo(ctx, in)
	return result, errors.Wrap(err, fmt.Sprintf("get dataset info from %s failed", m.name))
}

func (m *grpcDataCatalog) RegisterDatasetInfo(context.Context, *pb.RegisterAssetRequest) (*pb.RegisterAssetResponse, error) {
	result, err := m.client.RegisterDatasetInfo(ctx, in)
	return result, errors.Wrap(err, fmt.Sprintf("register dataset info in %s failed", m.name))
}

func (m *grpcDataCatalog) Close() error {
	return m.connection.Close()
}
```

A mock implementation of it used in unit tests:

```go
// Ensure that grpcDataCatalog implements the DataCatalog interface
var _ DataCatalog = (*MockDataCatalog)(nil)

type MockDataCatalog struct {
    pb.UnimplementedDataCatalogServiceServer
}

func (m *grpcDataCatalog) GetDatasetInfo(ctx context.Context, in *pb.CatalogDatasetRequest) (*pb.CatalogDatasetInfo, error) {
    // A mocked implementation like in the original code at
    // https://github.com/IBM/the-mesh-for-data/blob/b1f74c3f2cd3682f1bc3cad210333413a56d2b8a/manager/controllers/mockup/mockup_catalog_connector.go#L24 
    // or whatever you want (e.g., adding registration of matchers and mocked responses, etc.)
}

func (m *MockDataCatalog ) Close() error {
	return nil
}
```

Then the controller gets a `DataCatalog` to use in its initialization, using the same pattern as in your PR here. Opinions?

roee88 (approved):

Issue #537, opened 2021-04-20
Fybrik Application Taxonomy


eletonia:
Task list to create and use a Fybrik Application Taxonomy:
- [x] Finalize the values in Fybrik application JSON schema taxonomy
- [x] Validate schema using @roee88 's [schema validator](https://github.com/roee88/the-mesh-for-data/tree/taxonomy-tool/pkg/taxonomy) #533 
- [x] Move all taxonomy json files to `charts/files` to create config maps #652 
- [x] Write JSON schema taxonomy to config maps #652 
- [x]  Mount config maps to manager deployment as volumeMounts #652 
- [x] Remove smart comments of kubebuilder with hard-coded enums #645 
- [x] Read taxonomy from config map in Fybrik application webhook and validate Fybrik application #645 
- [x] [Check](https://github.com/mesh-for-data/mesh-for-data/blob/664cdbe977f082315194b1a6472661a7036aa1ed/manager/README.md#run-and-debug-locally) if taxonomy validation is working in webhook
- [x] Add validation code to Reconciler and execute if webhooks flag is disabled
- [x] Remove data API strings from the manager code @shlomitk1

shlomitk1:
@eletonia Additional change is to remove smart comments of kubebuilder with hard-coded enums.

PR #475, opened 2021-04-05
 Issue #238 - Additional Taxonomies

simanadler:
Part of Issue #238

Adds taxonomies for policymanager and modules, and relevant tests.
Moves taxonomies to top level config

Issue #533, opened 2021-04-19
Develop a tool to validate taxonomy against the spec


roee88:
What: A go tool to validate taxonomy YAMLs against https://github.com/IBM/the-mesh-for-data/blob/master/config/taxonomy/HOWTO_SPEC.md

roee88:
Blocked by https://github.com/getkin/kin-openapi/issues/344

roee88:
Done in #637

PR #543, opened 2021-04-23
M4D Application Taxonomy - values schemas

eletonia:
Dependent on the merge of #475 

This PR adds the M4D Values Taxonomy JSON schemas
It addresses 2 of the tasks in #537


config/taxonomy/application.values.schema.json
```diff
@@ -0,0 +1,84 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "$id": "application.values.schema.json",
+    "title": "Application Values Taxonomy",
+    "description": "Values contributed to by the mesh for data applications.",
+    "definitions": {
+        "intent": {"$ref": "policymanager.structs.schema.json#/definitions/intent"},
```
ronenkat, line 7: These are definitions not values, but the filename indicates that it should include only values.
The definitions should be part of another scheme file, probably the in the M4DModule scheme

config/taxonomy/application.values.schema.json
```diff
@@ -0,0 +1,84 @@
...
+        "role": {"$ref": "policymanager.structs.schema.json#/definitions/intent"},
+        "app_info": {
+            "type": "object",
+            "properties": {
+                "intent": { "$ref": "#/definitions/intent" },
```
ronenkat, line 12: this is double reference, why not directly reference policymanager.structs.schema.json# ?

config/taxonomy/application_taxonomy_test.go
```diff
@@ -0,0 +1,53 @@
...
+	interfaceBadFlight = "{\"interface\":{\"protocol\":\"m4d-arrow-flight\", \"data_format\":\"parquet\"}}"
+
+	// {"interface":{"protocol":"kafka", "data_format":"json"}}
+	interfaceGoodKafka = "{\"interface\":{\"protocol\":\"kafka\", \"data_format\":\"json\"}}"
+
+	// {"interface":{"protocol":"whatever", "data_format":"avro"}}
+	interfaceBadKafka = "{\"interface\":{\"protocol\":\"whatever\", \"data_format\":\"avro\"}}"
+)
+
+func TestApplicationTaxonomy(t *testing.T) {
```
ronenkat, line 41: Assuming we have #533 how useful are these tests? 
What are you testing here?

We should have a couple of run scenarios (schemes):
1. M4D with minimal verification, i.e., check only protocol and data_format (with values that are supported today), nothing else. (additional properties True for app_info)
3. M4D with additional check for intent and role accepting specific values as a sample scheme values. Here we should have values as example for intent and role in app_info.
We should proceed in parallel in both scenarios to show that they work.

config/taxonomy/application.values.schema.json
```diff
@@ -0,0 +1,84 @@
...
+        
+                        "enum":["arrow"] 
+                    }
+                }
+            },
+            {
+                "type": "object",
+                "properties": {
+                    "protocol": {
+                        "type":"string",
```
ronenkat, line 46: should not define the type here

config/taxonomy/application.values.schema.json
eletonia: Both structs and values schemas have definitions for example [here](https://github.com/IBM/the-mesh-for-data/blob/9928d662fc48085667f0d340fc2617bdf41d7fbe/config/taxonomy/module.values.schema.json#L6) in the M4D module values schema, we still use definitions. The term `definitions` just enables us to reuse these in other json schemas.

config/taxonomy/application.values.schema.json
eletonia: Fixed

config/taxonomy/application.values.schema.json
eletonia: Fixed

config/taxonomy/application_taxonomy_test.go
eletonia: That tool will only validate that the taxonomy defined in a `schema.json` file is in the correct format. The Go tests are to test specific scenarios for values that the user will provide in M4D application and validate that against the taxonomy defined. We are using the [ValidateTaxonomy function](https://github.com/IBM/the-mesh-for-data/blob/9928d662fc48085667f0d340fc2617bdf41d7fbe/config/taxonomy/validate.go#L16) that @simanadler implemented. The Go tests cover both the run scenarios that you listed.

ronenkat: Thank you. Would it make more sense to validate in the test with M4Dapplication as an input? the test that were defined in the starting point scheme are somewhat over-simplified.
For example, defining (i) m4dapplication w/o intent/role, (ii) m4dapplication with intent/role
and check if it passes validation?

config/taxonomy/application.values.schema.json
```diff
@@ -0,0 +1,73 @@
...
+    "$id": "application.values.schema.json",
+    "title": "Application Values Taxonomy",
+    "description": "Values contributed to by the mesh for data applications.",
+    "definitions": {
+        "intent": {"$ref": "policymanager.structs.schema.json#/definitions/intent"},
+        "role": {"$ref": "policymanager.structs.schema.json#/definitions/role"},
+        "app_info": {
+            "type": "object",
+            "properties": {
+                "intent": { "$ref": "policymanager.structs.schema.json#/definitions/intent"},
```
ronenkat, line 12: do you need intent in "definitions"? can you reference in line 12 directly policymanager.structs.schema.json#/definitions/intent" ?

config/taxonomy/application_taxonomy_test.go
eletonia: I thought the CRD will already check if the user submits an m4dapplication without intent/role. These Go tests are checking that the intent/role values are one of [these](https://github.com/IBM/the-mesh-for-data/blob/83c8ae31b788b3a78d6db24e9642b3e2b1f678ce/config/taxonomy/policymanager.values.schema.json#L7) defined in the taxonomy.

config/taxonomy/application.values.schema.json
eletonia: Since we're not referencing that line anymore, I removed intent and role from the definitions

config/taxonomy/application.values.schema.json
```diff
@@ -0,0 +1,73 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "$id": "application.values.schema.json",
+    "title": "Application Values Taxonomy",
+    "description": "Values contributed to by the mesh for data applications.",
+    "definitions": {
+        "intent": {"$ref": "policymanager.structs.schema.json#/definitions/intent"},
+        "role": {"$ref": "policymanager.structs.schema.json#/definitions/role"},
```
ronenkat: do you still need lines 7 and 8 ?

config/taxonomy/application_taxonomy_test.go
ronenkat: The CRD does not define AppInfo values, it is defined as a map in the CRD,
AppInfo is not used by the manager, so it does not understand its content directly.

My feedback regarding the testing is that it would make more sense to start from a m4dapp data structure and get the values from it, rather than have just one field from the m4dapp structure. That test will evolve to a unit test of the actual code eventually.

config/taxonomy/application.values.schema.json
eletonia: I already removed these lines

config/taxonomy/application_taxonomy_test.go
eletonia: The intent/role Go tests are similar to the tests [here](https://github.com/IBM/the-mesh-for-data/blob/83c8ae31b788b3a78d6db24e9642b3e2b1f678ce/config/taxonomy/policymanager_taxonomy_test.go#L60). These simply check whether an intent is provided. The `interface` Go tests are checking if the `protocol` and `data_format` are paired up correctly as defined in the taxonomy. 

@ronenkat I see your point. We can have another Go test file which takes an m4dapplication data structure as an input, and validates all the field values with our taxonomy.

config/taxonomy/application.values.schema.json
```diff
@@ -0,0 +1,73 @@
...
+        "app_info": {
+            "type": "object",
+            "properties": {
+                "intent": { "$ref": "policymanager.structs.schema.json#/definitions/intent"},
+                "role": { "$ref": "policymanager.structs.schema.json#/definitions/role"}
+            },
+            "required": ["intent"]
+        },
+        "interface": {
```
ronenkat, line 15: You should factor out the interface to another file as it should be check for m4dmodule as well.
See Source and Sink definitions in [m4dmodules_types.go](https://github.com/IBM/the-mesh-for-data/blob/master/manager/apis/app/v1alpha1/m4dmodule_types.go#L50)

eletonia: You can refer to the `interface` in this file by pointing to `application.values.schema.json#/definitions/interface` from another schema file.

ronenkat: Right, and I actually gave the wrong motivation.
Down the road interface would be generated instead of manually written.
When a new m4dmodule is applied we would re-generate the values scheme for interface adding new protocols that were newly added in the m4module. For example, someone add a m4module with protocol REST and data_format JSON, we would add these values to the validation scheme for interface.
So it would be good that the ```$ref``` be the other way.

eletonia: I added the `interface` schema to `module.values.schema.json` and am referring to it from `application.values.schema.json` now

ronenkat (approved):

Issue #547, opened 2021-04-25
Credentials connector code cleanup


revit13:
The credentials  connector is used to return the dataset credentials in order for the manager to register them in Vault such that the modules could later retrieve them from Vault tand use them o access the dataset.
As the dataset credentials are now retrieved using Vault plugin directly by the modules as described in issue #406;  credentials  connector code becomes absolute.
This issue is about cleaning all related credentials connector code.

Issue #406, opened 2021-03-03
Retrieve dataset credentials using Vault plugin


revit13:
This is the second milestone of the solution to distributing credentials as described in #140.

This milestone focuses on using Vault secret plugins to retrieve datasets credentials directly from where they are stored: in a catalog or as kubernetes secrets, avoiding the need to copy them to the internal Vault as is done in the current implementation.

More information about the Vault plugin:
https://www.vaultproject.io/docs/internals/plugins

Sub-issues:

- [x] #414  Pluging setup: Add to vault helm installation different values.yaml files to load the plugin depending on the configuration.
- [x]  #448 Extend the catalog connector to return the secretPath in Vault which includes  vault plugin path and the details of the kubernetes secret to retrieve the credentials, if such secret exists. This information will be used by the modules when they send requests to Vault to retrieve the datasets credentials.
- [ ] (PR #592) Add documentation in project web-pages describing what is needs to be done to add a new plugin to the project.

roee88:
Is removing the secret provider part of this milestone too?

revit13:
Secret-provider removal is actually post milestone 1 phase (https://github.com/IBM/the-mesh-for-data/issues/337) as it could be removed after all modules will be able to send requests directly to Vault. (the removal is not part of milestone 1 though). I will add another issue once the modules will be ready if that makes sense. Anyway, I prefer to wait with the removal for a while even after such ability exists to validate everything works fine.

revit13:
Rethinking about that, I'll move all the code removal (including in the connector) to last milestone.

PR #553, opened 2021-04-26
Credentials connector code cleanup

revit13:
The credentials connector is used to return the dataset credentials in order for the manager to register them in Vault such that the modules could later retrieve them from Vault tand use them o access the dataset.
As the dataset credentials are now retrieved using Vault plugin directly by the modules as described in issue #406; credentials connector code becomes absolute.
This PR which addresses issue #547 is about cleaning all related credentials connector code.


charts/m4d/templates/m4d-config.yaml
```diff
@@ -8,14 +8,10 @@ data:
...
-  CREDENTIALS_PROVIDER_NAME: {{ .Values.coordinator.credentialsManager | quote }}
-  CREDENTIALS_CONNECTOR_URL: {{ .Values.coordinator.credentialsManagerConnectorURL | default (printf "%s-connector:80" .Values.coordinator.catalog) | quote }}
   MAIN_POLICY_MANAGER_NAME: {{ .Values.coordinator.policyManager | quote }}
   MAIN_POLICY_MANAGER_CONNECTOR_URL: {{ .Values.coordinator.policyManagerConnectorURL | default (printf "%s-connector:80" .Values.coordinator.policyManager) | quote }}
   USE_EXTENSIONPOLICY_MANAGER: "false" # deprecated
   VAULT_ADDRESS: {{ .Values.coordinator.vault.address | quote }}
-  VAULT_DATASET_HOME: {{ printf "m4d/dataset-creds/" | quote }} # temporary
-  VAULT_DATASET_MOUNT: {{ printf "/v1/sys/mounts/m4d/dataset-creds" | quote }} # temporary
   USER_VAULT_ADDRESS: {{ .Values.coordinator.vault.address | quote }} # deprecated
   USER_VAULT_PATH: "external" # deprecated
```
shlomitk1, line 16: USER_xxx should not be a part of m4d-config. Are they needed at all?

manager/main.go
```diff
@@ -182,15 +182,13 @@ func main() {
 	}
 }
 
-// init vault client and mount the base directory for storing credentials
+// init vault client
+// TODO: the vault client should be used to add roles for the modules.
 func initVaultConnection() (vault.Interface, error) {
```
shlomitk1, line 187: In this implementation manager does not communicate with vault at all. This means:
1. No need for VAULT_TOKEN passed to the manager pod
2. No need for vault.Interface inside manager

froesef:
There are going to be conflicts with #541 . So we should coordinate the merge order of the PRs.

manager/main.go
revit13: I left it as is as the next step would be to let the manager to set Vault policies and Vault roles for modules.

charts/m4d/templates/m4d-config.yaml
revit13: These should have been removed. I'll create a separate PR for it. Thanks

roee88 (approved):

shlomitk1 (approved):

froesef (approved):

rohithdv (approved):
Once this PR is merged I will raise a PR for the wkc connector changes. Thanks.

PR #567, opened 2021-04-29
Use local schema instead of global schema

froesef:
Some tests had trouble when running in parallel. This PR changes the scheme to be local for the tests that are run in parallel.

### DIFF
--- a/manager/controllers/app/m4dapplication_controller_unit_test.go
+++ b/manager/controllers/app/m4dapplication_controller_unit_test.go
@@ -54,11 +54,7 @@ func TestM4DApplicationControllerCSVCopyAndRead(t *testing.T) {
 	}
 
 	// Register operator types with the runtime scheme.
-	s := runtime.NewScheme()
-	err = corev1.AddToScheme(s)
-	g.Expect(err).NotTo(gomega.HaveOccurred())
-	err = app.AddToScheme(s)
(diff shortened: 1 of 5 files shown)


manager/controllers/app/m4dapplication_controller_unit_test.go
```diff
@@ -55,8 +54,10 @@ func TestM4DApplicationControllerCSVCopyAndRead(t *testing.T) {
 	}
 
 	// Register operator types with the runtime scheme.
-	s := scheme.Scheme
-	err = app.AddToScheme(scheme.Scheme)
+	s := runtime.NewScheme()
+	err = corev1.AddToScheme(s)
+	g.Expect(err).NotTo(gomega.HaveOccurred())
+	err = app.AddToScheme(s)
 	g.Expect(err).NotTo(gomega.HaveOccurred())
```
shlomitk1, line 61: We don't really test AddToScheme errors. Can we have a single function that creates a scheme, similar to:
```
s := runtime.NewScheme()
_ = corev1.AddToScheme(s)
...
_ = app.AddToScheme(s)
return s

```

froesef: Sure. We can extract a method that creates a custom scheme for us and registers the most common apis.

shlomitk1 (approved):

PR #571, opened 2021-04-29
fix: static check errors

roee88:
- Fix: usage of deprecated zapr method
- Fix: errors should start with lower case
- Fix: scheme imported twice in same file

@froesef the last one is in `manager/controllers/motion/suite_test.go` please check if I didn't break it.


froesef (approved):
Are we ignoring the static check errors in the CI until now?

PR #574, opened 2021-05-03
Remove manager config

### LINKED ISSUE #516: chore: remove the use of `manager/config`
Kustomize is no longer used so any usage of the layers in `manager/config` should be removed and a proper alternative to each use case should be documented.

froesef:
Fixes #516


charts/m4d/templates/secure-control-plane.yaml
```diff
@@ -0,0 +1,43 @@
+{{- if .Values.secureControlPlane }}
```
roee88, line 1: The Istio resources where going to be added in https://github.com/IBM/the-mesh-for-data/pull/450 but then I decided to drop them in favor of matching documentation in the website for now. I suggest to drop and re-open the other PR if we think it should be added (and decide how best to add them).

charts/m4d/templates/secure-control-plane.yaml
```diff
@@ -0,0 +1,43 @@
...
+spec:
+  podSelector:
+    matchLabels:
+      m4d.ibm.com/componentType: connector
+  policyTypes:
+  - Ingress
+  ingress:
+  - from:
+    - podSelector: {}
+
```
roee88, line 42: This one already exists in the chart

pkg/taxonomy/catalog.structs.schema.json
```diff
@@ -0,0 +1,70 @@
+{
```
roee88, line 1: what are these taxonomy files doing here? Any taxonomy file that was under manager/config can be safely deleted completely. The rest (in top level config/taxonomy) shouldn't be touched in this PR. Let @ritwikc and @eletonia arrange them.

roee88 (approved):
I would remove the top level deploy/undeploy completely.

PR #575, opened 2021-05-03
Add tests for implicit copy module definitions

froesef:
Somehow the implicit copy module definitions were giving an error when deploying as ORC is not supported as format.

I fixed it by adding the ORC format and added a simple deploy in the integration tests so that future regression bugs are covered.

### DIFF
--- a/manager/apis/app/v1alpha1/ifdetails.go
+++ b/manager/apis/app/v1alpha1/ifdetails.go
@@ -8,7 +8,7 @@ package v1alpha1
 type IFProtocol string
 
 // DataFormatType defines data format type
-// +kubebuilder:validation:Enum=parquet;table;csv;json;avro;binary;arrow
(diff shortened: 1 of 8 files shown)


manager/apis/app/v1alpha1/ifdetails.go
```diff
@@ -18,6 +18,7 @@ const (
 	CSV     DataFormatType = "csv"
 	JSON    DataFormatType = "json"
 	AVRO    DataFormatType = "avro"
+	ORC     DataFormatType = "orc"
 	Binary  DataFormatType = "binary"
 	Arrow   DataFormatType = "arrow"
 )
```
shlomitk1, line 24: For consistency, please add the new format in https://github.com/IBM/the-mesh-for-data/blob/6c083e8b7427fbd775d81fea3b9c52f9702da4a1/samples/gui/front-end/src/formats/InputDataTable.js#L4

shlomitk1 (approved):

Issue #581, opened 2021-05-06
Support cert-manager 0.10.x


roee88:
cert-manager 0.10.x is still widely used. However, it uses a different group and api version (`certmanager.k8s.io/v1alpha1`) than what we deploy.

The chart should check for the existence of the API and apply the Issuer and Certificate resources accordingly.
```
{{- if .Capabilities.APIVersions.Has "certmanager.k8s.io/v1alpha1" }}
```

PR #582, opened 2021-05-06
Update controller tools

### LINKED ISSUE #501: Dependencies: bump controller-tools

froesef:
For the update I took https://book.kubebuilder.io/migration/manually_migration_guide_v2_v3.html as orientation as well as the example kubebuilder project.
So this PR does not only update the controller-tools library but also related parts in order to be compliant with the recommendations for kubebuilder v3

Fixes #501


roee88:
Remind me please what doesn't work?

froesef:
I made it work eventually. I thought you might want to try it out locally before we merge it. Some versions are not the same as kubebuilder would use and developers have to be aware to remove their `hack/tools/bin` folder before trying out the new versions as it won't automatically upgrade to the new controller-gen, etcd, kube-apiserver etc...

charts/m4d/templates/webhook-certificates.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if include "m4d.isEnabled" (tuple .Values.manager.enabled (or .Values.coordinator.enabled .Values.worker.enabled)) }}
-apiVersion: cert-manager.io/v1alpha2
+apiVersion: cert-manager.io/v1
```
roee88, line 2: I suggest to revert this change. In face, we have an open issue to support back to 0.10.x https://github.com/IBM/the-mesh-for-data/issues/581

roee88, line 2: I suggest to revert this change. In face, we have an open issue to support back to 0.10.x #581

manager/apis/motion/v1alpha1/streamtransfer_webhook.go
```diff
@@ -24,7 +24,7 @@ func (r *StreamTransfer) SetupWebhookWithManager(mgr ctrl.Manager) error {
 
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 
-// +kubebuilder:webhook:path=/mutate-motion-m4d-ibm-com-v1alpha1-streamtransfer,mutating=true,failurePolicy=fail,groups=motion.m4d.ibm.com,resources=streamtransfers,verbs=create;update,versions=v1alpha1,name=mstreamtransfer.kb.io
+// +kubebuilder:webhook:admissionReviewVersions=v1,sideEffects=None,path=/mutate-motion-m4d-ibm-com-v1alpha1-streamtransfer,mutating=true,failurePolicy=fail,groups=motion.m4d.ibm.com,resources=streamtransfers,verbs=create;update,versions=v1alpha1,name=mstreamtransfer.kb.io
```
roee88, line 27: Do you know why we get the linter issues?

hack/tools/create_kind.sh
```diff
@@ -8,7 +8,7 @@ op=$1
 
 source ./common.sh
 
-K8S_VERSION=${K8S_VERSION:-v1.16.9}
+K8S_VERSION=${K8S_VERSION:-v1.19.1}
```
roee88, line 11: What in the changes require 1.19? It's rather new and we should probably peek an older version to test against (are you sure that these changes add something not supported in 1.16?)

charts/m4d/files/webhook-configs.yaml
```diff
@@ -5,7 +5,6 @@ metadata:
   name: '{{ .Release.Namespace }}-mutating-webhook'
   annotations:
     cert-manager.io/inject-ca-from: '{{ .Release.Namespace }}/serving-cert'
-    certmanager.k8s.io/inject-ca-from: '{{ .Release.Namespace }}/serving-cert'
```
roee88, line 8: This line is required

hack/tools/create_kind.sh
froesef: We upgraded `admissionReviewVersions` to `v1`. And when I run it with 1.16.15 from kind I'm getting an error in the webhook that it requires `v1beta1`. I'll try to make it compatible.

charts/m4d/files/webhook-configs.yaml
froesef: Added again.

roee88 (approved):

PR #584, opened 2021-05-09
Remove credentials manager from site docs.

revit13:


ronenkat (commented):
Where is the credential plug-ins for Vault are described? or more broadly where is what and how we use Vault is described in the docs?

revit13:
@ronenkat I intend to address the Vault plug-ins documentation in a separate pr, I wonder where is a proper place to describe them? Does adding a new link under "Plugins" which is called "Vault plugins for data credentials" is ok? Thanks @roee88

roee88 (approved):

PR #586, opened 2021-05-09
Synchronize m4dapplication with the generated plotter

shlomitk1:
Related to https://github.com/IBM/the-mesh-for-data/discussions/556
This PR adds a new field to Status.Generated of M4DApplication: AppVersion. 
The controller updates this field using the Generation value of M4DApplication. 
Whenever an update from a plotter is received, M4DApplication controller checks that Status.Generated.AppVersion and Generation match, otherwise an update is ignored.


ronenkat:
@shlomitk1 I don't see a description for this PR.

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -74,11 +74,12 @@ func (r *M4DApplicationReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
 	}
 
 	observedStatus := applicationContext.Status.DeepCopy()
+	appVersion := applicationContext.GetGeneration()
 
 	// check if reconcile is required
 	// reconcile is required if the spec has been changed, or the previous reconcile has failed to allocate a Plotter resource
-	generationComplete := r.ResourceInterface.ResourceExists(applicationContext.Status.Generated)
-	if !generationComplete || observedStatus.ObservedGeneration != applicationContext.GetGeneration() {
+	generationComplete := r.ResourceInterface.ResourceExists(observedStatus.Generated) && (observedStatus.Generated.AppVersion == appVersion)
```
froesef, line 81: So it will only generate changes in the plotter if the resource does not exist or there have been changes in the M4DApplication? What if a policy in opa changed?

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -374,6 +373,7 @@ func (r *M4DApplicationReconciler) reconcile(applicationContext *app.M4DApplicat
 	blueprintPerClusterMap := r.GenerateBlueprints(instances, applicationContext)
 	setReadModulesEndpoints(applicationContext, blueprintPerClusterMap, moduleMap)
 	resourceRef := r.ResourceInterface.CreateResourceReference(applicationContext.Name, applicationContext.Namespace)
+	resourceRef.AppVersion = applicationContext.GetGeneration()
```
froesef, line 376: maybe this should be part of the `CreateResourceReference` function if it's part of the same struct and not set separately.

manager/controllers/app/m4dapplication_controller.go
shlomitk1: An interesting question. Suppose the resource exists, and no changes have been made. The end-user is using the data accessed via the received endpoint. Suddenly the policy is changing - some columns need to be redacted. What does this mean from the user perspective? For the next request - sure, data will be redacted. But for the running workload? Not sure how it is supposed to work. Seems more reasonable to me that a new policy will either be entered for a new data, without changing the existing apps, or there will be some sunset warning telling users to stop using old applications after some date, and force-restart services/applications that are older than xxx. @simanadler ?
In any case, this is not relevant to this PR and should be a separate issue.

manager/controllers/app/m4dapplication_controller.go
shlomitk1: ok, will change.

manager/controllers/app/m4dapplication_controller.go
froesef: Yes this is not relevant for this PR. Just wanted to make sure that it's the expected behaviour right now. We should address this in future issues.

simanadler: @froesef  @shlomitk1 When/if/how a blueprint should be updated if a policy changes is something that should be addressed in a separate issue I think.  We need to explore this issue from the perspective of the user.  How is he informed that the policy changed and its affect on the data he is using?  I would think that there should probably be a way for the person creating/changing policies to indicate it's (a) date when it goes into effect (b) whether it should affect currently running workloads - short running, long running, etc.  @shlomitk1 I would suggest adding some of these issues to the smart blueprint requirements as well, because we will have similar issues with non-data governance policies.

froesef (approved):

roee88:
Please rebase to ensure that the repo move had no effect on this PR

shlomitk1:
> 
> 
> Please rebase to ensure that the repo move had no effect on this PR

Done

PR #602, opened 2021-05-25
Register Vault plugins after Vault helm chart installation.

revital76:
The Vault plugin registration phase is done automatically in Vault "dev" mode. (https://www.vaultproject.io/docs/commands/server#dev-options)
This PR adds plugin registration so that the Vault plugin deployment could work also in non "dev" mode.


roee88 (approved):

roee88:
Awesome!

PR #605, opened 2021-05-26
Support different APIVersions of cert-manager.

revit13:
Addresses issue #581.

Tested with quick-start using cert-manager v0.10.1 and cert-manager v1.2.0

### DIFF
--- a/charts/m4d/templates/_helpers.tpl
+++ b/charts/m4d/templates/_helpers.tpl
@@ -93,9 +93,7 @@ Detect the version of cert manager crd that is installed
 Error if CRD is not available
 */}}
 {{- define "m4d.certManagerApiVersion" -}}
-{{- if (.Capabilities.APIVersions.Has "cert-manager.io/v1alpha3") -}}
-cert-manager.io/v1alpha3
-{{- else if (.Capabilities.APIVersions.Has "cert-manager.io/v1alpha2") -}}
+{{- if (.Capabilities.APIVersions.Has "cert-manager.io/v1alpha2") -}}
 cert-manager.io/v1alpha2
 {{- else if (.Capabilities.APIVersions.Has "certmanager.k8s.io/v1alpha1") -}}
 certmanager.k8s.io/v1alpha1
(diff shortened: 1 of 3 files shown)


charts/m4d/templates/_helpers.tpl
```diff
@@ -87,3 +87,19 @@ isRazeeEnabled checks if razee configuration is enabled
...
+{{- define "m4d.certManagerApiVersion" -}}
+{{- if (.Capabilities.APIVersions.Has "cert-manager.io/v1alpha3") -}}
+cert-manager.io/v1alpha3
+{{- else if (.Capabilities.APIVersions.Has "cert-manager.io/v1alpha2") -}}
+cert-manager.io/v1alpha2
+{{- else if (.Capabilities.APIVersions.Has "certmanager.k8s.io/v1alpha1") -}}
+certmanager.k8s.io/v1alpha1
+{{- else  -}}
+{{- fail "cert-manager CRD does not appear to be installed" }}
+{{- end -}}
```
roee88, line 104: Can you check the result of `.Capabilities.APIVersions.Has "cert-manager.io/v1alpha2"` on a cluster with cert-manager 1.3? My assumption is that v1alpha2 is convertible to newer versions so the v1alpha3 branch here is not needed.

revit13: @roee88 you are correct, it seems that cert-manager.io/v1alpha2 is automatically updated to apiVersion: cert-manager.io/v1, and when cert-manager.io/v1alpha3 is supported cert-manager.io/v1alpha2 is also supported (I guess its backward compatibility). I updated the pr to remove cert-manager.io/v1alpha3 as you suggested.
btw, cert-manager v1.2.0 deploying resources with version cert-manager.io/v1alpha3 (I assume you wanted me to test v1.3 to test cert-manager.io/v1alpha3)

roee88: I just assumed that v1.3 uses v1 api, but I didn't check. That's great thank you!

roee88 (approved):

PR #628, opened 2021-06-03
Fix: quickstart demo fails after controller-tools upgrade

roee88:
Following the upgrade of controller-tools the quickstart demo stopped working. This PR is the result of fixing it.

Fixes:
- RawExtension fields got omitted in Plotter. Resolved by using an alternative method that no longer requires RawExtension. A nice side affect is that we no longer have bytes printed to the manager log.
- certmanager version check can't work with `helm template` so I changed to have a fallback.
- Helm `manager.enabled` flag mistakenly controlled deployment of supporting components

Added:
- Documentation for running the manager locally and debugging


shlomitk1 (approved):

manager/README.md
```diff
@@ -8,48 +8,109 @@ The `manager` binary includes all of the controllers that this project defines b
...
+```
+
+### Run the manager
+
+### From command line
+
+You can now run the manager from the `manager` folder using one of these options:
+1. `make run`
+2. `go run main.go --enable-all-controllers --metrics-bind-addr=0`
+
```
revit13, line 74: That did not work for me (both options) I got "missing go.sum entry for module providing" errors

manager/README.md
```diff
@@ -8,48 +8,109 @@ The `manager` binary includes all of the controllers that this project defines b
...
+Components such as connectors need to be running before you run the manager.
+This can be done by one of these options:
+1. Running components locally directly (no instructions provided)
+2. Running components in a cluster 
+
+For option 2, the Helm installation allows you to pick which compoenents to install. 
+Follow the installation guide as usual but in the Helm installation for the control plane add `--set manager.enabled=false` to skip the deployment of the manager. For example:
+
+```bash
```
revit13, line 26: "kubectl create namespace m4d-system" before...

manager/controllers/app/data_catalog_interface.go
```diff
@@ -81,10 +81,11 @@ func (r *M4DApplicationReconciler) RegisterAsset(catalogID string, info *app.Dat
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 	defer cancel()
 
-	datasetDetails := &pb.DatasetDetails{}
-	if err := serde.FromRawExtention(info.Details, datasetDetails); err != nil {
-		return "", err
+	datasetDetails, ok := info.Details.Data.(*pb.DatasetDetails)
```
roee88, line 84: @shlomitk1 this would probably not work, I will fix it in a separate PR. Do we have a test that uses register asset?

manager/README.md
revit13, line 26: "kubectl create ns m4d-system" before...

revit13: btw, m4d-crd should be installed as well right?

manager/README.md
revit13, line 74: I get "missing go.sum entry for module providing package" errors when running both commands...

manager/controllers/app/data_catalog_interface.go
shlomitk1: We don't.

manager/README.md
```diff
@@ -8,48 +8,109 @@ The `manager` binary includes all of the controllers that this project defines b
...
+            "request": "launch",
+            "mode": "auto",
+            "program": "${workspaceFolder}/manager/main.go",
+            "envFile": "${workspaceFolder}/.env",
+            "args": ["--metrics-bind-addr=0", "--enable-all-controllers"]
+        }
+    ]
+}
+```
+
```
revit13, line 97: To debug the manager  m4dapp should be applied in the cluster, right?

revit13 (approved):

manager/README.md
roee88: It says follow the installation instructions as usual. So no need to create namespace (it's created during vault install) and do need to install crds.

manager/README.md
roee88: `go mod tidy`?

manager/README.md
roee88: This just runs the manager from code. If you want the manager to do something you need to give it some work to do yes.

Issue #625, opened 2021-06-02
Prepare release 0.2


roee88:
- [x] Release 0.2 of arrow-flight-module
- [x] Update quickstart vault installation to avoid installing injector
- [x] Update quickstart documentation (currently it links to latest release of M4D and latest master of flight(
- [x] Implement release workflow (sets version 0.2 in charts, tags 0.2) 
- [x] Fix build docs
- [ ] Run quickstart to ensure it works
- [x] Create releases/0.2.0 branch and verify that images are pushed and charts repo has an automatic PR

PR #633, opened 2021-06-06
Prepare for 0.2

roee88:
Related to #625

This PR adds a workflow to facilitate releases. For created a release a `releases/x.y.z` branch needs to be created. 

Changes:
- Updated quickstart docs
- fix: yq install script
- added release workflow

### DIFF
--- a/charts/m4d/Chart.yaml
+++ b/charts/m4d/Chart.yaml
@@ -5,12 +5,10 @@ apiVersion: v2
 name: m4d
 description: Mesh for Data Helm Chart
 type: application
-
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.0
-
+version: latest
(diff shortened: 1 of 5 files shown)


charts/m4d/Chart.yaml
```diff
@@ -5,12 +5,10 @@ apiVersion: v2
...
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.0
-
+version: latest
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 0.1.0
+appVersion: latest
```
revit13, line 14: a minor comment: this means that "--set global.tag=latest" is not needed anymore in the quickstart guide, right?

revit13 (approved):

charts/m4d/Chart.yaml
roee88: right but I'm keeping it for being future proof if we start using the x.y.z-SNAPSHOT approach

PR #636, opened 2021-06-06
Fix current taxonomy definition to be compatible with Draft4/Structual

ronenkat:
Add type to all schemas root
fix definitions of items
Remove recursive definitions of geography

### DIFF
--- a/charts/m4d-crd/Chart.yaml
+++ b/charts/m4d-crd/Chart.yaml
@@ -5,12 +5,10 @@ apiVersion: v2
 name: m4d-crd
 description: M4D CustomResourceDefinitions
 type: application
-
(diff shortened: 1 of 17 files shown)


eletonia (commented):
Do you need to add the `"type": object"` line to [application.values.schema.json](https://github.com/mesh-for-data/mesh-for-data/blob/master/config/taxonomy/application.values.schema.json) as well?

ronenkat:
> Do you need to add the `"type": object"` line to [application.values.schema.json](https://github.com/mesh-for-data/mesh-for-data/blob/master/config/taxonomy/application.values.schema.json) as well?

Thanks. added.

eletonia (approved):
Looks good to me

ritwikc (approved):

PR #638, opened 2021-06-07
Blueprint events

### LINKED ISSUE #629: Change blueprint manager to be namespace scoped
Currently the blueprint controller will respond to events on Blueprint resources that are created outside of m4d-blueprints. This should change.

simanadler:
Closes #629.  Prevents blueprints from being reconciled if created outside of namespace m4d-blueprints, thus preventing "rogue" blueprints from being created manually which would be a data security risk.


manager/controllers/app/blueprint_controller_test.go
```diff
@@ -82,6 +82,57 @@ func TestBlueprintReconcile(t *testing.T) {
...
 
+// Tries to create a blueprint in a namespace other than m4d-blueprints.
+// Result should be that no event is sent to the reconciler, since the predicate function
+// registered during the controller deployment filters it out.
+func TestIllegalBlueprint(t *testing.T) {
+	t.Parallel()
+	g := gomega.NewGomegaWithT(t)
+	logf.SetLogger(zap.New(zap.UseDevMode(true)))
+
+	blueprint, err := readBlueprint("../../testdata/blueprint.yaml")
```
roee88, line 93: Don't you need to set namespace to `rogue-namespace` here so the right value would be used in `objs` (line 97) too?

manager/controllers/app/blueprint_controller.go
```diff
@@ -327,8 +329,18 @@ func NewBlueprintReconciler(mgr ctrl.Manager, name string, helmer helm.Interface
 
 // SetupWithManager registers Blueprint controller
 func (r *BlueprintReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// 'UpdateFunc' and 'CreateFunc' used to judge if the event came from within the blueprint's namespace.
+	// If that is true, the event will be processed by the reconciler.
+	// If it's not then it is a rogue event created by someone outside of the control plane.
+	p := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
```
roee88, line 336: Is CreateFunc enough? (i.e., no need for UpdateFunc)

manager/controllers/app/blueprint_controller_test.go
```diff
@@ -82,6 +82,57 @@ func TestBlueprintReconcile(t *testing.T) {
...
+	}
+
+	res, err := r.Reconcile(context.Background(), req)
+	g.Expect(err).To(gomega.BeNil())
+
+	// Check the result of reconciliation to make sure it has the desired state.
+	// I.e. The event should not have been queued to reconcile
+	g.Expect(res.Requeue).To(gomega.BeFalse(), "reconcile did not requeue request - as is expected")
+
+	// Look for the blueprint.  Shouldn't have been created because the namespace was wrong
```
roee88, line 131: Just to be sure that I understand, in a real setting the blueprint is created but is ignored by the reconciler. In this test code the blueprint "exists" (known to the fake client because it was passed to `NewFakeClientWithScheme`) so why would `cl.Get` fail?

manager/controllers/app/blueprint_controller_test.go
simanadler: @roee88 This line reads the blueprint spec from the yaml.  The spec doesn't include the namespace.  The namespace is supplied with the reconcile request.  See line 121.

manager/controllers/app/blueprint_controller.go
simanadler: @roee88 I originally included Update as well, but removed it since there is no way to update something that cannot be created.  Right?

ronenkat:
If our goal is to restrict the blueprint to a single name space why not use [namespace scoped](https://medium.com/operators/namespace-vs-cluster-scoped-operators-d406f83d069) operators instead of a filter?

Regarding create and update events, my understanding is that the filter doesn't prohibits the CR to be created, it just make the reconsider ignore it, so I can create a rouge blueprint (which will be ignored), then update it, which will be processed.... am I missing anything?

roee88:
> If our goal is to restrict the blueprint to a single name space why not use [namespace scoped](https://medium.com/operators/namespace-vs-cluster-scoped-operators-d406f83d069) operators instead of a filter?

Because it's a single manager for all controllers, not just the blueprint controller.

manager/controllers/app/blueprint_controller_test.go
simanadler: Moved to integration tests

manager/controllers/app/blueprint_controller.go
simanadler: @roee88 I added Update

manager/controllers/app/blueprint_controller_test.go
simanadler: @roee88 You were right. It's not possible to test correctly using unit tests in this fashion.  I moved to using envtest instead.  Tests are now in blueprint-deploy.test

manager/controllers/app/blueprint_deploy_test.go
```diff
@@ -0,0 +1,76 @@
+// Copyright 2020 IBM Corp.
```
roee88, line 1: Please rename to file `blueprint_controller_test.go` to align with the other controller tests

manager/controllers/app/blueprint_deploy_test.go
```diff
@@ -0,0 +1,76 @@
...
+	defer func() {
+		blueprint := &app.Blueprint{ObjectMeta: metav1.ObjectMeta{Namespace: blueprintKey.Namespace, Name: blueprintKey.Name}}
+		_ = k8sClient.Get(context.Background(), blueprintKey, blueprint)
+		_ = k8sClient.Delete(context.Background(), blueprint)
+	}()
+
+	By("Expecting blueprint to be created")
+	Eventually(func() error {
+		return k8sClient.Get(context.Background(), blueprintKey, blueprint)
+	}, timeout, interval).Should(Succeed())
```
roee88, line 37: I think that the declaration of timeout and interval as package scoped variables (in `m4dapplication_controller_test.go`) was not intentional and these should have been defined like in `batchtrasnfer_controller_test.go` instead.

roee88 (commented):
Awesome! Just minor comments

manager/controllers/app/blueprint_deploy_test.go
simanadler: @roee88 There is already a file with that name that does unit tests, which is why I had to choose a different name.

manager/controllers/app/blueprint_deploy_test.go
simanadler: Fixed

roee88 (approved):

Issue #632, opened 2021-06-05
Reduce rbac for Vault plugin installation


roee88:
The current instructions use the upstream vault helm chart, custom values.yaml for it, and a separate yaml for rbac.

The main problem is that the RBAC installed allows reading any secret in the cluster. We want instead to allow this to be configurable (e.g., opt-in namespaces that the plugin can read from). An additional goal is ease of installation.

Proposal: Create a Helm chart that uses the upstream vault chart as a subchart. The new chart will augment the upstream one of custom values and configurable rbac. 

Note: required for operate-first.

PR #642, opened 2021-06-07
Create a new helm chart for Vault deployment

revit13:
Create a new vault helm chart as described in #632


charts/Makefile
```diff
@@ -25,19 +25,14 @@ vault:
...
 	kubectl create namespace m4d-system || true
 	helm repo add hashicorp https://helm.releases.hashicorp.com
-	helm install vault hashicorp/vault --version 0.9.1 \
-                --set "global.openshift=$(WITH_OPENSHIFT)" \
-                --set injector.enabled=false \
-                --set "server.dev.enabled=true" \
-		--values ../third_party/vault/plugin-secrets-kubernetes-reader/values.yaml \
+	helm dependency update vault
+	helm install vault vault \
+                --set "vault.global.openshift=$(WITH_OPENSHIFT)" \
```
roee88, line 30: I think that global is special and is shared with all subcharts so it should be set at top level, are we sure this one works?

charts/vault/Chart.yaml
```diff
@@ -0,0 +1,18 @@
...
+name: vault
+description: Vault Helm Chart
+type: application
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.0.0
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application.
+appVersion: latest
```
roee88, line 14: Would it make sense not to update it to the M4D version but rather match it to the appVersion of the subchart?

roee88 (commented):
Can you add README to describe the different values files?

charts/vault/Chart.yaml
revit13: yes, it makes sense. I will change that `appVersion: 1.6.2` which is the vault appVersion for Vault Version 0.9.1, leaving `version: 0.0.0.`

charts/Makefile
revit13: I checked that and it worked ok,.. for the testing I had to enable the injector to validate that the resources are generated as expected as openshift flag affects the injector deployment.

Makefile
```diff
@@ -33,9 +33,9 @@ run-integration-tests:
 	$(MAKE) docker
 	$(MAKE) -C test/services docker-build docker-push
 	$(MAKE) cluster-prepare-wait
-	$(MAKE) configure-vault
 	$(MAKE) -C charts m4d
 	$(MAKE) -C manager wait_for_manager
+	$(MAKE) configure-vault
```
simanadler, line 38: Changes here need to be reflected in the web site contributing page as well.

revit13: Thanks will update the web page as well then.

roee88 (approved):

PR #643, opened 2021-06-08
Refactoring of PolicyCompiler

### LINKED ISSUE #566: refactor policy compiler
Flatten to a facade, support mocking

roee88:
PolicyCompiler currently has legacy usage as a server and is used as an entry point for testing connectors. This PR refactors the policycompiler and replaces it with an equivalent PolicyManager façade. The main change is that pkg/connector now has interfaces for DataCatalog and PolicyManager clients (with different implementations; main one is a grpc client implementation). These should:
1. Eventually be used whenever a client to a connector is needed (e.g., from the OPA connector)
2. Eventually be generalized to avoid dependency on the protobuf definitions so we can create a parallel implementation of them that is REST based (taxonomy related) without impacting the manager code.

Closes #566


manager/controllers/mockup/policymanager.go
```diff
@@ -37,7 +38,8 @@ func (s *MockPolicyCompiler) GetPoliciesDecisions(in *pb.ApplicationContext) (*p
 		args := make(map[string]string)
 
 		var operationDecisions []*pb.OperationDecision
-		assetID := utils.GetAttribute("asset_id", dataset.GetDatasetId())
+		splittedID := strings.SplitN(dataset.GetDatasetId(), "/", 2)
+		assetID := splittedID[1]
```
shlomitk1, line 42: Safeguard in case datasetD is not of the form xxx/yyy ?

manager/testdata/unittests/m4dcopyapp-csv.yaml
```diff
@@ -17,7 +17,7 @@ spec:
   appInfo:
     intent: fraud-detection
   data:
-  - dataSetID: '{"asset_id": "xxx", "catalog_id": "s3-csv"}'
+  - dataSetID: "s3-csv/allow-dataset"
```
shlomitk1, line 20: This is not the same test. The original is equivalent to "s3-csv/xxx" or "s3-csv/redact-dataset".

manager/testdata/e2e/m4dapplication.yaml
```diff
@@ -17,7 +17,7 @@ spec:
   appInfo:
     intent: fraud-detection
   data:
-    - dataSetID: '{"asset_id": "xxx", "catalog_id": "s3"}'
+    - dataSetID: 's3/allow-dataset'
```
shlomitk1, line 20: This is not the same test. The original required transformations.

manager/controllers/app/data_catalog_interface.go
```diff
@@ -22,44 +20,6 @@ import (
...
-	result, err := d.catalogClient.RegisterDatasetInfo(ctx, req)
-	return result, errors.Wrap(err, "register dataset info failed")
-}
-
-func (d *DataCatalogImpl) Close() error {
-	err := d.catalogConnection.Close()
-	return err
-}
-
 // RegisterAsset registers a new asset in the specified catalog
```
shlomitk1, line 23: Seems as the data catalog facade is not used in
https://github.com/mesh-for-data/mesh-for-data/blob/e01317957492ba978efd922c1dd60cd645ba102c/manager/controllers/app/data_catalog_interface.go#L32

manager/testdata/unittests/m4dcopyapp-csv.yaml
roee88: Will change, but how is that reflected in the test code? (Given that both pass the test)

manager/controllers/mockup/policymanager.go
roee88: Added

manager/controllers/app/data_catalog_interface.go
roee88: Changed

manager/testdata/unittests/m4dcopyapp-csv.yaml
roee88: Got it. Changed according to your suggestion in case the test will need it in the future.

shlomitk1 (approved):

rohithdv (approved):
As per our discussion, we will use this new interface to implement the open api based policy manager. Thanks.

PR #645, opened 2021-06-08
Webhook for validating M4DApplications

shlomitk1:

Fixes https://github.com/mesh-for-data/mesh-for-data/issues/623

This PR introduces a webhook for validating m4dapplication objects.
- Removed hard-coded kubebuilder:validation comments for data format and protocol
- Converted pre-defined types for protocol and data format to strings
- Added checks in the webhook as a place-holder. **To be replaced** by taxonomy validation checks.
- Removed the hard-coded check for size of Data array, added them temporarily into the webhook. **To be removed** after validating that manager works correctly with an empty Data list.
- Fixed incorrect kubebuilder comments and some tests


manager/apis/app/v1alpha1/blueprint_types.go
```diff
@@ -117,7 +117,7 @@ type DataFlow struct {
 	Name string `json:"name"`
 
 	// +required
-	// +kubebuilder::validation:MinItems:=1
+	// +kubebuilder:validation:MinItems:=1
```
roee88, line 120: Does it really make sense to add annotations in a resource that we auto generate? I mean it doesn't check user input but rather that our codebase has no bug and I'm not sure if that's the best way to check it (it will just keep rescheduling a reconcile)

manager/apis/app/v1alpha1/ifdetails.go
```diff
@@ -3,40 +3,24 @@
...
-	CSV     DataFormatType = "csv"
-	JSON    DataFormatType = "json"
-	AVRO    DataFormatType = "avro"
-	ORC     DataFormatType = "orc"
-	Binary  DataFormatType = "binary"
-	Arrow   DataFormatType = "arrow"
-)
-
-// IFProtocol valid values
+// Values used in tests and mockup servers
```
roee88, line 6: I  think it would be better to move these out of `apis`. Do you have a good suggestion where to?

roee88: OK It's still used in `CatalogDatasetToDataDetails` so not just for testing until we get rid of the GRPC stuff.

manager/apis/app/v1alpha1/m4dapplication_webhook.go
```diff
@@ -0,0 +1,129 @@
...
+		return nil
+	case "orc":
+		return nil
+	case "binary":
+		return nil
+	case "arrow":
+		return nil
+	default:
+		return errors.New("Value should be one of these: parquet, table, csv, json, avro, orc, binary, arrow")
+	}
```
roee88, line 128: Not important because it's temporary but you can rewrite as a shorter swtich:
```go
	switch format {
	case "parquet", "table", "csv", "json", "avro", "orc", "binary", "arrow":
		return nil
	default:
		return errors.New("Value should be one of these: parquet, table, csv, json, avro, orc, binary, arrow")
	}
```

roee88 (approved):

manager/apis/app/v1alpha1/blueprint_types.go
shlomitk1: I remember that we decided to add the validation for developer purposes (to create a blueprint and get around all the opa/catalog configurations). I can remove the check, though.

ronenkat (approved):

PR #660, opened 2021-06-16
Allow empty application creation

shlomitk1:

Fixes https://github.com/mesh-for-data/mesh-for-data/issues/639

M4DApplication with no datasets can be created, successfully reconciled but without any generated blueprints.


manager/controllers/app/m4dapplication_controller_unit_test.go
```diff
@@ -813,3 +813,39 @@ func TestSyncWithPlotter(t *testing.T) {
 	g.Expect(getErrorMessages(newApp)).NotTo(gomega.BeEmpty())
 	g.Expect(newApp.Status.Ready).NotTo(gomega.BeTrue())
 }
+
+// This test checks that an empty m4dapplication can be created and reconciled
+func TestNoData(t *testing.T) {
```
ronenkat, line 818: Suggest to rename to `TestM4DApplicationWithNoDatasets`

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -108,7 +108,9 @@ func (r *M4DApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	if hasError(applicationContext) {
 		log.Info("Reconciled with errors: " + getErrorMessages(applicationContext))
 	}
-	if !applicationContext.Status.Ready {
+
+	// trigger a new reconcile if required (the m4dapplication is not ready and not empty)
+	if !applicationContext.Status.Ready && (len(applicationContext.Spec.Data) > 0) {
```
roee88, line 113: Why the extra (brackets) in `(len(applicationContext.Spec.Data) > 0)`?

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -364,6 +366,11 @@ func (r *M4DApplicationReconciler) reconcile(applicationContext *app.M4DApplicat
 		return ctrl.Result{RequeueAfter: 2 * time.Second}, allocErr
 	}
 	// generate blueprint specifications (per cluster)
+	if len(instances) == 0 {
```
roee88, line 369: What it it wasn't empty before? Shouldn't you clear up existing blueprints?

manager/controllers/app/m4dapplication_controller_unit_test.go
shlomitk1: Renamed.

manager/controllers/app/m4dapplication_controller.go
shlomitk1: Removed.

manager/controllers/app/m4dapplication_controller.go
shlomitk1: Added clean-up

roee88 (approved):

manager/controllers/app/m4dapplication_controller.go
```diff
@@ -108,7 +108,9 @@ func (r *M4DApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	if hasError(applicationContext) {
 		log.Info("Reconciled with errors: " + getErrorMessages(applicationContext))
 	}
-	if !applicationContext.Status.Ready {
+
+	// trigger a new reconcile if required (the m4dapplication is not ready and not empty)
+	if !applicationContext.Status.Ready && len(applicationContext.Spec.Data) > 0 {
```
simanadler, line 113: You could write a filter so that the reconcile loop would not be called at all in this situation. For an example, see blueprint_manager.go SetupWithManager (line 331)

shlomitk1: We do need a reconcile loop in this situation, to delete older resources. But only once, i.e. no need to return requeue = true in this case.

ronenkat (approved):

PR #661, opened 2021-06-16
Changing the generated bucket/dataset name

shlomitk1:

Fixes https://github.com/mesh-for-data/mesh-for-data/issues/649

This PR generates a bucket/dataset name without dots because this name is propagated to PVC resource which complains about such names.

### DIFF
--- a/manager/controllers/app/storage.go
+++ b/manager/controllers/app/storage.go
@@ -6,6 +6,7 @@ package app
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/go-logr/logr"
 	app "github.com/mesh-for-data/mesh-for-data/manager/apis/app/v1alpha1"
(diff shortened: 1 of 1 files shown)


rohithdv (approved):
Please ensure the file restrictions are not merely limited to having "."s. For ex: is there any restriction on file name length ? I am not sure about this..but if there is a place where dlf  states any specifications for a valid bucket name  then we need to ensure the generated data set name follows the restrictions. This is just a suggestion. Thanks.

shlomitk1:
> 
> 
> Please ensure the file restrictions are not merely limited to having "."s. For ex: is there any restriction on file name length ? I am not sure about this..but if there is a place where dlf states any specifications for a valid bucket name then we need to ensure the generated data set name follows the restrictions. This is just a suggestion. Thanks.

utils.K8sConformName restricts the dataset resource name, and the bucket name accordingly. I am not aware of any further restrictions.

Issue #666, opened 2021-06-27
Plotter should not respond to events from outside m4d-system


simanadler:
The plotter controller should not respond to reconcile events from outside of m4d-system, which it does today.  In theory someone could manually create a plotter.yaml from another namespace and deploy it.  This would enable blueprints to be deployed, enabling unintended access to data and modules.  I.e. This is a security risk.

PR #667, opened 2021-06-27
Resolves issue #666.

simanadler:
Prevents plotter from receiving events from outside namespace.
Signed-off-by: Sima Nadler <sima@il.ibm.com>

Resolves issue #666


ronenkat:
@simanadler, thank you for addressing this important topic. Please give the PR a descriptive name reflecting it focus/contribution.

manager/controllers/app/plotter_controller_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+	// Create the plotter yaml from the hard coded testdata file
+	plotter := &app.Plotter{}
+	Expect(readObjectFromFile("../../testdata/plotter.yaml", plotter)).ToNot(HaveOccurred())
+
+	// Set the namespace we received
+	plotter.SetNamespace(namespace)
+
+	// Create the Plotter
+	Expect(k8sClient.Create(context.Background(), plotter)).Should(Succeed())
```
ronenkat, line 30: Should each  one be a test on itself? otherwise success and failures will not be deterministic.
I usually one test to check on condition, otherwise, it would be hard to connect a problem with a failure.

manager/controllers/app/plotter_controller_test.go
```diff
@@ -0,0 +1,82 @@
...
+	Eventually(func() error {
+		return k8sClient.Get(context.Background(), plotterKey, plotter)
+	}, timeout, interval).Should(Succeed())
+
+	// Depending on the test being run, sometimes it should succeed and sometimes it shouldn't.
+	if !shouldSucceed {
+		By("Expecting Plotter to never be ready because reconcile should not be called")
+		Eventually(func() bool {
+			Expect(k8sClient.Get(context.Background(), plotterKey, plotter)).To(Succeed())
+			return plotter.Status.ObservedState.Ready
```
froesef, line 50: Isn't `plotter.Status.ObservedState.Ready` always false by default? So while a reconcile is still running it would be false and this test would succeed. I think a solution could be to mandatorily wait for maybe 1-3 intervals and then also check `plotter.Status.ObservedGeneration`.

manager/controllers/app/plotter_controller.go
```diff
@@ -297,7 +300,20 @@ func NewPlotterReconciler(mgr ctrl.Manager, name string, manager multicluster.Cl
...
+	// 'UpdateFunc' and 'CreateFunc' used to judge if the event came from within the system namespace.
+	// If that is true, the event will be processed by the reconciler.
+	// If it's not then it is a rogue event created by someone outside of the control plane.
+	p := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return e.Object.GetNamespace() == utils.GetSystemNamespace()
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return e.ObjectOld.GetNamespace() == utils.GetSystemNamespace()
+		},
```
shlomitk1, line 312: Is there a delete event as well? Delete can force manager to delete an existing blueprint, for example.

manager/controllers/app/plotter_controller_test.go
simanadler: Split into two separate tests

manager/controllers/app/plotter_controller_test.go
simanadler: @froesef In the interval defined in the test the reconcile should have run.  However, I added the check for ObservedGeneration as you suggested just in case.

manager/controllers/app/plotter_controller.go
simanadler: I added the delete event filter, and related test cases

manager/controllers/app/plotter_controller_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+	By("Expecting Plotter to eventually be ready")
+	Eventually(func() bool {
+		Expect(k8sClient.Get(context.Background(), plotterKey, plotter)).To(Succeed())
+		return plotter.Status.ObservedState.Ready
+	}, timeout, interval).Should(BeTrue(), "Plotter is not ready after timeout!")
+
+	// Now create a request to delete the same plotter
+	// Set the namespace to the namespace being used for this test
+	delPlotter := &app.Plotter{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: plotterKey.Name}}
```
shlomitk1, line 140: If it's not from m4d-system, then delPlotter should be created first.

manager/controllers/app/plotter_controller_test.go
```diff
@@ -0,0 +1,185 @@
...
+		})
+
+		// Plotter is successfully deleted when deletion is done from m4d-system only
+		It("Test Plotter Deletion from Correct Namespace", func() {
+			deployDeleteAndCheck(utils.GetSystemNamespace(), true)
+		})
+	})
+})
+
+var _ = Describe("Plotter Controller Illegal Delete Event", func() {
```
shlomitk1, line 170: I think what we should check in the delete test is the following:
    create a Plotter in m4d-system
    create a similar Plotter in default
    delete Plotter in default
    check that a blueprint was not deleted
    delete Plotter in m4d-system
    check that a blueprint was deleted

manager/controllers/app/plotter_controller_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// Now create a request to delete the same plotter
+	// Set the namespace to the namespace being used for this test
+	delPlotter := &app.Plotter{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: plotterKey.Name}}
+
+	// Delete the Plotter
+	if shouldSucceed {
+		Expect(k8sClient.Delete(context.Background(), delPlotter)).Should(Succeed())
+	} else {
+		Expect(k8sClient.Delete(context.Background(), delPlotter)).ShouldNot(Succeed())
```
froesef, line 146: The event filter that is introduced does not prohibit creation, deletion etc events. It just won't call the reconcile loop from events that you filtered out. And as @shlomitk1 pointed out. This statement does not succeed because the plotter you try to delete does not exist and not because the event loop filters it out.

manager/controllers/app/plotter_controller.go
```diff
@@ -297,7 +300,23 @@ func NewPlotterReconciler(mgr ctrl.Manager, name string, manager multicluster.Cl
 
 // SetupWithManager registers Plotter controller
 func (r *PlotterReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// 'UpdateFunc' and 'CreateFunc' used to judge if the event came from within the system namespace.
+	// If that is true, the event will be processed by the reconciler.
+	// If it's not then it is a rogue event created by someone outside of the control plane.
```
froesef, line 305: I think rogue event is a mit misleading. Those events were not created by someone. Those events come from the creation/update/deletion of plotter objects. Those objects just happen to be in the wrong namespace and are thus ignored by the controller. It might be a good idea to log these events being filtered out. So that if someone expects the objects to progress but they don't because they are in the wrong namespace it's mentioned in the logs that the plotter is in the wrong namespace and is thus ignored.

roee88: I think the point is to minimize any work being done. Doesn't it include logs? I don't think a plotter would be created manually "by mistake".

manager/controllers/app/plotter_controller_test.go
simanadler: @shlomitk1 It's not possible to do the test you describe because there is a filter that prevents the reconcile loop from being called if the plotter is created from outside of m4d-system.  Thus, no blueprint will be created when Plotter is created in default.

manager/controllers/app/plotter_controller_test.go
simanadler: @froesef Agreed.  Now that I think about it though, I don't think there is any way to test the delete filter if the create filter is in place.  There's no way to create the plotter outside of m4d-system and so there's no way to show that deleting it outside of m4d-system does not call the reconcile.  I propose to remove the Illegal Delete Event test completely.  Does this make sense to you?

manager/controllers/app/plotter_controller_test.go
shlomitk1: > 
> 
> @shlomitk1 It's not possible to do the test you describe because there is a filter that prevents the reconcile loop from being called if the plotter is created from outside of m4d-system. Thus, no blueprint will be created when Plotter is created in default.

This is exactly the intent of the test. Deleting a plotter in default is not supposed to delete the blueprint that was created by the "legal" plotter.

manager/controllers/app/plotter_controller.go
froesef: Ok. I'm ok with that. A plotter is an internal object.

manager/controllers/app/plotter_controller_test.go
froesef: One could put counters into the controller and count how many times a reconcile loop was executed and check that it's still the same after deleting the plotter object. But that would mean modifying the code and adding a counter just so we can verify this one test. That's probably not the way to go. Let's drop this test.

simanadler: Agreed.  Removing

manager/controllers/app/plotter_controller_test.go
simanadler: @shlomitk1 I can't delete a plotter in default, because there is no way to successfully create one in default.

shlomitk1: You can create a plotter in default. A reconcile won't be called, but there is no relation between two. You can or can't create resources in different namespaces based on the rbac permissions only.

roee88 (approved):

roee88:
@shlomitk1 will #717 cover this?

shlomitk1:
> 
> 
> @shlomitk1 will #717 cover this?

Yes, it will.

roee88:
@simanadler @shlomitk1 can this be closed?

PR #668, opened 2021-06-29
Reduce rbac for Vault plugin installation 

### LINKED ISSUE #632: Reduce rbac for Vault plugin installation

revit13:
This PR reduces rbac for Vault plugin helm installation by applying either https://github.com/mesh-for-data/mesh-for-data/blob/master/charts/vault/templates/vault-crb.yaml or role bindings for selected namespaces if such provided.

Closes #632


charts/vault/templates/vault-crb.yaml
```diff
@@ -1,13 +1,14 @@
+{{- if not .Values.plugins.readKuberenetsSecrets.namespaces }}
```
roee88, line 1: I suggest to replace by something more explicit than the no namespaces case (e.g., a different flag, a `*` namespace, etc.).

charts/vault/templates/vault-crb.yaml
```diff
@@ -1,13 +1,14 @@
...
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: vault
+  name: "vault-crb"
```
roee88, line 5: Make the name reflect that this is for the Kubernetes secrets plugin

charts/vault/templates/vault-rb.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if .Values.plugins.readKuberenetsSecrets.namespaces }}
```
roee88, line 1: I don't think this if is needed, the range should be enough isn't it?

charts/vault/templates/vault-rb.yaml
```diff
@@ -0,0 +1,19 @@
...
+{{- $releaseNamespace := .Release.Namespace -}}
```
roee88, line 2: Consider using "{{ $.Release.Namespace }}" directly instead of adding this variable

charts/vault/templates/vault-rb.yaml
```diff
@@ -0,0 +1,19 @@
...
+{{- range $ns := .Values.plugins.readKuberenetsSecrets.namespaces }}
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: "vault-rb-{{ $ns }}"
```
roee88, line 7: Make the name reflect that this is for the Kubernetes secrets plugin. No need for the namespace in the name, it's namespace scoped.

charts/vault/templates/vault-rb.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if .Values.plugins.readKuberenetsSecrets.namespaces }}
+{{- $releaseNamespace := .Release.Namespace -}}
+{{- range $ns := .Values.plugins.readKuberenetsSecrets.namespaces }}
```
roee88, line 3: Are you sure it's the right syntax for capturing the values from the list? (I'm only aware of capturing both index and value, or a bare range with values captured in `.`.  Just verifying.

charts/vault/values.yaml
```diff
@@ -0,0 +1,11 @@
+plugins:
+  # Settings related to vault-plugin-secrets-kubernetes-reader plugin  
+  readKuberenetsSecrets:
```
roee88, line 3: Consider using the plugin name

roee88 (changes requested):

charts/vault/templates/vault-rb.yaml
revit13: it did not work for some reason... that is why I used this variable here

charts/vault/templates/vault-rb.yaml
revit13: Please see the section "Iterating over a .Values List" in https://kb.novaordis.com/index.php/Helm_Template_range.

charts/vault/values.yaml
revit13: Will do that. Thanks

charts/vault/templates/vault-rb.yaml
revit13: Thanks will do that

charts/vault/templates/vault-rb.yaml
revit13: Fixed that. Thanks

charts/vault/templates/vault-rb.yaml
revit13: Changed that. Thanks

charts/vault/templates/vault-rb.yaml
roee88: It shows what I meant in `or a bare range with values captured in .` 
Anyway, just test that it works, in go it would capture indexes but maybe in go templates it's different.

revit13: yes, it works I tested it before sending the pr :-)

roee88: Awesome, thanks. Indeed https://golang.org/pkg/text/template/  says:

> Note that if there is only one variable, it is assigned the element; this is opposite to the convention in Go range clauses.

charts/vault/templates/vault-rb.yaml
```diff
@@ -0,0 +1,16 @@
+{{- range $ns := .Values.plugins.vaultPluginSecretsKubernetesReader.namespaces }}
```
roee88, line 1: Maybe only if cluster flag is false?

roee88 (approved):
Great, minor comments to consider before merge

charts/vault/templates/vault-rb.yaml
```diff
@@ -0,0 +1,16 @@
...
+  namespace: "{{ $ns }}"
+subjects:
+- kind: ServiceAccount
+  name:  vault
+  namespace: {{ $.Release.Namespace }}
+roleRef:
+  kind: ClusterRole
+  name: vault-secrets-manager
+  apiGroup: rbac.authorization.k8s.io
+---
```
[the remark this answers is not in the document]
revit13: It was needed because it generates two resources. maybe there is another way to do that...

charts/vault/values.yaml
```diff
@@ -0,0 +1,12 @@
+plugins:
+  # Settings related to vault-plugin-secrets-kubernetes-reader plugin  
+  vaultPluginSecretsKubernetesReader:
```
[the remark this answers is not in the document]
revit13: I will check...

charts/vault/templates/vault-rb.yaml
revit13: either way is fine... no harm in having cluster-role and per namespace role I think... but as you prefer

charts/vault/templates/vault-rb.yaml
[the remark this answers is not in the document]
roee88: Right, forgot the loop 😆

charts/vault/templates/vault-rb.yaml
roee88: It might cause confusion, but good enough for now.

charts/vault/values.yaml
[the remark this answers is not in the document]
revit13: its a syntax error... helm template does not like the dash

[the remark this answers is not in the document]
roee88: :(

ronenkat (approved):

PR #670, opened 2021-06-30
Minor fixes to docs/contribute/modules.md.

revit13:

### DIFF
--- a/site/docs/contribute/modules.md
+++ b/site/docs/contribute/modules.md
@@ -23,13 +23,13 @@ Modules that access or write data need credentials in order to access the data s
 An example for Vault Login API call which uses the Vault parameters is as follows:
 
 ```
-$ curl -v --request POST <address>/<authPath> -H "Content-Type: application/json" --data '{"jwt": <module service account token>, "role": <role>}'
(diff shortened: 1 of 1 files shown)


site/docs/contribute/modules.md
```diff
@@ -23,13 +23,13 @@ Modules that access or write data need credentials in order to access the data s
 An example for Vault Login API call which uses the Vault parameters is as follows:
 
 ```
-$ curl -v --request POST <address>/<authPath> -H "Content-Type: application/json" --data '{"jwt": <module service account token>, "role": <role>}'
+$ curl -v -X POST <address>/<authPath> -H "Content-Type: application/json" --data '{"jwt": <module service account token>, "role": <role>}'
```
shlomitk1, line 26: @eres Would this work with the latest kind release?

revit13: @shlomitk1 yes. if you refer to the error reported on https://github.com/mesh-for-data/mesh-for-data/issues/672 so the fix is when configuring Vault (not when retrieving  the secrets), more sepcifcally the following `issuer` should be added to the vault configuration:
`vault write auth/kubernetes/config issuer="https://kubernetes.default.svc.cluster.local" kubernetes_host="https://kubernetes.default.svc:443"`

roee88 (approved):

PR #676, opened 2021-07-04
Manage Dataset objects as unstructured to reduce import dependencies  

shlomitk1:

Fixes https://github.com/mesh-for-data/mesh-for-data/issues/674


go.mod
```diff
@@ -76,25 +74,4 @@ replace (
 	github.com/google/addlicense => github.com/the-mesh-for-data/addlicense v0.0.0-20200913135744-636c44b42906
 	github.com/onsi/gomega => github.com/onsi/gomega v1.10.0
 	helm.sh/helm/v3 v3.5.2 => github.com/the-mesh-for-data/helm/v3 v3.1.0-rc.1.0.20210505082940-f9c923d43d3a
-	k8s.io/api => k8s.io/api v0.19.2
```
roee88, line 79: Do we need the replacement for the docker stuff? Why are they even there.

pkg/storage/storage_provisioning.go
```diff
@@ -62,24 +67,57 @@ func NewProvisionImpl(c client.Client) *ProvisionImpl {
...
+
+	objectKey := client.ObjectKeyFromObject(object)
+
+	if err := r.Client.Get(context.Background(), objectKey, object); err != nil {
+		return nil, err
+	}
+	return object, nil
+}
+
+func getValue(obj map[string]interface{}, path []string) string {
```
roee88, line 92: Consider rewriting as a variadic function (path ...string) to simplify calling code.

go.mod
shlomitk1: > 
> 
> Do we need the replacement for the docker stuff? Why are they even there.

Because of helm dependency. I have tried to remove them - got:
```
go: helm.sh/helm/v3@v3.5.2 requires
	github.com/deislabs/oras@v0.10.0 requires
	github.com/docker/distribution@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
```

pkg/storage/storage_provisioning.go
shlomitk1: OK.

roee88 (approved):

PR #680, opened 2021-07-05
draft: New Smart blueprints examples

froesef:


manager/testdata/new-blueprints/read-transform-blueprint.yaml
```diff
@@ -0,0 +1,92 @@
...
+  - name: transform-step
+    assetID: m4d-notebook-sample/paysim-csv
+    templateSpec:
+      chart: ghcr.io/mesh-for-data/m4d-transform-conf-arrow-flight:0.1.0
+      moduleLabel: "mod: common-transform-arrow-flight"
+    source:
+      step: read-data # the name of step for debugging purposes
+      selector: "mod: common-arrow-flight-read-service" # Selector of the dependency
+      connection:
+        service:
```
ronenkat, line 45: is this an indent issue for service? or what is service under "connection"?

manager/testdata/new-blueprints/copy-read-blueprint.yaml
```diff
@@ -0,0 +1,96 @@
...
+        authPath: /v1/auth/kubernetes/login
+        role: module
+        secretPath: /v1/kubernetes-secrets/paysim-csv?namespace=m4d-notebook-sample
+      format: parquet
+      connection:
+        s3:
+          endpoint: localhost:8001
+          bucket: destbucket
+          object: data.parq
+    transformations:
```
ronenkat, line 41: in m4dmodule it is called actions - lets have consistent term in the new blueprints.

manager/testdata/new-blueprints/copy-read-blueprint.yaml
```diff
@@ -0,0 +1,96 @@
+apiVersion: app.m4d.ibm.com/v1alpha2
+kind: Blueprint
+metadata:
+  name: copy-and-read
+  labels:
+spec:
+  cluster: theshire
+  appSelector: "app:demoapp"
+  appCluster: theshire
```
shlomitk1, line 9: Instead of appCluster and appSelector you can reuse https://github.com/mesh-for-data/mesh-for-data/blob/bedb8ecbbd215d733c2a4c6f2904efd8eeccb4df/manager/apis/app/v1alpha1/selector.go#L11

```
appSelector:
    clusterName: theshire
    workloadSelector:
      matchLabels:
        app: demoapp
```

manager/testdata/new-blueprints/copy-read-blueprint.yaml
```diff
@@ -0,0 +1,96 @@
...
+    - action: redact
+      column: blood_group
+  - name: read-step
+    templateSpec:
+      chart: ghcr.io/mesh-for-data/m4d-arrow-flight:0.1.0
+    moduleSelector: "mod: demoapp-read-step"
+    assetID: 'm4d-notebook-sample/paysim-csv'
+    source:
+      step: copy-step # the name of step for debugging purposes
+      selector: "mod: demoapp-copy-step" # Selector of the dependency
```
shlomitk1, line 51: Can there be multiple dependencies?

manager/testdata/new-blueprints/copy-read-blueprint.yaml
froesef: I'll rename it in the m4dmodule. It's called transformations in the configuration design document as well. Please tell me if you think it should be called actions anyways. Then we should rename it everywhere.

manager/testdata/new-blueprints/read-transform-blueprint.yaml
froesef: This is the source specification for the transform step. The source in this case is a service that is using arrow-flight protocol etc that is defined in the `read-data` step. I think it's easier to reuse existing fields explicitly here to state that there is a dependency to the first step.

manager/testdata/new-blueprints/copy-read-blueprint.yaml
revit13: When we discussed it before we said that we do not merge or combine two datasets for governance enforcement purposes.

manager/testdata/new-blueprints/write-module.yaml
```diff
@@ -0,0 +1,47 @@
...
+  description: "This module knows how to configure a cluster wide arrow flight service"
+  # this field is true if this module serves multiple users. Otherwise it
+  # is false. It is used for example when setting the endpoint to
+  # access the module service - if it is a multi-user module then
+  # the service name will be fixed and will be taken from
+  # M4DModule.Capabilities.Api.endpoint.host.
+  multi-user: true
+  confChart:
+    name: ghcr.io/mesh-for-data/afwm:0.1.0
+    values: {}
```
revit13, line 18: This module will deploy the arrow-flight server in a lazy fasion, right? i.e., it a configuration module but also deploys the chart if it is not running. in case the server was deployed already (not by m4d for example) I guess that the chart name will be omitted.

manager/testdata/new-blueprints/write-plotter.yaml
```diff
@@ -0,0 +1,135 @@
...
+    - id: 02
+      module: arrow-flight-common-transform-conf
+      cluster: thegreendragon
+      parameters:
+        source:
+          step: 01 # Not sure if this is redundant if we see the dash listings as flow.
+          # source details are taken from spec of flow 01. No need to repeat here
+        transformations:
+        - action: redact
+          column: blood_group
```
revit13, line 67: should an api field be added to step 2?

manager/testdata/new-blueprints/write-plotter.yaml
```diff
@@ -0,0 +1,135 @@
...
+  - - id: 03
+      module: arrow-flight-common-write-conf
+      parameters:
+        api:
+          service:
```
revit13, line 72: step 3 is independent of step 1 or 2?

manager/testdata/new-blueprints/copy-read-blueprint.yaml
roee88: +1 for actions everywhere

manager/testdata/new-blueprints/write-plotter.yaml
```diff
@@ -0,0 +1,135 @@
...
+        destination:
+          connection:
+            s3:
+              endpoint: localhost:8001
+              bucket: destbucket
+              object: appout.parq
+      cluster: thegreendragon
+  - - id: 04
+      module: credit-card-detector
+      parameters:
```
simanadler, line 86: See comment in module definitions about how to indicate where/how the results of this module are provided to the user.

manager/testdata/new-blueprints/write-plotter.yaml
```diff
@@ -0,0 +1,135 @@
...
+    status: Ready
+  - name: isolation-02-to-app
+    status: Ready
+  - name: isolation-app-to-03
+    status: Ready
+  - name: isolation-app-to-04
+    status: Ready
+  - name: multicluster-01-to-02
+    status: Ready
+  steps:
```
simanadler, line 110: Is this section redundant?  Won't all these steps and their status be listed under either global or assets?

manager/testdata/new-blueprints/read-transform-blueprint.yaml
```diff
@@ -0,0 +1,95 @@
...
+metadata:
+  name: read-and-transform
+  labels:
+spec:
+  cluster: thegreendragon
+  appSelector: # Selector of the application that uses this workload
+    clusterName: thegreendragon
+    workloadSelector:
+      matchLabels:
+        app: demoapp
```
simanadler, line 12: add namespace

manager/testdata/new-blueprints/read-transform-blueprint.yaml
```diff
@@ -0,0 +1,95 @@
...
+  steps:
```
simanadler, line 13: I would suggest changing steps to modules, since this is a list with no particular order.  I.e. blueprint doesn't include data path order, just the modules that need to be deployed/configured

manager/testdata/new-blueprints/read-transform-blueprint.yaml
```diff
@@ -0,0 +1,95 @@
...
+        protocol: arrow-flight
+        dataformat: arrow
+    transformations: [] # No transformations done in this module
+  - name: transform-step
+    assetID: m4d-notebook-sample/paysim-csv
+    templateSpec:
+      chart: ghcr.io/mesh-for-data/m4d-transform-conf-arrow-flight:0.1.0
+      moduleLabel: "mod: common-transform-arrow-flight"
+    source:
+      step: read-data # the name of step for debugging purposes
```
simanadler, line 45: I don't think this is necessary

manager/testdata/new-blueprints/read-transform-blueprint.yaml
```diff
@@ -0,0 +1,95 @@
...
+    moduleSelector: "mod: common-transform-arrow-flight"
+  - name: isolation-step2-to-app
+    templateSpec: # Configure here or in “reading” module?
+      name: ghcr.io/mesh-for-data/networkpolicy-isolation:0.1.0
+    appSelector: "app: demoapp"
+    moduleSelector: "mod: common-transform-arrow-flight"
+status:
+  observedState: Ready
+  observedGeneration: 1
+  steps:
```
simanadler, line 78: modules rather than steps

manager/testdata/new-blueprints/read-transform-blueprint.yaml
```diff
@@ -0,0 +1,95 @@
...
+    - name: read-data
+      status: Ready
+    - name: transform-step
+      status: Ready
+    - name: isolation-step1
+      status: Ready
+    - name: isolation-step2-to-app
+      status: Ready
+  assets:
```
simanadler, line 87: Should this be in the blueprint?  A particular asset data path may have modules running in different clusters / blueprints, thus this would not represent its true state.

simanadler (commented):
Great job @froesef .  This is great progress.  Don't be scared by the number of my comments.  Many are just to confirm my understanding.  :)

manager/testdata/new-blueprints/write-module.yaml
froesef: No there is no lazy fashion. (At least for now I would suggest) The service will be installed by the admin when the `M4DModule` CRD will be installed in each cluster. The chart is containing a K8s object that configures the service. (Be it a config map or a pod that sends a curl command to the service)

manager/testdata/new-blueprints/copy-read-blueprint.yaml
froesef: Sounds good! Integrated it in the new version.

manager/testdata/new-blueprints/copy-module.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: app.m4d.ibm.com/v1alpha2
+kind: M4DModule
+metadata:
+  name: copy
+  namespace: m4d-system
+  labels:
+    version: 0.0.1
```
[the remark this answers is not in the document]
froesef: I'll add it.

manager/testdata/new-blueprints/copy-module.yaml
```diff
@@ -0,0 +1,34 @@
...
+    sinks:
+    - protocol: s3
+      dataformat: parquet
+    transformations:
+    - id: redact
+      level: column
+    - id: removed
+      level: column
+status:
+  observedStatus: Ready
```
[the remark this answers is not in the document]
froesef: This is just as an example what could be there. We don't have to implement a controller but still plan for a status. I still believe that it's a very important feature.

manager/testdata/new-blueprints/copy-module.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: app.m4d.ibm.com/v1alpha2
+kind: M4DModule
+metadata:
+  name: copy
+  namespace: m4d-system
```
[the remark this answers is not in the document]
froesef: The control plane does not install M4DModules in different clusters. `M4DModule`s are deployed by admins at mesh installation and mostly reside in m4d-system namespace. The old example yamls don't specify a namespace but they are installed in the `m4d-system` namespace.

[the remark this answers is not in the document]
roee88: I suggest to focus on the spec.. anything else is noise at the moment :)

manager/testdata/new-blueprints/credit-card-detection-module.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: app.m4d.ibm.com/v1alpha2
+kind: M4DModule
+metadata:
+  name: credit-card-detector
+  namespace: m4d-system
```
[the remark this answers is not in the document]
froesef: See my comment above. Even now this is not the case. The control plane only distributes `Blueprint` objects.
These `M4DModules` are the "classes" and what get's distributed are "instances" of these classes.

manager/testdata/new-blueprints/credit-card-detection-module.yaml
```diff
@@ -0,0 +1,22 @@
...
+  labels:
+    version: 0.0.1
+spec:
+  multi-user: false
+  chart:
+    name: ghcr.io/mesh-for-data/cc-detector:0.1.0
+  statusIndicators: # Indicators for a K8s deployment
+  kind: deployment # Assume for now that the program is a K8s deployment per dataset
+  capabilities:
+  - capability: audit
```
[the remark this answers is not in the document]
froesef: This is a good point. This module might be special because it'll probably come from the policies. Output may have multiple options here. It could write logs to COS, send out mail, notify a slack channel, ...

manager/testdata/new-blueprints/network-policy-module.yaml
```diff
@@ -0,0 +1,23 @@
...
+  description: "This module knows how to configure a kubernetes network policy"
+  multi-user: true
+  confChart:
+    name: ghcr.io/mesh-for-data/networkpolicy-isolation:0.1.0
+    values: {}
+  statusIndicators:
+  kind: configuration
+  capabilities:
+  - capability: isolation
+    scope: dataset
```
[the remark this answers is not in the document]
froesef: I would say it may have many scopes. Let me list some examples:
1. A Blueprint has steps that deploy an arrow flight service per data asset (like current implementation) -> It's dataset scoped
2. A Blueprint has steps that configure a common flight service -> Might seem workload related but when you think about e.g. the example where a common read service gets connected with a common transform service then the isolation definition between those services would be the same for every workload. This it would be cluster scoped.

manager/testdata/new-blueprints/quality-check-module.yaml
```diff
@@ -0,0 +1,22 @@
...
+  labels:
+    version: 0.0.1
+spec:
+  multi-user: true
+  chart:
+    name: ghcr.io/mesh-for-data/m4d-quality-check:0.1.0
+  statusIndicators: # Indicators for a K8s job
+  kind: deployment
+  capabilities:
+  - capability: quality-check
```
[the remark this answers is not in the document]
froesef: I'll add some examples.

manager/testdata/new-blueprints/read-module.yaml
```diff
@@ -0,0 +1,43 @@
...
+  namespace: m4d-system
+  labels:
+    version: 0.0.1
+spec:
+  description: "This module knows how to configure a cluster wide arrow flight service"
+  # this field is true if this module serves multiple users. Otherwise it
+  # is false. It is used for example when setting the endpoint to
+  # access the module service - if it is a multi-user module then
+  # the service name will be fixed and will be taken from
+  # M4DModule.Capabilities.Api.endpoint.host.
```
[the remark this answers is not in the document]
froesef: I'll add the endpoint into the status of the blueprint. (It's already in the plotter status. I just did not come along to also update the blueprints) 
I think returning possible different endpoints for every time an application runs is not user friendly. This is also why @tomersolomon1 added the possibility to specify an endpoint that the user expects.

What would be nice for the user:
1. User specifies one endpoint (service name) in the same namespace as the application.
2. M4D manages this service and makes sure that it points to the correct module
3. In the best case ALL data access goes via arrow-flight through this endpoint. (Thus all the user does is change the asset name and a different asset will be retrieved)

manager/testdata/new-blueprints/read-transform-blueprint.yaml
froesef: This represents the state of all modules related to this data asset on this cluster correct. This I thought might be good for debugging purposes for developers/admins. (As `Blueprints` are not meant to be viewed by users anyways)

manager/testdata/new-blueprints/transform-module.yaml
```diff
@@ -0,0 +1,40 @@
...
+  namespace: m4d-system
+  labels:
+    version: 0.0.1
+spec:
+  multi-user: true
+  chart:
+    name: ghcr.io/mesh-for-data/m4d-arrow-flight-common-transform-conf:0.1.0
+    values: {}
+  statusIndicators:
+  kind: configuration
```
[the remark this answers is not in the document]
froesef: A jar is very specific to a module that implements thinks on the JVM. I do not think that the m4d should go in such detail. What if a module is implemented in golang or something else? This is an implementation detail of the helm chart. M4D should input which action (conforming to the M4D taxonomy model) should be configured. 

As an example: The request says to `redact` a column.
The helm chart can have an if else statement where for a `redact` action get's resolved to write a configmap that contains information about the jar etc. 
The implicit copy module does this similarly already: https://github.com/mesh-for-data/mesh-for-data/blob/5e853da67bf519bae30a74df9bcdc291716d36c5/modules/m4d-implicit-copy-batch/templates/batchtransfer.yaml#L62 (not with jars but translating M4D taxonomy to implicit copy taxonomy)

manager/testdata/new-blueprints/transform-module.yaml
```diff
@@ -0,0 +1,40 @@
...
+  dependencies: # Possibly define dependencies for the module to be "Ready"
```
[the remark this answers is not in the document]
froesef: I'll add a comment that describes this better. It's related to a field that already exists in the M4DModule that is useful. But this is more related to the status.Ready field that you don't like ;-)
This describes what pods and services the module needs to be up and running in order to be functional. The service here is the already running common arrow-flight server that is needed)

manager/testdata/new-blueprints/transform-module.yaml
```diff
@@ -0,0 +1,40 @@
...
+    scope: dataset
+    source:
+      protocol: m4d-arrow-flight
+      dataformat: arrow
+    transformations:
+    - id: redact
+      level: column
+    - id: remove
+      level: column
+  api:
```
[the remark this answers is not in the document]
froesef: It's meant to be put into the data path when accessing the read module. So the access path would be app reads from transform which reads form read service which reads from S3.

manager/testdata/new-blueprints/write-plotter.yaml
```diff
@@ -0,0 +1,135 @@
...
+        address: http://vault.m4d-system:8200
+        authPath: /v1/auth/kubernetes/login
+        role: module
+        secretPath: /v1/kubernetes-secrets/paysim-csv?namespace=m4d-notebook-sample
+      connection:
+        s3:
+          endpoint: localhost:8001
+          bucket: srcbucket
+          object: data.parq
+    dataPathFlowType: read # @Sima only one flow type?
```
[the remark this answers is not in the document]
froesef: Well the data sets that you specify here above are the ones registered in a catalog somewhere. So they are either read/transform/copy. 
We did not yet talk about how to specify write assets. Don't write assets have to be specified in the M4DApplicaton as well? The user has to specify what data should be written, to which targets etc.

manager/testdata/new-blueprints/write-plotter.yaml
```diff
@@ -0,0 +1,135 @@
...
+  # Flows used in this workflow
+  # The syntax is reversed to argo as the default for m4d is parallel workflows.
+  # Two dashes start a new parallel action and the sub-flows with one dash are the in-path dataflows.
+  # flow 01 reads the data asset and offers it on an interface
+  # flow 02 reads the data asset from flow 01, applies transformations and offers it to the application
```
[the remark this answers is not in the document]
froesef: It's a chain. 02 reads from 01. We should discuss about terminology what we want to call flows and what steps. My guess is you mean flows run in parallel and steps are sequential items in these flows right?

manager/testdata/new-blueprints/write-plotter.yaml
```diff
@@ -0,0 +1,135 @@
...
+  # flow 03 takes data written out as data asset "out1" to the arrow flight service and stores it on S3
+  # flow 04 takes data written out as data asset "out1" to an arrow flight service and does a misuse detection on it
+  #
+  # Networking modules
+  # Open questions:
+  # - Should the global modules be generated during blueprint creation or should they be part of the flow and output
+  #     of the policy compiler/optimizer?
```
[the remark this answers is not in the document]
froesef: Ok. If the latter generates them they would be part of this spec. So I would introduce another field `modules`.

manager/testdata/new-blueprints/write-plotter.yaml
froesef: Yes. Forgot to specify it here. Adding it.

manager/testdata/new-blueprints/write-plotter.yaml
froesef: correct. I'll leave it up for discussion if we want to keep both representations or go with one.

manager/testdata/new-blueprints/credit-card-detection-module.yaml
```diff
@@ -0,0 +1,22 @@
...
+    scope: dataset
+    sources:
+    - protocol: m4d-arrow-flight
+      dataformat: arrow
+status:
```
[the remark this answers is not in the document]
revit13: currently helm status and kuberenetes status is currently used. is the purpose of the statusIndicators field is in addition to the helm status?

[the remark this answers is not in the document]
revit13: in a multi-cluster setup where two dependent modules are in separate clusters, we will need to find a way to when they are ready.

manager/testdata/new-blueprints/credit-card-detection-module.yaml
```diff
@@ -0,0 +1,22 @@
...
+  name: credit-card-detector
+  namespace: m4d-system
+  labels:
+    version: 0.0.1
+spec:
+  multi-user: false
+  chart:
+    name: ghcr.io/mesh-for-data/cc-detector:0.1.0
+  statusIndicators: # Indicators for a K8s deployment
+  kind: deployment # Assume for now that the program is a K8s deployment per dataset
```
revit13, line 14: Currently, statusIndicator allow to check status of a non-standard resource that can not be computed by helm/kstatus.
IIUC it meaning Is now changed - it no longer refers to only non-standard resource but also indicates the resource type that we should look for its status - if its configuration kind then we will look in the status of the config-map. if its deployment we will look at the status of the deployment (possibly with the same name of the deployed m4dmodule). is my understanding correct? Thanks

froesef: It did not change. It's the same as currently. I'm just writing with the comment that this is probably a K8s deployment type. (same as in current arrow flight service module that does not have any statusIndicator at all)

manager/testdata/new-blueprints/write-plotter.yaml
```diff
@@ -0,0 +1,150 @@
...
+            assetID: m4d-notebook-sample/paysim-csv
+            interfaceDetails:
+              protocol: arrow-flight
+              dataformat: arrow
+  - - id: 03
+      module: arrow-flight-common-write-conf
+      parameters:
+        api:
+          service:
+            assetId: "out1"
```
revit13, line 83: Can you please explain which step was responsible for supplying assertId "out1 " in arriwflight format?  there should have been a read module that puts the data in flight format before writing it, is that correct?

manager/testdata/new-blueprints/write-module.yaml
```diff
@@ -0,0 +1,39 @@
...
+    version: 0.0.1
+    name: arrow-flight-common-write-conf
+spec:
+  description: "This module knows how to configure a cluster wide arrow flight service"
+  # this field is true if this module serves multiple users. Otherwise it
+  # is false. It is used for example when setting the endpoint to
+  # access the module service - if it is a multi-user module then
+  # the service name will be fixed and will be taken from
+  # M4DModule.Capabilities.Api.endpoint.host.
+  multi-user: true
```
revit13, line 16: does it makes sense to rename multi-user field to configuration-module as this field is true if this module configures another service.

manager/testdata/new-blueprints/copy-read-blueprint.yaml
```diff
@@ -0,0 +1,99 @@
...
+    actions:
+    - action: redact
+      column: blood_group
+  - name: read-step
+    templateSpec:
+      chart: ghcr.io/mesh-for-data/m4d-arrow-flight:0.1.0
+    moduleSelector: "mod: demoapp-read-step"
+    assetID: 'm4d-notebook-sample/paysim-csv'
+    source:
+      step: copy-step # the name of step for debugging purposes
```
revit13, line 53: Can you please explain again how the read module would know if the bucket in the source is ready? `step: copy-step` is the release name of the dependent copy module which the current read module could investigate for its status but I think you had another solution in mind. Thanks.

manager/testdata/new-blueprints/write-module.yaml
froesef: I don't even think that we need this field at all. As @ronenkat also mentioned what would a controller do with this information? In the end it does not matter what a module deploys. It just has to make sure that it can manage the status and knows how to access the results. No matter if it's configuring a service or starting a K8s job/deployment.

manager/testdata/new-blueprints/copy-read-blueprint.yaml
froesef: The read module does not need to know if its source is ready. The read module is ready when the arrow flight server is up and running. The copy is marked ready when the data is copied. Only when both modules are ready the plotter as well as the application is marked as ready. If a user wants to access a resource before it's marked as ready there may of course be errors. The point is that the readyness check get's pushed to the overall ready status of the application. (Only when all modules that are involved in an asset are successful the asset is also ready)

manager/testdata/new-blueprints/write-plotter.yaml
froesef: It's an output asset that is created by the application. It needs a name so that a common arrow flight service may differentiate different outputs. This is just a proposal on how such an asset could be specified. We still need to define a write path in general. (An application should e.g. specify which outputs it's going to write)

manager/testdata/new-blueprints/write-module.yaml
revit13: I think that it matters for the api.endpoint - if its a configuration module this field must be filled and will hold the common service endpoint otherwise it is like the current implementation- the controller generates the release name based on the steps etc to make sure there are no conflicts in the resources names.

manager/testdata/new-blueprints/copy-read-blueprint.yaml
revit13: Thanks for the explanation! that makes sense. so we assume no module will have to check another module's status or this is a specific case? Thanks

manager/testdata/new-blueprints/write-module.yaml
froesef: Yes. But then the `api.endpoint` is the differentiating factor. For modules that configure a service it's set for others it's not. As we don't have a controller or webhooks we won't check if the module definition is valid anyways.

manager/testdata/new-blueprints/credit-card-detection-module.yaml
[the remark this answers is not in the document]
shlomitk1: @froesef @simanadler I thought the decision was that the audit capability will not be a module but rather an external component that will receive events from the manager or analyze m4dapp status. Otherwise denials will not be recorded.

manager/testdata/new-blueprints/network-policy-module.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: app.m4d.ibm.com/v1alpha2
+kind: M4DModule
+metadata:
+  name: network-policy
+  namespace: m4d-system
```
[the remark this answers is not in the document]
shlomitk1: @simanadler It is a namespace of m4dmodule, not of the deployed module resources.

manager/testdata/new-blueprints/read-module.yaml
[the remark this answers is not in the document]
shlomitk1: Are you assuming that only multi-user services (one per cluster) will have an endpoint? Otherwise, like in the current read scenarios, multiple services are deployed in the same cluster+namespace, thus manager calculates the endpoints given some module-defined prefix/suffix plus the release name in order to distinguish between different releases of the same service.

PR #682, opened 2021-07-05
Update website

ronenkat:
Add content to m4d homepage.
Preview https://60e550cfb91f3b00dd043495--mesh-for-data.netlify.app


site/overrides/home.html
```diff
@@ -262,9 +262,9 @@ <h2>
     <div class="feature-item">
         <a href="https://medium.com/mesh-for-data">
             <h2>
-            <center>Our blogs @ Medium.com</center>
+            <center>Read our blogs</center>
```
roee88, line 265: I favor just "Blog" but if not then is it "blogs" or "blog posts"?

roee88 (approved):
The squared blocks look clickable but they don't, can this be removed?

site/overrides/home.html
ronenkat: Addressed.

PR #683, opened 2021-07-06
Add issuer to vault configuraton for kubernetes version 1.21

### LINKED ISSUE #672: Python error: NoneType object is not subscriptable - on Kind v0.11.0
If I follow the steps in https://mesh-for-data.github.io/mesh-for-data/v0.2/get-started/quickstart/ and https://mesh-for-data.github.io/mesh-for-data/v0.2/samples/notebook/ using a Kind v0.11.0 cluster, the result is:
```
---------------------------------------------------------------------------
FlightServerError                         Traceback (most recent call last)
<ipython-input-2-affc70c65584> in <module>
     14 
     15 # Send request and fetch result as a pandas DataFrame
---> 16 info = client.get_flight_info(fl.FlightDescriptor.for_command(json.dumps(request)))
     17 reader: fl.FlightStreamReader = client.do_get(info.endpoints[0].ticket)
     18 df: pd.DataFrame = reader.read_pandas()

/opt/conda/lib/python3.9/site-packages/pyarrow/_flight.pyx in pyarrow._flight.FlightClient.get_flight_info()
/opt/conda/lib/python3.9/site-packages/pyarrow/_flight.pyx in pyarrow._flight.check_flight_status()
FlightServerError: 'NoneType' object is not subscriptable. Detail: Python exception: TypeError
```

revit13:
For the latest kubernetes version (1.21) issuer should be added when configuring Vault Kubernetes auth method:
https://www.hashicorp.com/blog/retrieve-hashicorp-vault-secrets-with-kubernetes-csi
https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-issuer-discovery

This PR adds the issuer based on the Kubernetes version.

Tested with kind v0.10.0 default k8s version v1.20.2 and kind V0.11.1 default k8s version  v1.21.1
Closes #672

### DIFF
--- a/charts/vault/env/dev/vault-single-cluster-values.yaml
+++ b/charts/vault/env/dev/vault-single-cluster-values.yaml
@@ -4,19 +4,40 @@ vault:
     volumes:
       - name: plugins
         emptyDir: {}
+      - name: poststartvolume
+        emptyDir: {}
+      - name: configvolume # this volume is read only
+        configMap:
+          name: vault-poststart-cm
+          defaultMode: 0744
     volumeMounts:
       - mountPath: /usr/local/libexec/vault
         name: plugins
+        readOnly: false
+      - mountPath: /opt/config
+        name: poststartvolume
+        readOnly: false
+      - mountPath: /etc/config/poststart.sh
+        subPath: poststart.sh
+        name: configvolume
         readOnly: false
     extraInitContainers:
       - name: kubesecrets-plugin
         image: ghcr.io/mesh-for-data/vault-plugin-secrets-kubernetes-reader:latest
         command: [/bin/sh, -ec]
         args:
-          - cp /vault-plugin-secrets-kubernetes-reader /usr/local/libexec/vault/vault-plugin-secrets-kubernetes-reader
+          - |
+            cp /vault-plugin-secrets-kubernetes-reader /usr/local/libexec/vault/vault-plugin-secrets-kubernetes-reader
+            cp /etc/config/poststart.sh /opt/config/poststart.sh
         volumeMounts:
         - name: plugins
           mountPath: /usr/local/libexec/vault
+          readOnly: false
+        - name: poststartvolume
+          mountPath: /opt/config
+          readOnly: false
+        - name: configvolume
+          mountPath: /etc/config
           readOnly: false
     extraEnvironmentVars:
       VAULT_API_ADDR: http://127.0.0.1:8200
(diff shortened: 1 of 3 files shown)


charts/vault/env/dev/vault-single-cluster-values.yaml
```diff
@@ -4,20 +4,41 @@ vault:
     volumes:
       - name: plugins
         emptyDir: {}
+      - name: poststartvolume
+        emptyDir: {}
+      - name: configvolume # this volume is read only
```
roee88, line 9: Then why mounting with `readOnly: false`?

charts/vault/env/dev/vault-single-cluster-values.yaml
```diff
@@ -27,28 +48,7 @@ vault:
     postStart:
       - "sh"
       - "-c"
-      # Sleep command is needed to avoid synchronization issues with the container pod. Please see
-      # https://kubernetes.io/docs/tasks/configure-pod-container/attach-handler-lifecycle-event/#discussion
-      # FIXME: Use a proper way to configure Vault after Vault start-up
```
roee88, line 32: Why did you remove this comment?

charts/vault/templates/post-pod-creation-cm.yaml
```diff
@@ -0,0 +1,34 @@
+{{- $major := .Capabilities.KubeVersion.Major -}}
+{{- $minor := .Capabilities.KubeVersion.Minor -}}
```
roee88, line 2: Use https://helm.sh/docs/chart_template_guide/function_list/#semantic-version-functions instead of custom logic.

hack/tools/vault_utils.sh
```diff
@@ -50,6 +49,7 @@ enable_k8s_auth() {
         kubectl get secret $2 -n $3 -o jsonpath="{.data['ca\.crt']}" | base64 --decode > tmp/ca.crt
 
         # Configure the k8s sa auth
+        # TODO: Add issuer for kubernetes versions greater than 1.20
```
roee88, line 52: Is that still a TODO? I thought that's what post-pod-creation-cm.yaml does

roee88 (changes requested):

charts/vault/env/dev/vault-single-cluster-values.yaml
revit13: apparently `readOnly: false` does not have an effect when loading from configMap (https://github.com/kubernetes/kubernetes/issues/62099)..

charts/vault/env/dev/vault-single-cluster-values.yaml
revit13: I put it back thanks

charts/vault/templates/post-pod-creation-cm.yaml
revit13: Fixed thanks

charts/vault/env/dev/vault-single-cluster-values.yaml
roee88: Still you explicitly specify it as false which is just confusing right?

hack/tools/vault_utils.sh
revit13: Yes it is still a TODO. we currently have two kinds of values.yaml for vault deployments:
1) vault-single-cluster-values.yaml - single cluster setup. used in quick start. it includes vault configuration of auth methods, policies and etc
2) plugin-secrets-kubernetes-reader-values.yaml - deployment of vault without vault configuration of auth methods, policies etc (only loading the plugin).  it is now used for testing including for multi-cluster setup. vault_utils.sh  script is used to configure vault after its deployment and thus the change in vault-single-cluster-values.yaml which is in this pr is not applied to this case.

charts/vault/templates/post-pod-creation-cm.yaml
```diff
@@ -0,0 +1,32 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  namespace: {{ .Release.Namespace }} 
+  name: vault-poststart-cm
+data:
+  poststart.sh: | 
+      # configuring kubernetes auth method and use it later to create roles.
+      vault auth enable kubernetes
+{{- if semverCompare "<1.21-0" .Capabilities.KubeVersion.GitVersion }}
```
roee88, line 10: https://github.com/helm/helm/blob/21251378b432fb629cb1a49af3892e20c1ec9624/pkg/chartutil/capabilities.go#L82

charts/vault/env/dev/vault-single-cluster-values.yaml
revit13: I removed it. Thanks

charts/vault/templates/post-pod-creation-cm.yaml
revit13: Done. Thanks

roee88 (approved):

PR #637, opened 2021-06-06
Add taxonomy schema validate option to the command line tool

ronenkat:
Add taxonomy schema validate option to the command line tool.
Add test cases for a valid and invalid taxonomy.

PR #688, opened 2021-07-07
Introduce a command line interface (currently for internal use)

### LINKED ISSUE #686: Introduce a command line interface (currently for internal use for taxonomy purposes)
Add a `cmd` folder with a `taxonomy` subcommand. 

This will later be extended to support more subcommands to root and to taxonomy, for example:
```bash
m4d taxonomy validate ...
```

roee88:
Adds a cmd folder with 

- A root command. 
- A `taxonomy` subcommand to root

Both without functionality,  subcommands to `taxonomy` to be added later, like `mesh-for-data taxonomy validate` in #637.

Closes #686


charts/vault/.gitignore
```diff
@@ -0,0 +1,2 @@
+charts
+Chart.lock
```
ronenkat, line 2: why is this needed?

roee88: Running make verify, test, etc. generated these files.

hack/make-rules/tools.mk
```diff
@@ -10,11 +10,6 @@ $(TOOLBIN)/controller-gen:
 	GOBIN=$(ABSTOOLBIN) GO111MODULE=on go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.5.0
 	$(call post-install-check)
 
-INSTALL_TOOLS += $(TOOLBIN)/crd-ref-docs
-$(TOOLBIN)/crd-ref-docs:
-	GOBIN=$(ABSTOOLBIN) GO111MODULE=on go get github.com/elastic/crd-ref-docs@v0.0.5
-	$(call post-install-check)
-
```
ronenkat, line 17: how is this related to the cmd line tool?

main.go
```diff
@@ -0,0 +1,22 @@
+/*
```
roee88, line 1: Wrong license header

ronenkat (commented):
where does it get built?

roee88:
> where does it get built?

It doesn't yet. Internal use only.

ronenkat:
> > where does it get built?
> 
> It doesn't yet. Internal use only.

even internal tools needs to get built. We can add in a follow-on PR.

ronenkat (approved):

roee88:
> even internal tools needs to get built. We can add in a follow-on PR.

There will be a go release workflow for the command line as part of the release workflow, just like our other go repos. It will be added later when the command line is useful for users.  Internal use means you can `go build` it or `go run` it directly.

PR #703, opened 2021-07-10
Fybrik application taxonomy - webhook validation 

eletonia:
This PR adds the following:

- a common validation function called `CheckTaxonomy` in the `pkg/taxonomy/validate` package which takes a resource JSON, taxonomy path string, and resourceName string as parameters and returns any validation errors
- taxonomy validation code in the Fybrik application webhook which reads from the taxonomy configMap and stops the creation of an invalid Fybrik application according to the schema defined
- application webhook test which checks whether the validation logic is correct
- necessary changes to integration testdata files to accommodate for the new application schema

Resolves tasks in #537


manager/controllers/app/m4dapplication_controller.go
```diff
@@ -60,6 +60,10 @@ func (r *M4DApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		log.V(0).Info("The reconciled object was not found")
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+
+	// Call validate update function from webhook (only if webhooks are disabled - check flag )
+	// ValidateUpdate()
+
```
ronenkat, line 66: Is this the right location in the reconcile to check?
We should be checking the taxonomy only if there a change in the spec part.
I.e., if we reconcile on error and we already checked the taxonomy once, no need to repeat.

ronenkat (commented):
can you add test code for the webhook validation?

eletonia:
@ronenkat All tests are passing. I think this PR is ready to merge

PR #711, opened 2021-07-12
Tool to generate taxonomy.json

roee88:
Implements a task from https://github.com/mesh-for-data/mesh-for-data/issues/693: `Tool to generate taxonomy.json from base and layers`. Refer to the issue and the attached design document for reference.

In draft mode for early review in parallel to review of the design.


pkg/taxonomy/compile/transform.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+		t.removeComplexValidation(schema.Items)
+		if schema.AdditionalProperties != nil {
+			t.removeComplexValidation(schema.AdditionalProperties.Schema)
+		}
+	}
+	return schema
+}
+
+// func (t *transformer) adjustAdditionalProperties(schema *model.SchemaRef) *model.SchemaRef {
```
rohithdv, line 179: do we need this ?

rohithdv (commented):
This is just a general suggestion. Is the code here fully compliant with [Openapi 3 specification](https://swagger.io/specification/) ? We have seen some open source code generators not fully compliant with Openapi 3 specs. For example do we support features like [inheritance and polymorphism](https://swagger.io/docs/specification/data-models/)  in our context (if it makes sense) or features like multiple file resolution (of $ref references).  If not, maybe we can highlight somewhere about the aspects of the specification not supported currently so that we can take it up as enhancement sometime later. Thanks.

rohithdv (approved):

roee88:
> This is just a general suggestion. Is the code here fully compliant with [Openapi 3 specification](https://swagger.io/specification/) ? We have seen some open source code generators not fully compliant with Openapi 3 specs. For example do we support features like [inheritance and polymorphism](https://swagger.io/docs/specification/data-models/) in our context (if it makes sense) or features like multiple file resolution (of $ref references). If not, maybe we can highlight somewhere about the aspects of the specification not supported currently so that we can take it up as enhancement sometime later. Thanks.

OpenAPI 3.x is too advanced for our needs and as you said many things don't really work with tools (code generation and CRDs are examples). That's why I wrote [HOWTO_SPEC.md](https://github.com/mesh-for-data/mesh-for-data/blob/master/config/taxonomy/HOWTO_SPEC.md). I think that with the exception of the use of oneOf for unions, all definitions adhere to HOWTO_SPEC.md. We should eventually write an easy guide that takes the essence of that and of the [design doc](https://docs.google.com/document/d/1J7f8co97C5AXtpbgUKlSPCDMU60lkYmQnDEbx-aKuOM) so it would be straight forward to write layers.

ronenkat:
@roee88 does this makes the taxonomy verification tool obsolete? do we still need it? the output taxonomy should be properly formatted.
Do we still need it for the inputs? base and layer? I think it is checked by the compile as it otherwise will fail.

ronenkat (approved):

roee88:
> @roee88 does this makes the taxonomy verification tool obsolete? do we still need it? the output taxonomy should be properly formatted.
> Do we still need it for the inputs? base and layer? I think it is checked by the compile as it otherwise will fail.

The taxonomy.json file is not guaranteed to be structural so validation is still needed. For example, someone can define a new connection type for Kafka that includes  properties in `allOf` that don't appear in `properties`.

roee88:
@ritwikc @rohithdv should I change the output JSON to be formatted with indents instead of being a single line?

ritwikc:
Yes. A formatted version of taxonomy.json could be easier to look at. @roee88

ritwikc (approved):

PR #717, opened 2021-07-18
Using cache restrictions to control list access

shlomitk1:

Fixes https://github.com/mesh-for-data/mesh-for-data/issues/671

- This PR adds restrictions on cache for allowing access to specified resources only in a specified namespace. For example, fybrikmodules, fybrikstorageaccounts, plotters can be accessed (e.g. using get, list) in fybrik-system namespace, whereas blueprints and associated resources (secrets, pods, batchtransfers, ...) can be accessed if they are deployed in fybrik-blueprints only.
- Added tests showing that any attempt to access a resource like secret or fybrikmodule in a different namespace will fail.
- Replaced cluster role bindings with role bindings, cluster roles with roles, when applicable.


manager/controllers/app/suite_test.go
```diff
@@ -79,9 +82,17 @@ var _ = BeforeSuite(func(done Done) {
 		k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
 		Expect(err).ToNot(HaveOccurred())
 	} else {
+		systemNamespaceSelector := fields.SelectorFromSet(fields.Set{"metadata.namespace": utils.GetSystemNamespace()})
+		workerNamespaceSelector := fields.SelectorFromSet(fields.Set{"metadata.namespace": "m4d-blueprints"})
```
froesef, line 86: Do we have a constant for `m4d-blueprints` somewhere that can be used? This would make future refactoring easier.

manager/main.go
```diff
@@ -52,15 +54,30 @@ func init() {
...
+		&appv1.Plotter{}:                {Field: systemNamespaceSelector},
+		&appv1.M4DModule{}:              {Field: systemNamespaceSelector},
+		&appv1.M4DStorageAccount{}:      {Field: systemNamespaceSelector},
+		&appv1.Blueprint{}:              {Field: workerNamespaceSelector},
+		&motionv1.BatchTransfer{}:       {Field: workerNamespaceSelector},
+		&motionv1.StreamTransfer{}:      {Field: workerNamespaceSelector},
+		&kbatch.Job{}:                   {Field: workerNamespaceSelector},
+		&corev1.Secret{}:                {Field: workerNamespaceSelector},
+		&corev1.Pod{}:                   {Field: workerNamespaceSelector},
+		&corev1.PersistentVolumeClaim{}: {Field: workerNamespaceSelector},
```
froesef, line 69: The motion controllers are also using `kbatch.CronJob` and `apps.Deployment`

shlomitk1: Right... Actually, how about creating a new manager for motion, namespace-scoped? We would need less resources to restrict (only M4Dxxx, Plotters, Blueprints) this way. I was thinking about running two manager processes in main.go - what would it require? Different ports, leader-election, metrics?

manager/controllers/app/suite_test.go
shlomitk1: Replaced with constant.

manager/main.go
shlomitk1: Added missing resources.

roee88 (approved):
Awesome 😎

I added one comment on admin-cr. Also we lost the support for referring to a secret from FybrikApplication because the manager won't be able to read those secrets. This is something we need to think about.

charts/fybrik/templates/fybrik-admin-cr.yaml
```diff
@@ -5,53 +5,6 @@ kind: ClusterRole
 metadata:
   name: {{ template "fybrik.fullname" . }}-admin-cr
 rules:
```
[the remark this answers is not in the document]
shlomitk1: fybrik-user role still exists, it was not a part of this PR.
The rest of the admin-cr went either to fybrik-system-role (all but blueprints) and blueprint-role (only blueprints). The reason for separation is a different namespace to bound to.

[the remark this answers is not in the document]
roee88: So an administrator will for now will need to have the exact privileges as the manager. Not sure if that's exactly what we want but ok for now.

shlomitk1:
> 
> 
> Awesome 😎
> 
> I added one comment on admin-cr. Also we lost the support for referring to a secret from FybrikApplication because the manager won't be able to read those secrets. This is something we need to think about.

The current code version does not refer to a secret but rather passes the vault path to it to the connectors. BTW, vault still has a cluster role+binding to read secrets from all namespaces.

roee88:
> > Awesome 😎
> > I added one comment on admin-cr. Also we lost the support for referring to a secret from FybrikApplication because the manager won't be able to read those secrets. This is something we need to think about.
> 
> The current code version does not refer to a secret but rather passes the vault path to it to the connectors. BTW, vault still has a cluster role+binding to read secrets from all namespaces.

I was referring to https://github.com/fybrik/fybrik/blob/bd78861d93d31a16f25b9412440c2cd9bd69b045/manager/apis/app/v1alpha1/fybrikapplication_types.go#L80

PR #729, opened 2021-07-21
Use different stream binary for StreamTransfer deployment

froesef:

### DIFF
--- a/manager/controllers/motion/streamtransfer_deployment.go
+++ b/manager/controllers/motion/streamtransfer_deployment.go
@@ -40,7 +40,7 @@ func (reconciler *StreamTransferReconciler) CreateDeployment(streamTransfer *mot
 			},
 		}},
 		{
-			Name: "configuration",
+			Name: motionv1.ConfigSecretVolumeName,
 			VolumeSource: v1.VolumeSource{
 				Secret: &v1.SecretVolumeSource{
 					SecretName: name,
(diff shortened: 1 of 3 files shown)


manager/controllers/motion/streamtransfer_deployment.go
```diff
@@ -100,8 +100,8 @@ func (reconciler *StreamTransferReconciler) CreateDeployment(streamTransfer *mot
 						Name:            "transfer",
 						Image:           streamTransfer.Spec.Image,
 						ImagePullPolicy: streamTransfer.Spec.ImagePullPolicy,
-						Args:            []string{"/etc/mover/conf.json"},
-						Command:         []string{"/streaming-entrypoint.sh"},
+						Args:            []string{motionv1.ConfigSecretMountPath + "/conf.json"},
+						Command:         []string{motionv1.StreamtransferBinary},
```
shlomitk1, line 104: In the previous version this was an entry point rather than the binary. Are they equivalent?

froesef: The mover has one image with an entrypoint called `/mover`. This is the default entrypoint for batch processing. For stream processing there is `/stream`. The `command` in K8s overwrites the used entrypoint of a container. 
The `/stream` executable in the mover is a bash script that calls the old `/streaming-entrypoint.sh` script. I thought it's easier to make it an executable with no file ending if other mover implementations were e.g. direct executables.

shlomitk1 (approved):

PR #734, opened 2021-07-28
Generate vanity import definitions for website

ronenkat:
Add vangen to hack/tools
Add vangen.json definition as an initial package list generation.

Signed-off-by: RONENKAT <ronenkat@il.ibm.com>


ronenkat:
functionality moved to a separate repo - vanity-imports

PR #737, opened 2021-07-29
Refactor implicit copy modules to include more configurable helm variables

froesef:
…a (to have the same defaults as the mover)

### DIFF
--- a/manager/apis/motion/v1alpha1/streamtransfer_webhook.go
+++ b/manager/apis/motion/v1alpha1/streamtransfer_webhook.go
@@ -69,7 +69,7 @@ func (r *StreamTransfer) Default() {
 	}
 
 	if r.Spec.ReadDataType == "" {
-		r.Spec.ReadDataType = ChangeData
+		r.Spec.ReadDataType = LogData
 	}
 
 	if r.Spec.WriteDataType == "" {
--- a/manager/apis/motion/v1alpha1/streamtransfer_webhook.go
+++ b/manager/apis/motion/v1alpha1/streamtransfer_webhook.go
@@ -69,7 +69,7 @@ func (r *StreamTransfer) Default() {
 	}
 
 	if r.Spec.ReadDataType == "" {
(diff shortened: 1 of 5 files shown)


modules/m4d-implicit-copy-stream/templates/streamtransfer.yaml
```diff
@@ -61,3 +61,6 @@ spec:
   imagePullPolicy: {{ .Values.imagePullPolicy | quote }}
   {{ end }}
   noFinalizer: {{ .Values.noFinalizer }}
+  readDataType: "logdata"
```
shlomitk1, line 64: LogData

modules/m4d-implicit-copy-stream/templates/streamtransfer.yaml
```diff
@@ -61,3 +61,6 @@ spec:
...
+  writeDataType: "logdata"
+  writeOperation: "append"
```
shlomitk1, line 66: Append

modules/m4d-implicit-copy-stream/values.yaml
```diff
@@ -35,3 +35,7 @@ copy:
 #        object_key: ""
 #    format: ""
 #    vault: {}
+
+readDataType: "ChangeData"
```
shlomitk1, line 39: Didn't you want to change to LogData?

froesef: Yes I thought about it and did another PR for the mover. The ChangeData option makes most sense here as a default.
With this configuration we would take Kafka data (which is always a key,value pair) and only take the value part and write this  (e.g. as JSON or parquet) to COS.

shlomitk1 (approved):

shlomitk1 (changes requested):
Please rebase. Some of the files have been renamed as a result of the project renaming.

shlomitk1 (approved):

PR #748, opened 2021-08-02
Fybrik application taxonomy - modifying for taxonomy generator

eletonia:

- [x] adds AppInfo and Interface to `base.yaml` and layers to validate Fybrik applications
- [x] moves the JSON files in `config/taxonomy/objects` to `charts/fybrik/files/taxonomy`
- [x] modifies Makefile and codegen spec in `config/taxonomy` to point to new path of taxonomy JSON files
- [x] modifies Fybrik application webhook and webhook test to read from new `fybrik_application.json` taxonomy file in ConfigMap
- [x] adds webhook test to validate against base taxonomy
- [x] modifies Reconciler to read from `fybrik_application.json` taxonomy file in ConfigMap
- [x] modifies Makefile to copy files in `charts/fybrik/files/taxonomy` to `manager/testdata/unittests/basetaxonomy` and generate enhanced taxonomy in `manager/testdata/unittests/enhancedtaxonomy` 
- [ ] adds Helm flag to set custom taxonomy.json file 

Completes some tasks in #752


config/taxonomy/base/base.yaml
```diff
@@ -16,4 +16,16 @@ definitions:
...
+  Dataformat:
+    type: string
+  Interface:
+    type: object
+    properties:
+      protocol:
+        $ref: "#/definitions/Protocol"
+      dataformat:
+        $ref: "#/definitions/Dataformat"
+    additionalProperties: false
```
ronenkat, line 31: AppInfo needs to be defined in the base.yaml as object similar to Tags

config/taxonomy/objects/application.json
```diff
@@ -0,0 +1,33 @@
...
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "$id": "application.values.schema.json",
+    "title": "Application Values Taxonomy",
+    "type": "object",
+    "description": "Values contributed to by Fybrik applications.",
+    "properties": {
+        "spec": {
+            "type": "object",
+            "properties": {
+                "appInfo": {"$ref": "taxonomy.json#/definitions/AppInfo"},
```
ronenkat, line 11: To make this work without the layers you need AppInfo in the base.yaml
When you generate the taxonomy.json with layers it will add checks for Intent and Role.

config/taxonomy/example/application/appinfo.yaml
```diff
@@ -0,0 +1,21 @@
+definitions:
+  AppInfo:
+    type: object
```
ronenkat, line 3: Type should be defined in the base, here we should have only properties.

ronenkat (changes requested):
@eletonia looks good.
I references a couple of things that needs to be changed.

config/taxonomy/base/base.yaml
eletonia: I thought AppInfo (role and intent) is not required for a Fybrik application, shouldn't it be an optional layer?

config/taxonomy/objects/application.json
eletonia: Yes this application.json is assuming that taxonomy.json is compiled with the AppInfo layer

config/taxonomy/base/base.yaml
ronenkat: AppInfo is required, but intent and role is not required.
See how fybrikApp CRD is defined.

config/taxonomy/objects/application.json
ronenkat: That is an issue for our out-of-the-box deployment that uses only the base yaml.

config/taxonomy/example/application/appinfo.yaml
ritwikc: Should AppInfo be a part of the base.yaml. Asking this because you are referring to appinfo in taxonomy.json and hence it should it not include appinfo ?

eletonia: I just added AppInfo to base.yaml

ronenkat (commented):
For testing purposes I suggest we have two test phases:
1. For the base taxonomy
2. base + layers
We will need to create the taxonomy.json twice, and run tests. For example, with the extra layers we can test invalidate interfaces, but cannot with the base layer only - Thanks for catching this @eletonia.
One way to do it is create a separate directory for base taxonomy testing and for extended taxonomy testing and placing the fybrik_application.json and taxonomy.json there.

charts/fybrik/files/taxonomy/fybrik_application.json
```diff
@@ -0,0 +1,33 @@
+{
```
ronenkat, line 1: This file is a duplicate of the files in taxonomy/objects - would be better to copy the files during a build process rather than have a duplicate copy.

eletonia: Ok I deleted the `config/taxonomy/objects` directory and moved all the files here

ronenkat (commented):
How different is the taxonomy.jason in testdata/basetaxonomy and testdata/enhencedtaxnonmy from the generated taxonomy using bases and examples?
Are there any differences? if no, can you think of a way for the files in  testdata/basetaxonomy to be generated during (or before) the test run? and not to have duplicated in the git? I am worried about their long-run maintenance.

eletonia:
> How different is the taxonomy.jason in testdata/basetaxonomy and testdata/enhencedtaxnonmy from the generated taxonomy using bases and examples?
> Are there any differences? if no, can you think of a way for the files in testdata/basetaxonomy to be generated during (or before) the test run? and not to have duplicated in the git? I am worried about their long-run maintenance.

There are a lot of differences.. [taxonomy.json in enhanced taxonomy](https://github.com/eletonia/the-mesh-for-data/blob/fybrikapp-taxonomy-compile/manager/testdata/unittests/enhancedtaxonomy/taxonomy.json) checks the values and pairs of `dataformat/protocol` and also checks the `intent` and `role` and makes sure it is one of the values defined in the enum. The [taxonomy.json in the base taxonomy](https://github.com/eletonia/the-mesh-for-data/blob/fybrikapp-taxonomy-compile/manager/testdata/unittests/basetaxonomy/taxonomy.json) does not check any values and seems unnecessary for a FybrikApplication since the FybrikApplication CRD checks everything that the base taxonomy checks (if `AppInfo` and `Interface` exist)

manager/Makefile
```diff
@@ -15,6 +15,11 @@ include $(ROOT_DIR)/hack/make-rules/verify.mk
 pre-test: generate manifests $(TOOLBIN)/etcd $(TOOLBIN)/kube-apiserver
 	mkdir /tmp/taxonomy || true                                                                                                                          
 	cp ../charts/fybrik/files/taxonomy/*.json /tmp/taxonomy/
+	mkdir testdata/unittests/basetaxonomy || true
+	mkdir testdata/unittests/enhancedtaxonomy || true
+	cp ../charts/fybrik/files/taxonomy/*.json testdata/unittests/basetaxonomy
+	cp ../charts/fybrik/files/taxonomy/*.json testdata/unittests/enhancedtaxonomy
```
ronenkat, line 21: Can you change `enhancedtaxonomy `to `sampletaxonomy` and add the samples for catalog and module when adding the layers?

eletonia: Ok I changed the `taxonomy compile` command to generate a `taxonomy.json` with the layers in `application`, `module`, and `catalog` and saved it in `testdata/unittests/sampletaxonomy`

ritwikc (approved):

ronenkat (approved):

PR #749, opened 2021-08-03
Workflow to auto-generate vanity imports HTML meta-tags

### LINKED ISSUE #746: Automate vanity import path generation
### Problem description: 

When a new go submodule is added to the project the go-import meta tags should be updated to reflect that in order to serve requests for the module.
Currently, the HTML meta tags are generated in https://github.com/fybrik/vanity-imports repository by triggering a workflow which uses a tool called vangen (https://github.com/leighmcculloch/vangen) where the packages paths are hard-coded in vangen.json .

We need to automate the generation of the modules list in vangen.json and trigger that upon a push to the go repos that use vanity import.

### Possible solutions:

Add a new workflow in every go repos in the organization that uses vanity import which will trigger a workflow in vanity-imports repo upon a push.
The workflow in the vanity-imports repo will run a script to auto-generate the modules list and run vangen tool to create the meta-tags for them based on the json file.  Lastly, pushing the meta-tags to site branch as is done now.

Alternatively, a new github workflow can be added to every go repos in the organization that uses vanity import which will use vangen tools in https://github.com/fybrik/vanity-imports repository.

revit13:
This PR adds a workflow to auto-generate  HTML meta-tags using the tools in https://github.com/fybrik/vanity-imports repository.
The workflow is triggered on every push to the master branch.
If the HTML meta-tags were changed it creates a PR against site branch.

Closes #746

### DIFF
--- a/.github/workflows/vanity-imports.yaml
+++ b/.github/workflows/vanity-imports.yaml
@@ -1,0 +1,48 @@
+name: Fybrik Golang Vanity Imports
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - '*'
+
+jobs:
+  updateVanityImports:
+    runs-on: [ubuntu-latest]
+    steps:
+      - name: Checkout vanity-imports repository
+        uses: actions/checkout@v2
+        with:
+          repository: fybrik/vanity-imports
+          path: vanity-imports
+      - name: Checkout master branch
+        uses: actions/checkout@v2
+        with:
+          path: fybrik-master-branch
+      - name: Checkout site branch
+        uses: actions/checkout@v2
+        with:
+          path: fybrik-site-branch
+          ref: site
+      - name: Generate vanity import meta-tags
+        run: |
+          cp vanity-imports/hack/tools/vangen.sh fybrik-master-branch
+          cd fybrik-master-branch; ./vangen.sh; cd ..
+          cp fybrik-master-branch/vangen.json vanity-imports/
+          make -C vanity-imports generate
+      - name: Copy vanity import meta-tags to fybrik-site-branch
+        run: |
+          rm -rf fybrik-site-branch/fybrik
+          cp -r vanity-imports/vangen/fybrik fybrik-site-branch
+      - name: Create Pull Request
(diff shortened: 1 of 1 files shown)


.github/workflows/vanity-imports.yaml
```diff
@@ -0,0 +1,48 @@
+name: Fybrik Golang Vanity Imports
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - '*'
```
roee88, line 8: Is it needed?

.github/workflows/vanity-imports.yaml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          cp vanity-imports/hack/tools/vangen.sh fybrik-master-branch
+          cd fybrik-master-branch; ./vangen.sh; cd ..
+          cp fybrik-master-branch/vangen.json vanity-imports/
+          make -C vanity-imports generate
+      - name: Copy vanity import meta-tags to fybrik-site-branch
+        run: |
+          rm -rf fybrik-site-branch/fybrik
+          cp -r vanity-imports/vangen/fybrik fybrik-site-branch
+      - name: Create Pull Request
```
roee88, line 38: Once we're sure it works then maybe commit and push directly?

roee88 (approved):
Great, please do the same for the other repos.

.github/workflows/vanity-imports.yaml
revit13: sure we can...

.github/workflows/vanity-imports.yaml
revit13: I am not sure... I saw it in other workflows...

roee88: Imports won't change between master and tags

.github/workflows/vanity-imports.yaml
roee88: So let's try with PR first and then see how it goes

revit13:
Taking the approach of adding the meta-tags to 404.html file

PR #750, opened 2021-08-03
Add 404.html and remove unneeded directories for vanity import paths.

revit13:
This PR removes unneeded directories that were used for vanity import paths and adds 404.html to handle the paths.


404.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
```
roee88, line 1: Try with a clone of https://github.com/fybrik/fybrik/blob/site/dev/404.html + meta tags at the top.

revit13: It seems to work so I used that.

roee88 (approved):
Links may not be rendered correctly because they are relative. Let's merge and check and submit a fix PR accordingly.

PR #775, opened 2021-08-12
Make blueprint namespace configurable

laurieaw63:
Allow deploying multiple instances of Fybrik to different namespaces in the same cluster.
* Add support for configuring the blueprint namespace name when deployed
* Add support for scoping to a application namespace for controlled environments
* Adjust pipeline tests so use system namespace setup by pipeline bootstrap

Co-authored-by: Nicholas Goracke ngoracke@us.ibm.com
Co-authored-by: Laurie Williams laurieaw@us.ibm.com
Signed-off-by: Laurie Williams laurieaw@us.ibm.com


charts/fybrik/templates/fybrik-admin-crb.yaml
```diff
@@ -3,7 +3,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: {{ template "fybrik.fullname" . }}-admin-crb
+  name: {{ .Release.Namespace }}-{{ template "fybrik.fullname" . }}-admin-crb
```
shlomitk1, line 6: Can a limit on the name length be exceeded?

charts/fybrik/templates/fybrik-admin-role.yaml
```diff
@@ -0,0 +1,121 @@
...
+  resources:
+  - blueprints/status
+  verbs:
+  - get
+  - patch
+  - update
+- apiGroups:
+  - app.fybrik.io
+  resources:
+  - fybrikapplications
```
shlomitk1, line 58: Are FybrikApplications created in the control-plane namespace?

laurieaw63: Yes can setup to create in the control-plane namespace or a specific namespace.

charts/fybrik/templates/fybrik-admin-crb.yaml
laurieaw63: Since we only create the clusterrolebinding when running cluster scoped = true. We could remove the namespace from the metadata.name if concerned about long namespace names.

simanadler:
> Allow deploying multiple instances of Fybrik to different namespaces in the same cluster.
@laurieaw63 I assume you mean FybrikApplication rather than multiple instances of the Fybrik control plane.
> 
>     * Add support for configuring the blueprint namespace name when deployed
@laurieaw63 Is this related to https://github.com/fybrik/fybrik/issues/793?
> 
>     * Add support for scoping to a application namespace for controlled environments
@laurieaw63 This assumes that FybrikApplication is trusted.  Correct?  How will we differentiate between a FybrikApplication that is trusted vs one that is not?  @roee88 Do we need to?

> 
>     * Adjust pipeline tests so use system namespace setup by pipeline bootstrap
> 
> 
> Co-authored-by: Nicholas Goracke [ngoracke@us.ibm.com](mailto:ngoracke@us.ibm.com)
> Co-authored-by: Laurie Williams [laurieaw@us.ibm.com](mailto:laurieaw@us.ibm.com)
> Signed-off-by: Laurie Williams [laurieaw@us.ibm.com](mailto:laurieaw@us.ibm.com)

manager/controllers/app/fybrikapplication_controller_test.go
```diff
@@ -75,10 +85,13 @@ var _ = Describe("FybrikApplication Controller", func() {
...
 			application := &apiv1alpha1.FybrikApplication{}
+
 			Expect(readObjectFromFile("../../testdata/e2e/fybrikapplication.yaml", application)).ToNot(HaveOccurred())
 			applicationKey := client.ObjectKeyFromObject(application)
-
+			fmt.Printf("Module:  %v\n", module.Namespace)
+			fmt.Printf("Application:  %v\n", application.Namespace)
 			// Create FybrikApplication and FybrikModule
 			Expect(k8sClient.Create(context.Background(), module)).Should(Succeed())
 			Expect(k8sClient.Create(context.Background(), application)).Should(Succeed())
```
simanadler, line 97: @laurieaw63 Perhaps include a test that tries to create a fybrikapplication from a namespace that's not allowed?

pipeline/README.md
```diff
@@ -4,20 +4,20 @@
 
 ## Bootstrapping
 
-Initial install for a cluster must happen in fybrik-system (currently).  It won't hurt anything if you aren't sure, and reinstall in fybrik-system.
-```
-. source-external.sh
-bash -x bootstrap-pipeline.sh fybrik-system
-# follow on screen instructions
-```
+Run the following command to bootstrap a Fybrik instance. By default the instance of Fybrik will be deployed as namespaced scoped, meaning the controller will only watch and process FybrikApplications from a specific namespace. Therefore limiting you to create FybrikApplications from one namespace. By that namespace is the one specific on the bootstrap script (a.k.a *system* namespace). You can change this behavior by setting environment variables. For details see [Scoping the deployment](#scoping-the-deployment).
```
simanadler, line 7: @roee88 Do we want this to be the default behavior for open source?

pipeline/README.md
```diff
@@ -4,20 +4,20 @@
...
 . source-external.sh
 bash -x bootstrap-pipeline.sh fybrik-myname
 # follow on screen instructions
 ```
 
+After running the "tkn" command that was returned from the bootstrap script, the following namespaces will be created. Illistrated by using the example *system* namespace of fybrik-myname:
+1. fybrik-myname - This will be the *system* namespace where Fybrik is deployed. By default the controller will only process FybrikApplications from this namespace.
+2. fybrik-myname-blueprints - this is the designated namespace for blueprints custom resources. This is also where the data access module is deployed to. 
+3. fybrik-myname-app - This namespace is only created if running namespace scoped AND the following environment variable is set: use_application_namespace=true.  
+
```
simanadler, line 20: @laurieaw63 @roee88 What about the namespaces for the data plane for each workload?  Should this be configurable as well?

pipeline/README.md
```diff
@@ -30,10 +30,35 @@ From command line, it can be done with this script:
...
+By default the bootstrap script will deploy Fybrik namespaced scoped. Deploying as namespaced scoped allows multiple instances of Fybrik to be installed in unique namespaces in the same cluster.  You can change this behavior but you must do so with care.
+
+### Environment variables
+| Name  | Default  | Valid Values | Description |
+|-------|----------|--------------|-------------|
+| cluster_scoped | false | true, false | Indicates if deploy as cluster scoped meaning the controller will watch and process FybrikApplications from all namespaces. For caveats see [tips](#tips) |
+| use_application_namespace | false | true, false | Specifies if you want the bootstrap script to create a namespace for FybrikApplications. This will be ignore if cluster_scoped=true. If set to false the *system* namespace will be used. |
+
+### Tips
```
simanadler, line 44: Great documentation!  Thanks @laurieaw63

pipeline/README.md
```diff
@@ -30,10 +30,35 @@ From command line, it can be done with this script:
...
+1. When deploying a Fybrik instance as cluster scoped - **Disclaimer: Only use cluster scope if you know what you’re doing.**
+   1. There can only be one instance of Fybrik deployed in cluster. 
+   2. The controller will be setup so it watches all namespaces for FybrikApplications.
+   3. You cannot run namespaced scoped and cluster scoped in same cluster.
+
+2. When deploying a Fybrik instance as namespaced scoped:
+   1. You can only create FybrikApplication custom resources from one namespace. The default will be to allow only creating from the *system* namespace. You can override this behavior by setting the use_application_namespace environment variable to true, the bootstrap code will create the *system*-app namespace and necessary objects to allow creating FybrikApplications. Hint: You will see the application namespace in the -p mesh-for-data-values= in the tkn command that is returned from running the bootstrap script.
+
+3. When using wkc-connector, the wkc-credentials secret must be in the same namespace you are using when creating FybrikApplication custom resources. The bootstrap script will unconditionally create a wkc-credentials secret in *system* namespace.
```
simanadler, line 53: @laurieaw63 @roee88 Should this be in the open source?

simanadler (commented):
Lots of great work was done here.  It would have been easier to review if it referred to an issue describing the problem being addressed.

simanadler (commented):
@laurieaw63 Please note that the sample GUI provided will also need to be changed.  See https://github.com/fybrik/fybrik/tree/master/samples/gui

charts/fybrik/templates/manager-auth-proxy-rb.yaml
```diff
@@ -0,0 +1,17 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ template "fybrik.fullname" . }}-proxy-role
+subjects:
+- kind: ServiceAccount
+  name: {{ .Values.manager.serviceAccount.name | default "default" }}
+  namespace: {{ .Release.Namespace }}
+{{- end }}
+{{- end }}
+
```
roee88, line 17: Did you verify that this works? I'm not sure the rbac proxy (used for access to metrics endpoint) works without a cluster role. One thing we can do for now is to introduce a separate flag for enabling/disabling the rbac proxy and just keep if disabled by default.

charts/fybrik/templates/manager-deployment.yaml
```diff
@@ -80,18 +80,35 @@ spec:
             {{- end }}
           env:
             - name: ENABLE_WEBHOOKS
+            {{- if .Values.clusterScoped }} 
               value: "true"
+            {{- else }}
+              value: "false"
+            {{- end }}
```
roee88, line 87: Can probably use `value: {{ .Values.clusterScoped | quote }}` or something similar.

manager/controllers/app/blueprint_controller_test.go
```diff
@@ -22,7 +23,7 @@ func deployBlueprint(namespace string, shouldSucceed bool) {
 
 	// Set the correct namespace
 	blueprint.SetNamespace(namespace)
-
+	fmt.Printf("deploy blueprint -  namespace: " + namespace + "\n")
```
roee88, line 26: Consider using format string %s instead of concat

manager/controllers/app/fybrikapplication_controller_test.go
```diff
@@ -5,28 +5,35 @@ package app
...
 	Context("FybrikApplication", func() {
+
+		controllerNamespace := getControllerNamespace()
+		fmt.Printf("FybrikApplication: controller namespace " + controllerNamespace)
+
 		BeforeEach(func() {
 			// Add any setup steps that needs to be executed before each test
 			module := &apiv1alpha1.FybrikModule{}
 			Expect(readObjectFromFile("../../testdata/e2e/module-read.yaml", module)).ToNot(HaveOccurred())
+      module.Namespace = controllerNamespace
```
roee88, line 34: Indent issue?

pipeline/README.md
roee88: This is the default for the pipeline installation. I think it's fine.

pipeline/README.md
roee88: @simanadler What do you mean? This PR does not introduce separate data plane namespace per fybrik application

pipeline/README.md
roee88: > @laurieaw63 @roee88 Should this be in the open source?

That's a question to @ronenkat

manager/main.go
```diff
@@ -56,9 +56,24 @@ func init() {
 func run(namespace string, metricsAddr string, enableLeaderElection bool,
 	enableApplicationController, enableBlueprintController, enablePlotterController, enableMotionController bool) int {
 	setupLog.Info("creating manager")
+
+	var applicationNamespaceSelector fields.Selector
+	applicationNamespace := os.Getenv("APPLICATION_NAMESPACE")
+	if len(applicationNamespace) > 0 {
```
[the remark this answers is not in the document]
laurieaw63: If it does not exist, the controller will accept FybrikApplications from any namespace.

manager/main.go
```diff
@@ -56,9 +56,24 @@ func init() {
...
+		applicationNamespaceSelector = fields.SelectorFromSet(fields.Set{"metadata.namespace": applicationNamespace})
+	}
+	setupLog.Info("Application namespace: " + applicationNamespace)
+
+	// Replace with the getBlueprintNamespace()
+	blueprintNamespace := os.Getenv("BLUEPRINT_NAMESPACE")
+	if len(blueprintNamespace) <= 0 {
+		blueprintNamespace = app.BlueprintNamespace
```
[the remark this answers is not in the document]
laurieaw63: app.BlueprintNamespace is a constant defined in Plotter it is set to the default blueprint namespace name: fybrik-blueprints.

pipeline/README.md
laurieaw63: @roee88 is correct this PR is only addressing allowing the name of the blueprint namespace to be configurable; as well as allowing Fybrik to be setup such that it only accepts FybrikApplications from a specific namespace.

manager/main.go
```diff
@@ -56,9 +56,24 @@ func init() {
...
+
+	var applicationNamespaceSelector fields.Selector
+	applicationNamespace := os.Getenv("APPLICATION_NAMESPACE")
+	if len(applicationNamespace) > 0 {
+		applicationNamespaceSelector = fields.SelectorFromSet(fields.Set{"metadata.namespace": applicationNamespace})
+	}
+	setupLog.Info("Application namespace: " + applicationNamespace)
+
+	// Replace with the getBlueprintNamespace()
+	blueprintNamespace := os.Getenv("BLUEPRINT_NAMESPACE")
```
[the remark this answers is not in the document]
laurieaw63: Yes will update

laurieaw63:
> 
> 
> @laurieaw63 Please note that the sample GUI provided will also need to be changed. See https://github.com/fybrik/fybrik/tree/master/samples/gui

What did you think the changes need to be? Fybrik can run with defaults as it does today.  It is only if want to configure the blueprint namespace to a different name or limit the FybrikApplication namespace.

pipeline/README.md
laurieaw63: By default pipeline will setup namespace scoped app controller.

charts/fybrik/templates/fybrik-blueprints-ns.yaml
```diff
@@ -1,8 +1,7 @@
 {{- if or .Values.coordinator.enabled .Values.worker.enabled }}
-{{- if .Values.clusterScoped }}
```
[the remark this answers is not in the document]
laurieaw63: The pipeline deployment will create the namespace for the blueprint when NOT running cluster scope.  At least this is the case for this PR. So this still means if clusterScoped is false then cluster scoped items must be pre-created and in addition the blueprint namespace needs to be pre-created.  Do you see that as an issue?

charts/fybrik/templates/manager-auth-proxy-rb.yaml
laurieaw63: I will add a separate flag, in some environments this option should be omitted.

charts/fybrik/templates/manager-deployment.yaml
laurieaw63: Unless you insist I am inclined to leave as is. Setting it explicitly simplifies the check in the go code.

manager/controllers/app/fybrikapplication_controller_test.go
laurieaw63: This will take some time to setup since tests needs to know when should fail or not. Would like to do this separate so can get this PR though.

manager/controllers/app/blueprint_controller_test.go
laurieaw63: Done

manager/controllers/app/fybrikapplication_controller_test.go
laurieaw63: Looks ok on my end

charts/fybrik/templates/fybrik-blueprints-ns.yaml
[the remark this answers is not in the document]
roee88: No but the change above makes the namespace be created even if cluster scoped is false. That's the issue.

charts/fybrik/templates/manager-auth-proxy-rb.yaml
roee88: I think that we can reuse the Prometheus flag, I don't think the rbac proxy is used for anything else (although I might be wrong). Yes, it makes sense to me to require both. If you have a counter suggestion please advise, I just want to be sure that whatever we do actually works.

charts/fybrik/templates/manager-deployment.yaml
roee88: I don't see how the go code is related but ok I don't really mind.

manager/controllers/app/fybrikapplication_controller_test.go
roee88: Unless I'm missing something I believe that it doesn't look right on GitHub view at least. There are also some lint issues detected by the CI checks. A formatter should solve these things automatically.

pipeline/README.md
laurieaw63: @ronenkat So are we talking only about #3. Or number 2 as well?

charts/fybrik/templates/fybrik-blueprints-ns.yaml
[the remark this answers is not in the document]
laurieaw63: I miss-stated in my comment above. I have changed the pipeline to create the blueprint namespace if does not already exist when cluster_scope is false. I added the cluster scope check back to the namespace yaml.

charts/fybrik/templates/manager-auth-proxy-rb.yaml
laurieaw63: I remove the role/rolebinding for the rbac proxy; and only enable it if clusterscoped and manager.prometheus  are both true.

charts/fybrik/templates/_helpers.tpl
```diff
@@ -99,3 +99,15 @@ certmanager.k8s.io/v1alpha1
 cert-manager.io/v1alpha2
 {{- end -}}
 {{- end -}}
+
+{{/*
+Get blueprints namespace
+*/}}
+{{- define "fybrik.getBlueprintNamespace" -}}
```
roee88, line 106: FYI there is a built-in `default` function to achieve the same

roee88 (approved):

roee88:
I mostly reviewed the Helm files. There are build errors and conflicts that need to be fixed before merge.

manager/controllers/app/fybrikapplication_controller_test.go
laurieaw63: Fixed

charts/fybrik/templates/_helpers.tpl
laurieaw63: leaving as is for now

pipeline/README.md
laurieaw63: Only removed #3.

laurieaw63:
Closing this PR. See https://github.com/fybrik/fybrik/pull/803

PR #778, opened 2021-08-15
Report status per asset

shlomitk1:
Fixes https://github.com/fybrik/fybrik/issues/640
Fixes https://github.com/fybrik/fybrik/issues/664

This PR modifies a status structure of FybrikApplication to support reporting status per asset.
Assets that got Deny do not prevent a blueprint from being constructed.
Data instructions are removed from the status and are not reported in GUI.

Not included:
- report status per asset in Plotter status, thus Ready will be reported when achieved for all orchestrated modules.
- data access instructions are not removed from blueprint+plotter resources and are collected by Blueprint controller from helm release notes.


manager/apis/app/v1alpha1/fybrikapplication_types.go
```diff
@@ -155,25 +155,34 @@ type DatasetDetails struct {
 	Details serde.Arbitrary `json:"details,omitempty"`
 }
 
-// FybrikApplicationStatus defines the observed state of FybrikApplication.
-type FybrikApplicationStatus struct {
+// AssetState defines the observed state of an asset
+type AssetState struct {
```
simanadler, line 159: @shlomitk1 Would this be per data asset or per data asset flow?  (read/write/delete).  Currently seems like we have mixed them all together in a single structure, and made assumptions about how the user is supposed to figure out which flow it's for.  Ex: If CatalogAsset is populated then we assume we wrote a new asset, so I assume there shouldn't be an endpoint ... and vice versa.  I think at a minimum we should be this logic in the comments OR declare it explicitly in the structure.

simanadler (commented):
I think it is important to keep the GUI up to date so we can continue to do demos.  You write that you didn't update it, but I see in the files that you did.  Is there part that you did not do?

shlomitk1:
> 
> 
> I think it is important to keep the GUI up to date so we can continue to do demos. You write that you didn't update it, but I see in the files that you did. Is there part that you did not do?

It is work in progress. Starting with changes on top of the existing PR.

manager/apis/app/v1alpha1/fybrikapplication_types.go
shlomitk1: It is the question of granularity. Suppose the workload reads A, reads B, updates A. The current approach indicates that A is ready if both read and write flows including A are deployed. Do we need to tell the user "you can read A but not write" or "A can be used"?

manager/apis/app/v1alpha1/fybrikapplication_types.go
```diff
@@ -155,25 +155,34 @@ type DatasetDetails struct {
...
 
-	// CatalogedAssets provide the new asset identifiers after being registered in the enterprise catalog
-	// It maps the original asset id to the cataloged asset id.
+	// ErrorMessage indicates that an error has happened during the reconcile, unrelated to a specific asset
 	// +optional
-	CatalogedAssets map[string]string `json:"catalogedAssets,omitempty"`
+	ErrorMessage string `json:"errorMessage,omitempty"`
+
+	// AssetStates provides a status per asset `json:"assetStates,omitempty"`
+	// +optional
```
revit13, line 184: minor comment: json:"assetStates,omitempty could be removed from the comment

shlomitk1: Thanks, removed.

manager/apis/app/v1alpha1/fybrikapplication_types.go
shlomitk1: @simanadler I understand it should be per data asset flow. We need to change the spec, and status will be changed accordingly. Will be handled in a future PR.

manager/apis/app/v1alpha1/fybrikapplication_types.go
```diff
@@ -155,25 +155,34 @@ type DatasetDetails struct {
...
+	// CatalogedAsset provides a new asset identifier after being registered in the enterprise catalog
+	// +optional
+	CatalogedAsset string `json:"catalogedAsset,omitempty"`
 
-	// Conditions represent the possible error and failure conditions
+	// Endpoint provides the endpoint spec from which the asset will be served to the application
 	// +optional
-	Conditions []Condition `json:"conditions,omitempty"`
+	Endpoint EndpointSpec `json:"endpoint,omitempty"`
+}
```
revit13, line 171: Shouldn't at least one of the AssetState fields will be required?

manager/apis/app/v1alpha1/fybrikapplication_types.go
```diff
@@ -155,25 +155,34 @@ type DatasetDetails struct {
...
 	// +optional
-	DataAccessInstructions string `json:"dataAccessInstructions,omitempty"`
+	Ready bool `json:"ready,omitempty"`
 
-	// CatalogedAssets provide the new asset identifiers after being registered in the enterprise catalog
-	// It maps the original asset id to the cataloged asset id.
+	// ErrorMessage indicates that an error has happened during the reconcile, unrelated to a specific asset
 	// +optional
-	CatalogedAssets map[string]string `json:"catalogedAssets,omitempty"`
+	ErrorMessage string `json:"errorMessage,omitempty"`
```
froesef, line 181: If there are multiple errors are those appended into the same string or will only one of the errors be displayed?

manager/apis/app/v1alpha1/fybrikapplication_types.go
shlomitk1: All status fields are optional.

revit13 (approved):

PR #783, opened 2021-08-19
Reduce reconcile time in blueprint deletion

shlomitk1:

Partially fixes https://github.com/fybrik/fybrik/issues/710

This PR reduces time required for a blueprint reconcile:
- No redundant calls for helm status are made. 
- After uninstall no checking is done to check whether uninstall is in progress.

### DIFF
--- a/manager/controllers/app/blueprint_controller.go
+++ b/manager/controllers/app/blueprint_controller.go
@@ -93,10 +93,6 @@ func (r *BlueprintReconciler) reconcileFinalizers(blueprint *app.Blueprint) (ctr
 			// the finalizer is present - delete the allocated resources
 			if err := r.deleteExternalResources(blueprint); err != nil {
 				r.Log.V(0).Info("Error while deleting owned resources: " + err.Error())
-				return ctrl.Result{}, err
-			}
-			if r.hasExternalResources(blueprint) {
-				return ctrl.Result{RequeueAfter: 2 * time.Second}, errors.NewPlain("helm release uninstall is still in progress")
 			}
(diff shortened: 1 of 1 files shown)


manager/controllers/app/blueprint_controller.go
```diff
@@ -93,10 +93,6 @@ func (r *BlueprintReconciler) reconcileFinalizers(blueprint *app.Blueprint) (ctr
 			// the finalizer is present - delete the allocated resources
 			if err := r.deleteExternalResources(blueprint); err != nil {
 				r.Log.V(0).Info("Error while deleting owned resources: " + err.Error())
-				return ctrl.Result{}, err
-			}
-			if r.hasExternalResources(blueprint) {
-				return ctrl.Result{RequeueAfter: 2 * time.Second}, errors.NewPlain("helm release uninstall is still in progress")
 			}
 			// remove the finalizer from the list and update it, because it needs to be deleted together with the object
 			ctrlutil.RemoveFinalizer(blueprint, finalizerName)
```
froesef, line 98: The behavior of the algorithm changes here. If errors occur during the uninstall the blueprint does no longer wait for it to recover but just logs an error in the blueprint controller and removes the finalizer which means that the blueprint is good for deletion. I'm ok with this as otherwise the blueprint may be indefinitely stuck. 
Please just confirm that this is the wanted behavior.

shlomitk1: This is the desired behavior.

froesef (approved):

PR #784, opened 2021-08-19
Fybrik Application taxonomy - fixing Reconciler validation 

### LINKED ISSUE #537: Fybrik Application Taxonomy

eletonia:

- [x] Adds 2 new fields to `FybrikApplication` Status: `ValidApplication` and `ValidatedGeneration`
- [x] Reconciler checks the `ValidatedGeneration` field to see if the application has already been validated
- [x] Reconciler sets `ValidApplication` to `true` or `false` depending on if the `FybrikApplication` passed the taxonomy check
- [x] Reconciler shows specific validation errors in the manager logs and `FybrikApplication` status
- [x] Modifies controller unit tests to update generation version
- [x] Checks `ENABLE_WEBHOOKS` flag to see if validation is needed in Reconciler

Fixes #537


ronenkat (commented):
Do we need two status indicators?
why not just `validatedGeneration` ? with the semantics of nil for never verified, a generation as the last one validated.
if current generation == `validatedGeneration` then we are good.
What is the down size for this approach? 
@shlomitk1 do you have a point of view?

shlomitk1:
> 
> 
> Do we need two status indicators?
> why not just `validatedGeneration` ? with the semantics of nil for never verified, a generation as the last one validated.
> if current generation == `validatedGeneration` then we are good.
> What is the down size for this approach?
> @shlomitk1 do you have a point of view?

The down size is that for an invalid application a new reconcile (for example, after a status update) will try to validate it again.  If this is the app without a plotter, the second reconcile will be the last one. If there is a plotter from a previous version, a reconcile will be called on each plotter update.

ronenkat:
> > Do we need two status indicators?
> > why not just `validatedGeneration` ? with the semantics of nil for never verified, a generation as the last one validated.
> > if current generation == `validatedGeneration` then we are good.
> > What is the down size for this approach?
> > @shlomitk1 do you have a point of view?
> 
> The down size is that for an invalid application a new reconcile (for example, after a status update) will try to validate it again. If this is the app without a plotter, the second reconcile will be the last one. If there is a plotter from a previous version, a reconcile will be called on each plotter update.

@shlomitk1, what you are saying is that any status update (triggered by a plotter) will advance the appVersion, triggering a revalidation --- this is the current code behaviour in the PR --- the validation is triggered by observedStatus.ValidatedGeneration != appVersion.

From looking at the PR the boolean indication is just for visibility to a user, and is not used as a control logic. @eletonia am I missing something in the logic?

eletonia:
> > > Do we need two status indicators?
> > > why not just `validatedGeneration` ? with the semantics of nil for never verified, a generation as the last one validated.
> > > if current generation == `validatedGeneration` then we are good.
> > > What is the down size for this approach?
> > > @shlomitk1 do you have a point of view?
> > 
> > 
> > The down size is that for an invalid application a new reconcile (for example, after a status update) will try to validate it again. If this is the app without a plotter, the second reconcile will be the last one. If there is a plotter from a previous version, a reconcile will be called on each plotter update.
> 
> @shlomitk1, what you are saying is that any status update (triggered by a plotter) will advance the appVersion, triggering a revalidation --- this is the current code behaviour in the PR --- the validation is triggered by `observedStatus.ValidatedGeneration != appVersion`.
> 
> From looking at the PR the boolean indication is just for visibility to a user, and is not used as a control logic. @eletonia am I missing something in the logic?

The latest commit changes the `ValidApplication` boolean to an enum with the options of `Unknown`, `True`, and `False`. This is now being used in the logic of the Reconciler as suggested by @shlomitk1

shlomitk1:
> @shlomitk1, what you are saying is that any status update (triggered by a plotter) will advance the appVersion, triggering a revalidation --- this is the current code behaviour in the PR --- the validation is triggered by `observedStatus.ValidatedGeneration != appVersion`.
> 
> From looking at the PR the boolean indication is just for visibility to a user, and is not used as a control logic. @eletonia am I missing something in the logic?

@ronenkat No, appVersion is not changed. Perhaps I have misunderstood your suggestion for ValidatedGeneration. If the suggestion is to update this field only for valid applications, then we'll get multiple validations for invalid applications as I mentioned. Otherwise, if every validation updates the field regardless of the result, then the next reconcile will not validate again but will not have enough information to know that the reconcile must not proceed. In other words, we need to know 1) when the validation was done to avoid extra validations and 2) validation result to avoid working with an invalid application.

Regarding the usage: if the application is invalid (regardless whether the validation was made in this reconcile or earlier, for the same generation), reconcile stops immediately.

manager/apis/app/v1alpha1/fybrikapplication_types.go
```diff
@@ -189,6 +197,14 @@ type FybrikApplicationStatus struct {
 	// +optional
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 
+	// ValidatedGeneration is the version of the FyrbikApplication that has been validated with the taxonomy defined.
```
shlomitk1, line 200: @eletonia Somehow after the merge the new structures appear twice.

eletonia: I just removed the duplicate variables in the latest commit

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -72,18 +73,27 @@ func (r *FybrikApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	observedStatus := applicationContext.Status.DeepCopy()
 	appVersion := applicationContext.GetGeneration()
 
-	// check if observed generation is not the same as appVersion
-	if observedStatus.ObservedGeneration != appVersion {
+	// check if webhooks are enabled and application has been validated before or if validated application is outdated
+	if os.Getenv("ENABLE_WEBHOOKS") == "false" && (string(applicationContext.Status.ValidApplication) == "" || observedStatus.ValidatedGeneration != appVersion) {
```
shlomitk1, line 77: Better make it` != "true"` to capture undefined variables as well

shlomitk1 (approved):
A minor comment regarding checking the environment variable.

manager/controllers/app/fybrikapplication_controller.go
eletonia: Updated in the latest commit

ronenkat (approved):

PR #785, opened 2021-08-19
Operate First - updating files, adding README, and creating GitHub workflow

eletonia:
- [ ]  Updates the Operate First script to put cluster-scoped resources into directories grouped by `apiVersion/kind`
- [x] Updates the generated yaml files given the latest Helm chart
- [x] Adds a README to document how to generate yaml files
- [x] Creates a GitHub workflow to automate the creation of a PR to the Operate First repo on each release of Fybrik


roee88 (changes requested):
The approach here won't work (if I got it right) and you should probably remove anything related to the charts repo and use the approach described in https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#push-pull-request-branches-to-a-fork instead. 

Also, I think it would be better to remove all yaml files from the source tree and only generate them on the fly in CI.

eletonia:
@roee88 I updated the GitHub workflow with the correct `APP_ID` and `APP_PRIVATE_KEY` and [tested it](https://github.com/eletonia/fybrik/runs/3559128596?check_suite_focus=true) in my master branch.

I am removing all the yaml files as a part of the workflow with `rm fybrik-repo/samples/operate-first/*.yaml`

PR #788, opened 2021-08-24
Make the concurrent reconciles and client qps and burst configurable …

### LINKED ISSUE #684: Make MaxConcurrentReconciles configurable
Currently all controllers use the default value for `MaxConcurrentReconciles` which is 1. 

To allow concurrency this should be configurable in the Helm chart for each controller (probably individually).
The value for `MaxConcurrentReconciles` can be set in `ctrl.NewControllerManagedBy` builders (like [the one for the M4DApplication controller](https://github.com/mesh-for-data/mesh-for-data/blob/5e853da67bf519bae30a74df9bcdc291716d36c5/manager/controllers/app/m4dapplication_controller.go#L464)) as an option.

In addition, the comment on the rest client configuration in https://github.com/kubernetes-sigs/controller-runtime/issues/1416#issuecomment-807872806 should also be taken into account.

froesef:
…for the manager

Fixes #684 

This PR allows to increase the limits of parallel reconciliation of objects as well as the client rates. 
The default client QPS is 5 and Burst is 10 which is fairly low when executing batch operations on thousands of objects.

### DIFF
--- a/manager/apis/motion/v1alpha1/streamtransfer_webhook.go
+++ b/manager/apis/motion/v1alpha1/streamtransfer_webhook.go
@@ -4,14 +4,13 @@
 package v1alpha1
 
 import (
-	"log"
-	"os"
-
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"log"
+	"os"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
--- a/manager/apis/motion/v1alpha1/streamtransfer_webhook.go
+++ b/manager/apis/motion/v1alpha1/streamtransfer_webhook.go
@@ -4,13 +4,14 @@
 package v1alpha1
 
 import (
+	"log"
+	"os"
+
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/validation/field"
-	"log"
-	"os"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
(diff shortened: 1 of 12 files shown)


manager/apis/motion/v1alpha1/streamtransfer_webhook.go
```diff
@@ -4,14 +4,13 @@
 package v1alpha1
 
 import (
-	"log"
```
roee88, line 7: Import order is inconsistent with the rest of the project

manager/controllers/app/blueprint_controller.go
```diff
@@ -319,7 +321,12 @@ func (r *BlueprintReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		},
 	}
 
+	numReconciles := utils.GetEnvAsInt("BLUEPRINT_CONCURRENT_RECONCILES", 5)
```
roee88, line 324: Do we have somewhere to put the defaults as constants?

manager/controllers/app/blueprint_controller.go
```diff
@@ -319,7 +321,12 @@ func (r *BlueprintReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	mgr.GetLogger().Info("Concurrent blueprint reconciles " + strconv.Itoa(numReconciles))
```
roee88, line 326: would Sprintf be cleaner?

manager/controllers/utils/utils.go
```diff
@@ -170,3 +172,27 @@ func GetModuleCapabilities(module *app.FybrikModule, requestedCapability app.Cap
 	}
 	return capFound, capList
 }
+
+// Returns the integer value of an environment variable.
+// If the environment variable is not set or cannot be parsed the default value is returned.
+func GetEnvAsInt(key string, defaultValue int) int {
```
roee88, line 178: Consider moving to a new package pkg/environment or something.

manager/main.go
```diff
@@ -72,7 +72,13 @@ func run(namespace string, metricsAddr string, enableLeaderElection bool,
 		&corev1.PersistentVolumeClaim{}: {Field: workerNamespaceSelector},
 	}
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	client := ctrl.GetConfigOrDie()
+	client.QPS = utils.GetEnvAsFloat32("CLIENT_QPS", 100.0)
```
roee88, line 76: Should we keep the defaults as the original? To avoid suggesting specific values

roee88 (commented):
Awesome 😎
Minor comments

roee88 (approved):

PR #804, opened 2021-09-10
Adding taxonomy layers for modules

eletonia:
This PR adds the taxonomy layers required to check `FybrikModules`


charts/fybrik/files/taxonomy/module.values.schema.json
```diff
@@ -1,100 +0,0 @@
-{
```
shlomitk1, line 1: This file is in use.

eletonia: Removed the test that was referencing this old schema

ronenkat (approved):
Thank you.
Let's wait for either @simanadler @shlomitk1 or @froesef to approve as well.

shlomitk1 (approved):

config/taxonomy/example/module/capability.yaml
```diff
@@ -0,0 +1,34 @@
+definitions:
```
roee88, line 1: Is that really just an example?

ronenkat: @eletonia suggest we move it to the base taxonomy. I expect that these fields are required.
I am not sure about the values of capabilities, maybe @froesef or @simanadler can comment if they can change or needs to be fixed.

eletonia: @roee88 @ronenkat Ok I deleted the `capability.yaml` example and added the `Capability` definition to `base.yaml`

charts/fybrik/files/taxonomy/fybrik_module.json
```diff
@@ -0,0 +1,20 @@
...
+    "type": "object",
+    "description": "Values contributed to by Fybrik modules.",
+    "properties": {
+        "spec": {
+            "type": "object",
+            "properties": {
+                "capabilities": {
+                    "type": "array",
+                    "items": {"$ref": "taxonomy.json#/definitions/Capability"}
+                }
```
simanadler, line 14: Shouldn't this include the plugin type as well?

config/taxonomy/base/base.yaml
```diff
@@ -31,3 +31,36 @@ definitions:
       dataformat:
         $ref: "#/definitions/Dataformat"
     additionalProperties: false
+  Source:
```
simanadler, line 34: Shouldn't the taxonomy include enums for the plugin types, dependency types, capability scope, and the supported actions?

simanadler, line 34: Shouldn't the taxonomy include enums for the plugin types, dependency types, capability scope, and the supported actions?

simanadler (commented):
There are more constants in the fybrikmodule that seem to be missing.

config/taxonomy/example/module/capability.yaml
```diff
@@ -3,8 +3,6 @@ definitions:
     properties:
       capability:
         enum: [copy, write, read, transform]
-      scope:
-        enum: [asset, workload, cluster]
```
shlomitk1: Why was this removed?

charts/fybrik/files/taxonomy/fybrik_module.json
ronenkat: @simanadler can you explain what should be defined in the CRD and what should be defined in the taxonomy.
Keep in mind that the taxonomy target cross components definitions., where is the plug-in used elsewhere?

config/taxonomy/base/base.yaml
ronenkat: @simanadler can you explain what should be defined in the CRD and what should be defined in the taxonomy.
Keep in mind that the taxonomy target cross components definitions. Are these definitions referenced beyond the module definition?
For actions, they are not part of the base, but a sample, as they are deployment specific.

config/taxonomy/example/module/capability.yaml
ronenkat: @shlomitk1 how is scope used?
Keep in mind that the taxonomy target cross components definitions., where is scope used elsewhere?

roee88: cross component + configurable at runtime

config/taxonomy/base/base.yaml
eletonia: These enums are defined in the CRD and are not configurable by the user so we decided not include them in the taxonomy

simanadler: @eletonia @roee88 @ronenkat The following are the constants in fybrikmodule_types.go and my thoughts about if/why they should be in the taxonomy:

CapabilityScope - (asset, workload, cluster) - If we think that these will not change or that we are happy with creating a new fybrik release if they change, then this can remain a constant and not go in the taxonomy.

DependencyType - same reasoning as CapabilityScope

CapabilityType (read, write, read, transform) - As new types of capabilities and modules are created additional values will likely need to be added.  Ex: isolation, data quality, logging, etc.  Thus, these should be in the taxonomy in my opinion.

PluginMechanism - should be in the taxonomy, so that new types of plugins  can be added without having to upgrade to a new fybrik version.

Actions - definitely need to be in the taxonomy.  If we provide out of the box actions in the quick start (or other pre-configured deployments), those actions should be included in the taxonomy provided with the quickstart.  Today the data movement operator is included in the fybrik repository, so shouldn't the actions it support be in the default taxonomy?

Data Protocols and Formats - don't know whether these should be in the data catalog or module taxonomy, but they need to be in one of them.  Again, those protocols and formats supported by the fybrik basic installation should be included in the taxonomy in my opinion.

Module Type - Currently a string whose documentation says it should be a service, config or plugin.  If we think that these will not change or that we are happy with creating a new fybrik release if they change, then I guess this doesn't need to go in the taxonomy.

simanadler (commented):
Responded to your questions about which constants I think should go into the taxonomy

config/taxonomy/base/base.yaml
roee88: @simanadler for everything you wrote as "if  happy to create a new release" I would go for excluding it from the taxonomy until we decide otherwise. 

With that in mind, I think that we're already aligned with the exception of CapabilityType which I'm unsure about (discussion needed; for this PR my suggestion was to exclude it and raise this in an open source meeting).

eletonia: @simanadler @roee88 Seems like the only fields we would potentially add to the current taxonomy are `PluginMechanism` and `CapabilityType`
I don't see `PluginMechanism` in any of the module specs - neither in the [arrow flight module](https://github.com/fybrik/arrow-flight-module/blob/master/module.yaml) nor the [hello world read module](https://github.com/fybrik/hello-world-read-module/blob/main/hello-world-read-module.yaml). 

Does this mean `PluginMechanism` is optional and not required? What are the possible values for `PluginMechanism`?

eletonia: @ronenkat Actually we already have `CapabilityType` in the taxonomy as an enum [here](https://github.com/fybrik/fybrik/blob/8d62b60521c9005b15ac067470e4172977528af0/config/taxonomy/base/base.yaml#L47) @roee88 Why do you think we should exclude it?

eletonia:
@ronenkat @roee88 @simanadler I made the changes to the module taxonomy that we discussed today. Can you review again?

rohithdv (commented):
This is just an observation. 
From my limited understanding, from our discussion earlier, I thought that there are currently only 2 fields which are taxonomy based : `Capability` and `pluginType`. At the moment in `fybrik_module.json` it looks like all the fields in this yaml are dependent on the `taxonomy.json` (which is dynamically generated based on base and layers combination). If this understanding is ok, then there is no issue in this file. 

Otherwise we may need to think if we can separate out some fields which are not dependent on `taxonomy.json` and put the entire definitions of those field in `fybrik_module.json `(instead of referring to `taxonomy.json`). For ex: you can check that there are some fields which are defined in the `policy_manager_request.json` like `PolicyManagerRequestAction` which have some fields dependent on `taxonomy.json` and some not dependent on `taxonomy.json`. If this is relevant in the fybrik module scenario we can do a similar thing. 

Basically I think we need to decide if we need to have "Interface", "Source" and "Sink" to be part of taxonomy or not. @roee88 @ronenkat.

charts/fybrik/files/taxonomy/fybrik_module.json
```diff
@@ -0,0 +1,65 @@
...
+                                    "properties": {
+                                        "id": {
+                                            "$ref": "taxonomy.json#/definitions/Action"
+                                        }
+                                    },
+                                    "required": ["id"]
+                                }
+                            }
+                        },
+                        "required": ["capability", "api", "supportedInterfaces", "actions"]
```
ronenkat, line 57: actions is optional, it should not be required, but should be checked if appears.

ronenkat:
> This is just an observation. From my limited understanding, from our discussion earlier, I thought that there are currently only 2 fields which are taxonomy based : `Capability` and `pluginType`. At the moment in `fybrik_module.json` it looks like all the fields in this yaml are dependent on the `taxonomy.json` (which is dynamically generated based on base and layers combination). If this understanding is ok, then there is no issue in this file.
> 
> Otherwise we may need to think if we can separate out some fields which are not dependent on `taxonomy.json` and put the entire definitions of those field in `fybrik_module.json `(instead of referring to `taxonomy.json`). For ex: you can check that there are some fields which are defined in the `policy_manager_request.json` like `PolicyManagerRequestAction` which have some fields dependent on `taxonomy.json` and some not dependent on `taxonomy.json`. If this is relevant in the fybrik module scenario we can do a similar thing.
> 
> Basically I think we need to decide if we need to have "Interface", "Source" and "Sink" to be part of taxonomy or not. @roee88 @ronenkat.

@rohithdv thank you.
What you are specifically saying that "Source" and "Sink" should move to `fybrik_module.json` as they are specific to the module definition. Sounds good to me.
Interface is part of FybrikApp so it should go into the `taxonomy.json`.
FYI @eletonia

eletonia:
Thanks for reviewing @rohithdv 
@ronenkat I moved `source` and `sink` from the taxonomy to `fybrik_module.json` and removed `actions` as a required field
Should be ready to merge now

ronenkat (approved):

rohithdv (approved):

charts/fybrik/files/taxonomy/fybrik_module.json
```diff
@@ -0,0 +1,70 @@
...
+                                    },
+                                    "required": ["id"]
+                                }
+                            }
+                        },
+                        "required": ["capability", "api", "supportedInterfaces"]
+                    }
+                }
+            },
+            "required": ["pluginType", "capabilities"]
```
roee88, line 66: pluginType is not a required field. See https://fybrik.io/dev/reference/crds/#fybrikmodulespec.

charts/fybrik/files/taxonomy/fybrik_module.json
```diff
@@ -0,0 +1,70 @@
...
+            "properties": {
+                "pluginType": {
+                    "$ref": "taxonomy.json#/definitions/PluginType"
+                },
+                "capabilities": {
+                    "type": "array",
+                    "items": {
+                        "type": "object",
+                        "properties": {
+		                    "capability": {
```
roee88, line 19: formatting: extra indent. Please use a formatter over the JSON files.

charts/fybrik/files/taxonomy/fybrik_module.json
```diff
@@ -0,0 +1,70 @@
...
+                                    "properties": {
+                                        "id": {
+                                            "$ref": "taxonomy.json#/definitions/Action"
+                                        }
+                                    },
+                                    "required": ["id"]
+                                }
+                            }
+                        },
+                        "required": ["capability", "api", "supportedInterfaces"]
```
roee88, line 62: Only capability is required. See https://fybrik.io/dev/reference/crds/#fybrikmodulespeccapabilitiesindex

charts/fybrik/files/taxonomy/fybrik_module.json
```diff
@@ -0,0 +1,70 @@
...
+                    "items": {
+                        "type": "object",
+                        "properties": {
+		                    "capability": {
+                                "$ref": "taxonomy.json#/definitions/Capability"
+                            },
+                            "api": {
+                                "type": "object",
+                                "properties": {
+                                    "interfaceDetails": {
```
roee88, line 25: There is no such property in `api`. Please see https://fybrik.io/dev/reference/crds/#fybrikmodulespeccapabilitiesindexapi. 

That is `api` itself is a variant of Interface that also includes endpoint as a required field.

charts/fybrik/files/taxonomy/fybrik_module.json
```diff
@@ -0,0 +1,70 @@
...
+                                            "source": {
+                                                "$ref": "taxonomy.json#/definitions/Interface"
+                                            }
+                                        },
+                                        {
+                                            "sink": {
+                                                "$ref": "taxonomy.json#/definitions/Interface"
+                                            }
+                                        }
+                                    ]
```
roee88, line 46: I don't think that it's oneOf. See for example https://github.com/fybrik/fybrik/blob/6138f68ad33ffc6f9179f58f29bf33b804939437/modules/implicit-copy-stream-module.yaml#L17-L22. 

@shlomitk1 why is supportedInterfaces defined as an object array in the CRD? What's the rational behind not making it a concrete type?

charts/fybrik/files/taxonomy/fybrik_module.json
```diff
@@ -0,0 +1,70 @@
...
+                                }
+                            },
+                            "actions": {
+                                "type": "array",
+                                "items": {
+                                    "type": "object",
+                                    "properties": {
+                                        "id": {
+                                            "$ref": "taxonomy.json#/definitions/Action"
```
roee88, line 55: `taxonomy.json#/definitions/Action` is an object with a required `name` field (and optionally additional fields). It's not the equivalent of `id`. I think that we must change the FybrikModule definition to match the taxonomy Action structure (instead of ID and level).

config/taxonomy/base/base.yaml
```diff
@@ -31,3 +31,7 @@ definitions:
       dataformat:
         $ref: "#/definitions/Dataformat"
     additionalProperties: false
+  Capability:
+    type: string
+  PluginType:
+    type: string
```
roee88, line 37: Missing empty end of line. Please run a formatter on YAML files too.

roee88 (changes requested):

charts/fybrik/files/taxonomy/fybrik_module.json
shlomitk1: @roee88 Not sure I understand. What is the alternative - to create an intermediate type that holds the same list of <source,sink> or changing the structure entirely?

roee88: > to create an intermediate type that holds the same list of <source,sink>

For example. Would that be ok?

shlomitk1: Yes.

charts/fybrik/files/taxonomy/fybrik_module.json
eletonia: @roee88 @ronenkat So we need to change the FybrikModule CRD structure?

PR #807, opened 2021-09-14
Policy compiler interface changes in manager to use policy manager openapi taxonomy

rohithdv:
This PR does porting from grpc objects to openapi objects in policy_compiler_interface.go file and related files. 


manager/controllers/app/policy_compiler_interface.go
```diff
@@ -6,75 +6,58 @@ package app
...
-						message = app.WriteNotAllowed
-					}
-					return actions, errors.New(message)
-				}
-				// Check if this is a real action (i.e. not Allow)
-				if utils.IsAction(action.GetName()) {
-					actions = append(actions, action)
-				}
-			}
-		}
```
shlomitk1, line 77: Why has this code been removed? 
1. In case of Deny it returns a corresponding error message
2. Allow actions are not appended

Not sure about Allow, perhaps it is not being returned at all, but indicating an error as a result of governance decision is essential.

manager/controllers/app/policy_compiler_interface.go
```diff
@@ -6,75 +6,58 @@ package app
...
-		CredentialPath: credentialPath,
-		Datasets: []*pb.DatasetContext{{
-			Dataset: &pb.DatasetIdentifier{
-				DatasetId: datasetID,
-			},
-			Operation: operation,
-		}},
-	}
+
+	return &req, credentialPath, nil
```
shlomitk1, line 47: An error field is always nil. Consider to remove.

manager/controllers/app/policy_compiler_interface.go
```diff
@@ -6,75 +6,58 @@ package app
...
+	reqContext := make(map[string]interface{})
+	for k, v := range input.Spec.AppInfo {
+		reqContext[k] = v
+	}
+	req.SetContext(reqContext)
+
 	var credentialPath string
 	if input.Spec.SecretRef != "" {
 		credentialPath = utils.GetVaultAddress() + vault.PathForReadingKubeSecret(input.Namespace, input.Spec.SecretRef)
 	}
```
shlomitk1, line 45: credentialPath should be constructed outside this function, it should not be returned by this function as well.

manager/controllers/app/policy_compiler_interface.go
rohithdv: Thanks @shlomitk1 . I have now changed the relevant part. The unit tests are passing. However, we may still need some changes to remove some hardcoding of action names and ids (in the new taxonomy definitions there is nothing like action ID only action name is provided ) in the code. This is similar to the comment (https://github.com/fybrik/fybrik/pull/808#discussion_r708532214)  in #808 . @ritwikc @roee88 .

manager/controllers/app/policy_compiler_interface.go
rohithdv: fixed. Thanks.

manager/controllers/app/policy_compiler_interface.go
rohithdv: fixed. Thanks.

manager/controllers/app/moduleinstance.go
```diff
@@ -134,18 +135,20 @@ func (m *ModuleManager) selectReadModule(item modules.DataInfo, appContext *app.
 	m.Log.Info("Select read path for " + item.Context.DataSetID)
 
 	// Read policies for data that is processed in the workload geography
-	var readActions []*pb.EnforcementAction
+	var readActions []*openapiclientmodels.ResultItem
 	var err error
-	readActions, err = LookupPolicyDecisions(item.Context.DataSetID, m.PolicyManager, appContext,
-		&pb.AccessOperation{Type: pb.AccessOperation_READ, Destination: m.WorkloadGeography})
+	reqAction := new(openapiclientmodels.PolicyManagerRequestAction)
```
roee88, line 140: We typically don't use `new` in our codebase. Please replace with something like:
`reqAction := openapiclientmodels.PolicyManagerRequestAction{}`

manager/controllers/app/moduleinstance.go
```diff
@@ -16,6 +16,7 @@ import (
 	local "fybrik.io/fybrik/pkg/multicluster/local"
 	"fybrik.io/fybrik/pkg/serde"
 	"fybrik.io/fybrik/pkg/storage"
+	openapiclientmodels "fybrik.io/fybrik/pkg/taxonomy/model/base"
```
roee88, line 19: I think that we need a better name than `openapiclientmodels`. Suggestions? :)

manager/controllers/app/moduleinstance.go
```diff
@@ -134,18 +135,20 @@ func (m *ModuleManager) selectReadModule(item modules.DataInfo, appContext *app.
...
+	readActions, err = LookupPolicyDecisions(item.Context.DataSetID, m.PolicyManager, appContext, reqAction)
 	if err != nil {
 		return nil, err
 	}
 	// select a read module that supports user interface requirements
 	// actions are not checked since they are not necessarily done by the read module
 	readSelector := &modules.Selector{Capability: app.Read,
 		Destination:  &item.Context.Requirements.Interface,
-		Actions:      []*pb.EnforcementAction{},
+		Actions:      []*openapiclientmodels.ResultItem{},
```
roee88, line 151: We probably need to change the base taxonomy so `ResultItem` would have another name with more context. Like PolicyManagerResultItem

manager/controllers/app/moduleinstance.go
```diff
@@ -404,19 +407,25 @@ func (m *ModuleManager) getCopyRequirements(item modules.DataInfo, readSelector
...
 }
 
-func (m *ModuleManager) enforceWritePolicies(appContext *app.FybrikApplication, datasetID string) ([]*pb.EnforcementAction, string, error) {
+func (m *ModuleManager) enforceWritePolicies(appContext *app.FybrikApplication, datasetID string) ([]*openapiclientmodels.ResultItem, string, error) {
 	var err error
-	actions := []*pb.EnforcementAction{}
+	actions := []*openapiclientmodels.ResultItem{}
 	//	if the cluster selector is non-empty, the write will be done to the specified geography if possible
 	if m.WorkloadGeography != "" {
+		reqAction := new(openapiclientmodels.PolicyManagerRequestAction)
```
roee88, line 415: same comment on `new`

manager/controllers/app/modules/moduleinterface.go
```diff
@@ -113,14 +114,14 @@ func (m *Selector) SupportsGovernanceActions(module *app.FybrikModule, actions [
...
-func (m *Selector) SupportsGovernanceAction(module *app.FybrikModule, action *pb.EnforcementAction) bool {
+func (m *Selector) SupportsGovernanceAction(module *app.FybrikModule, action *openapiclientmodels.ResultItem) bool {
 	// Check if the module supports the capability
 	if hasCapability, caps := utils.GetModuleCapabilities(module, m.Capability); hasCapability {
 		// There could be multiple structures for the same CapabilityType
 		for _, cap := range caps {
 			// Loop over the data transforms (actions) performed by the module for this capability
 			for _, act := range cap.Actions {
-				if act.ID == action.Id && act.Level == action.Level {
+				if act.ID == (action.GetAction().Name + "-ID") {
```
roee88, line 124: Why do we need this `-ID` thing?

manager/controllers/app/policy_compiler_interface.go
```diff
@@ -10,70 +10,74 @@ import (
...
-		},
-		CredentialPath: credentialPath,
-		Datasets: []*pb.DatasetContext{{
-			Dataset: &pb.DatasetIdentifier{
-				DatasetId: datasetID,
-			},
-			Operation: operation,
-		}},
+	if operationType == openapiclientmodels.WRITE {
+		action.SetActionType(openapiclientmodels.WRITE)
```
roee88, line 33: Why do we need the if conditions and not just set operationType

manager/controllers/app/policy_compiler_interface.go
roee88: @rohithdv  Are we expecting Allow actions to be a thing in the new version?

manager/controllers/app/moduleinstance.go
rohithdv: can we do this refactoring in another PR as it requires changes in many other places ? We can make a PR for this change and also any other changes to the taxonomy. Thanks.

manager/controllers/app/moduleinstance.go
rohithdv: Same as above. We can make this change in the refactoring PR once the end to end flow is tested.

manager/controllers/app/modules/moduleinterface.go
rohithdv: Earlier manager was comparing actions with action ID. In the current taxonomy changes, we donot provide provision for action id. There is only action name. We may need to change manager test case configurations to remove dependency on ID. @shlomitk1 can you suggest what could be the changes involved? I can do that in this PR itself if it is not too complex. I just want to cross verify once with you before I do the changes. Thanks.

manager/controllers/app/policy_compiler_interface.go
rohithdv: fixed. Thanks.

manager/controllers/app/moduleinstance.go
rohithdv: fixed. Thanks.

manager/controllers/app/moduleinstance.go
rohithdv: fixed. Thanks.

manager/controllers/app/policy_compiler_interface.go
rohithdv: Right now in #808 we have a rule which configures "Deny by default" (https://github.com/ritwikc/the-mesh-for-data/blob/2d010cd398af5bc88fd6f9d8dbd974b948283e64/third_party/opa/data-and-policies/user-created-policy-3/sample_policies.rego#L5). If that is the case then whenever this rule is loaded, there is no question of Allow. Either it is Redact or Deny.

roee88: The idea following previous discussions on "by default" was that it's allow unless otherwise specified. A user that wants to switch to deny by default would add the rule as in our example. 

We can have the "by default" policy loaded at installation time directly so there won't be a risk of synchronization delay.

manager/controllers/app/moduleinstance.go
roee88: Sure

manager/controllers/app/moduleinstance.go
roee88: Sure

manager/controllers/app/modules/moduleinterface.go
rohithdv: @roee88 @simanadler Based on discussion with @shlomitk1 , I have updated the mock policy manager and unit test configurations to avoid concatenation with "-ID" string. The hardcoding is now removed. 

However, I think there should be a separate issue (as @shlomitk1 suggested) to update the [ModuleCapability struct](https://github.com/rohithdv/fybrik/blob/247b95cf32adc1d712c75305f01ffc1c9ab3ef52/manager/apis/app/v1alpha1/fybrikmodule_types.go#L149). The `SupportedAction`  class used in `ModuleCapability struct` should be replaced by the [action taxonomy model struct](https://github.com/rohithdv/fybrik/blob/taxonomy_manager_changes/pkg/taxonomy/model/base/model_action.go). This is the larger issue which deals with how the action taxonomy will be used by the manager and the modules. Hence it would not be in the scope of this PR in my opinion.

manager/controllers/app/policy_compiler_interface.go
rohithdv: just to clarify, we should have an "Allow" action defined in the base taxonomy which is used by default ? Also should the hardcoding in the code be changed?

roee88: I was thinking that there will not be an allow action. If there are no actions in the result and 
there is no error, then it's interpreted as allow.

rohithdv: Updated [evaluate_opa.go](https://github.com/rohithdv/fybrik/blob/e7f8d19d2e18bdde0e99e55de19c2529cd2a7555/connectors/open_policy_agent/lib/evaluate_opa.go#L91) and [mockup policymanager.go](https://github.com/rohithdv/fybrik/blob/e7f8d19d2e18bdde0e99e55de19c2529cd2a7555/manager/controllers/mockup/policymanager.go#L39) to return empty result in case of Allow. Please check. Thanks. @roee88

roee88 (approved):

manager/controllers/app/policy_compiler_interface.go
```diff
@@ -10,70 +10,67 @@ import (
...
+			switch *openapiReq.GetAction().ActionType {
+			case openapiclientmodels.READ:
+				message = app.ReadAccessDenied
+			case openapiclientmodels.WRITE:
+				message = app.WriteNotAllowed
 			}
+			return actions, errors.New(message)
+		}
+		// Check if this is a real action (i.e. not Allow)
+		if utils.IsAction(result[i].GetAction().Name) {
```
shlomitk1, line 72: The condition is no longer required since we don't have Allow actions.

shlomitk1 (approved):

manager/controllers/app/policy_compiler_interface.go
rohithdv: fixed. Please check. Thanks.

PR #809, opened 2021-09-15
Operate First - updating script for cluster-scoped resources

eletonia:
- [x] Updates the Operate First script to generate cluster-scoped resource YAMLs in directories grouped by `apiVersion/kind` as requested by RedHat [here](https://github.com/operate-first/support/issues/370#issuecomment-917010573)
- [x] Updates Makefile with a new clean target which removes old generated YAML files and directories
- [x] Removes namespaces from this script since these should be generated with Operate First's CLI tool
- [x] Updates the generated yaml files given the latest Helm chart
- [x] Adds a README to document how to generate yaml files


roee88:
@ronenkat are you planning to review this PR?

PR #819, opened 2021-09-27
Use Role in katalog RBAC if clusterScoped=false

roee88:
**Changes in this PR:**

1. Katalog RBAC is using Role/RoleBinding when clusterScoped=false. Those RBAC resources are applied to the application namespace (or if unset, to the system namespace). 
2. The RBAC for manager access to FybrikApplication resources in the application namespace is also created with the chart. Previously a `*` role was applied in the pipeline boostrap script (cc @laurieaw63)

**Why:** 

We missed adjusting the permissions of katalog. This PR changes the katalog RBAC to be scoped to the application namespace when  installed with `--set clusterScoped=false`. For consistency all RBAC resources that should be applied to the applications namespace is now set from the chart itself.

**Usage:**

After this PR the following is required for a working control plane when using `clusterScoped=false` and `applicationNamespace=fybrik-applications`:

```bash
kubectl create ns fybrik-system
kubectl create ns fybrik-blueprints
kubectl create ns fybrik-applications
helm install fybrik-crd charts/fybrik-crd
helm install fybrik charts/fybrik --namespace fybrik-system --set clusterScoped=false --set applicationNamespace=fybrik-applications --set global.tag=master --set global.imagePullPolicy=Always
```

For proper support we need the installation and code to support multi-tenancy. A discussion/issue will be opened separately.

### DIFF
--- a/charts/fybrik/templates/fybrik-applications-rb.yaml
+++ b/charts/fybrik/templates/fybrik-applications-rb.yaml
@@ -1,0 +1,19 @@
+{{- if .Values.manager.enabled }}
+{{- if .Values.applicationNamespace}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ template "fybrik.fullname" . }}-applications-rb
+  namespace: {{ .Values.applicationNamespace }}
+  labels:
+    {{- include "fybrik.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ template "fybrik.fullname" . }}-applications-role
+subjects:
+- kind: ServiceAccount
+  name: {{ .Values.manager.serviceAccount.name | default "default" }}
+  namespace: {{ .Release.Namespace }}
+{{- end }}
+{{- end }}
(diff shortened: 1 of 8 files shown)


roee88:
Change required: use applicationNamespace

charts/fybrik/templates/katalog-connector/katalog-connector-rbac.yaml
```diff
@@ -1,18 +1,25 @@
...
 {{- if include "fybrik.isEnabled" (tuple .Values.katalogConnector.enabled $autoFlag) }}
-{{- if .Values.clusterScoped }}
 # Grant katalog-connector the katalog-editor Role.
-kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
+{{- if .Values.clusterScoped }}
+kind: ClusterRoleBinding
 metadata:
-  name:  katalog-connector-editor
+  name:  katalog-connector-viewer
```
revit13, line 8: samples/operate-first/ also should be changed

roee88: Thanks. Since samples/operate-first is already outdated I'll keep it off this PR (almost all CRDs are old). A separate PR to re-generate can be made by @eletonia if this will be required but currently it seems like we will only need CRDs as cluster scoped resources in operate-first.

charts/fybrik/templates/fybrik-applications-rb.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if .Values.manager.enabled }}
+{{- if .Values.applicationNamespace}}
```
laurieaw63, line 2: I like this change and seems like it should work. Does require the application namespace to exist prior to installing Fybrik.  Bootstrap will do this when deploying using that tooling. If deploying manually should be documented.

PR #822, opened 2021-09-28
add an option to install charts from a locally mounted volume in addi…

ngoracke:
…tion to oci image

Signed-off-by: Nicholas Goracke <ngoracke@us.ibm.com>

### DIFF
--- a/charts/fybrik/values.yaml
+++ b/charts/fybrik/values.yaml
@@ -171,6 +171,8 @@ manager:
 
   prometheus: false
 
+  chartVolume: ""
+
 # OPA connector component
 opaConnector:
   # Set to true to deploy the opa connector or false to skip its deployment.
--- a/charts/fybrik/values.yaml
+++ b/charts/fybrik/values.yaml
@@ -172,6 +172,11 @@ manager:
   prometheus: false
 
   chartVolume: ""
+    # Specify a persistent volume claim, to be mounted by the manager, that can contain
+    # helm charts that can be referenced by a FybrikModule.  Manager will check if a chart
+    # is available on volume first, then try to pull from a Docker registry if it does not exist.
+    # To populate the volume, create a 'charts' directory at the root of the mount,
+    # and place helm charts within the charts directory.
 
 # OPA connector component
 opaConnector:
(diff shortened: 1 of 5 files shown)


ngoracke:
We need the ability to install charts that are not available in a non-authenticated registry.  This gives the option to provide a PVC populated with charts.

roee88:
If there was support for authenticated registries would you still need this?

ngoracke:
@roee88 This is still quite convenient for development - it allows us to dump charts on volume, tweak as needed, and submit fybrik applications to exercise without publishing to an authenticated registry each iteration.

So I would say: yes.

ngoracke:
This *may* also be useful for air-gapped scenarios, where a registry that supports helm-charts-as-images is not available, and there is no desire to set up a separate helm chart repository.  That's debate-able, though.

charts/fybrik/values.yaml
```diff
@@ -171,6 +171,8 @@ manager:
 
   prometheus: false
 
+  chartVolume: ""
```
roee88, line 174: Consider adding a comment on how to use it.

roee88 (approved):

roee88:
> @roee88 This is still quite convenient for development - it allows us to dump charts on volume, tweak as needed, and submit fybrik applications to exercise without publishing to an authenticated registry each iteration.
> 
> So I would say: yes.

I think that an in cluster registry would be a better approach, I believe that we have used it successfully it in the past. Anyway, this PR looks good for me as a short term approach. The issue for adding authentication support is still open as a longer term goal. 

+ @revit13 for awareness because these changes need to be considered when upgrading to Helm 3.7.

charts/fybrik/values.yaml
ngoracke: Done

ngoracke:
Just a bit more detail on the last comment... We had done some experimentation with an in-cluster and out-of-cluster registry, but had issues with: https://github.com/helm/helm/issues/6324

Summary: the default communication protocol is https when not working with localhost:5000, and the operations we require don't support the --insecure options.  I believe we are getting around this with the KIND environment because we have configuration to route kube traffic to localhost:5000 to kind-registry:5001 running in docker, but this would likely not be something we could count on for a real kube cluster.  In other words, we'd require that in-cluster registry to be running https, and have valid non-self-signed certificates.

revit13 (approved):

PR #825, opened 2021-09-28
use ubi as the base image for manager

ngoracke:
Signed-off-by: Nicholas Goracke <ngoracke@us.ibm.com>

### DIFF
--- a/manager/Dockerfile
+++ b/manager/Dockerfile
@@ -1,9 +1,10 @@
 # Copyright 2020 IBM Corp.
 # SPDX-License-Identifier: Apache-2.0
 
-FROM gcr.io/distroless/static:nonroot
-WORKDIR /
-COPY manager .
-USER nonroot:nonroot
+FROM ubi8/ubi-minimal
+ENV HOME=/tmp
+WORKDIR /tmp
+COPY manager /
+USER 1001
 
 ENTRYPOINT ["/manager"]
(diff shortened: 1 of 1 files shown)


manager/Dockerfile
```diff
@@ -1,9 +1,10 @@
 # Copyright 2020 IBM Corp.
 # SPDX-License-Identifier: Apache-2.0
 
-FROM gcr.io/distroless/static:nonroot
-WORKDIR /
-COPY manager .
-USER nonroot:nonroot
+FROM ubi8/ubi-minimal
```
roee88, line 4: Should this be taken from registry.access.redhat.com?

manager/Dockerfile
```diff
@@ -1,9 +1,10 @@
...
+ENV HOME=/tmp
+WORKDIR /tmp
```
roee88, line 6: Just wondering, is it a best practice to use /tmp as the workdir?

manager/Dockerfile
ngoracke: I think that would be more explicit.  I'll change it

manager/Dockerfile
ngoracke: This is something we've taken to doing for openshift.  Often, you'll be running with an SCC that randomizes the user ID you run as, so you want to be sure you've got access to execute the binary, and access to write any files to $HOME or PWD if the application requires it.

roee88 (approved):
We should replicate

PR #827, opened 2021-09-29
Remove PluginMechanism from fybrikmodule_types.go

revit13:
Co-authored-by: Shlomit Koyfman <shlomitk@il.ibm.com>

Its actually a `string` value and can be removed for consistency.


shlomitk1 (approved):

simanadler:
If the intent is that the values will be in the taxonomy then I agree.  Has that been finalized though with @ronenkat and @eletonia?

PR #833, opened 2021-10-05
Added a cleanup step to the notebook sample

Raviv-S:
Added step 3 on cleanup which deletes configmap on fybrik-system namespace.


roee88 (approved):
Thank you

PR #836, opened 2021-10-06
Refine Vault credentials passed to copy module.

revit13:
This PR refines Vault credentials passed to the copy module: it specifies only the write permissions for the copy destination and only the read permissions for the copy source.


revit13:
@shlomitk1 right, this pr addresses the only copy flow.

PR #853, opened 2021-10-14
Quickstart update new opa policy

ritwikc:
This PR completes the following tasks:

- Modifies the OPA policy in quickstart to conform with the taxonomy
- Changed `asset.yaml` so that each column is tagged as `PII`
- Changed `opa_reader.go` file to handle a stand alone data tag such as `finance`

Need to test if the quickstart is working fine or not. Once, it is done this PR will be moved out of draft.

### DIFF
--- a/connectors/open_policy_agent/lib/opa_reader.go
+++ b/connectors/open_policy_agent/lib/opa_reader.go
@@ -46,9 +46,13 @@ func (r *OpaReader) updatePolicyManagerRequestWithResourceInfo(in *openapiclient
 
 					tagVal := make(map[string]interface{})
 					for i := 0; i < len(tagArr); i++ {
-						splitStr := strings.Split(tagArr[i], " = ")
-						// residency = Turkey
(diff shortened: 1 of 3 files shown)


connectors/open_policy_agent/lib/opa_reader.go
```diff
@@ -46,9 +46,13 @@ func (r *OpaReader) updatePolicyManagerRequestWithResourceInfo(in *openapiclient
 
 					tagVal := make(map[string]interface{})
 					for i := 0; i < len(tagArr); i++ {
-						splitStr := strings.Split(tagArr[i], " = ")
-						// residency = Turkey
-						tagVal[splitStr[0]] = splitStr[1]
+						if strings.Contains(tagArr[i], " = ") {
+							splitStr := strings.Split(tagArr[i], " = ")
+							// residency = Turkey
```
ronenkat, line 51: please make this comment more generic, e.g., metadata-key = metadata-value

PR #855, opened 2021-10-14
support retrieving helm charts from private helm registry using a dockerconfigjson secret

cdoron:

There are problems using `k8s.io/kubernetes` as a dependency.
One solution is to use https://github.com/vdemeester/k8s-pkg-credentialprovider , which is a "k8s.io/kubernetes/pkg/credentialprovider temporary fork"
Instead, I used a workaround suggested here:
https://stackoverflow.com/questions/59187781/revision-v0-0-0-unknown-for-go-get-k8s-io-kubernetes


roee88:
This implementation seems costly in terms of the work done on each reconcile. How about adding an explicit field for the secret next to the chart field in FybrikModule? That way you would know that login is required and what secret to use (the keyring code is still required in case there are multiple credentials in the secret). 

As for dependencies, using https://github.com/vdemeester/k8s-pkg-credentialprovider is fine by me if we can't easily extract just what we need.

PR #856, opened 2021-10-14
Update dashboard-sample.md

elsalant:
Editing changes, including a more accurate description of the EU use case, some corrections to the explanations given in the quick start etc.


site/docs/samples/dashboard-sample.md
```diff
@@ -76,47 +72,46 @@ helm install rel1-backend-server ./tmp/backend_server
...
 kubectl apply -n fybrik-blueprints -f https://raw.githubusercontent.com/fybrik/fogProtect-dashboard-sample/main/secrets/jwt_key_secret.yaml
 ```
 
-## Fybrik manager RBAC
+## Fybrik Manager RBAC
 
-In order for the fybrik manager to be able to access the assets and to deploy the fybrik module that we created, give 
-the manager the relevant [RBAC authorization](https://github.com/fybrik/fogProtect-dashboard-sample/blob/main/fybrik-system-manager-rbac.yaml):  
+In order for the Fybrik Control Plane to be able to access the JWT key that we created, give 
+the Manager the relevant [RBAC authorization](https://github.com/fybrik/fogProtect-dashboard-sample/blob/main/fybrik-system-manager-rbac.yaml):
```
ronenkat, line 103: why do we need this RBAC and access to jwt key?
More generally about the sample, it has cluster roles --- which it shouldn't. Please reduce the RBAC that the sample needs.

site/docs/samples/dashboard-sample.md
```diff
@@ -76,47 +72,46 @@ helm install rel1-backend-server ./tmp/backend_server
...
-from the user. Once a request is received a decision must be made regarding the request, it should be either allowed, 
-columns redacted or blocked depending on the role of the user. The decision is made using 
-[OpenPolicyAgent](https://www.openpolicyagent.org), and applying the policy described in 
-[About the dashboard application](#About the dashboard application) and specified 
+from the user. Once a request is received a decision must be made regarding the request based on the role of the user and the characterization of the data being requested; it should be either allowed, potentially with fields redacted, or blocked. The decision is made using 
+[OpenPolicyAgent](https://www.openpolicyagent.org), and applying the policy specified 
 [here](https://github.com/fybrik/fogProtect-dashboard-sample/blob/main/python/fogprotect-policy.yaml).  
-Deploy the fybrik module and application:
+    
+Deploy the Fybrik module and application:
```
ronenkat, line 114: please separate between control plane definition, i.e., registering of the module, and running the application.
You need to add a title that says something like **Running the sample**.

site/docs/samples/dashboard-sample.md
```diff
@@ -125,9 +120,10 @@ kubectl wait --for=condition=ready --all pod --timeout=120s
 
 ## Deploy the dashboard  
 
-We now deploy the dashboard that will display a table that contains the safety data of the factory, along with two 
-buttons to start and stop the manufacturing robot. One can change the role of the user using a pull down menu.  
-First create a port-forwarding to the service that will be intercepting the HTTP requests:  
+We now deploy the dashboard that will display a table containing the safety data of the factory, along with two
```
ronenkat, line 123: move the command to deploy first, and only afterwards explain the dashboard

site/docs/samples/dashboard-sample.md
```diff
@@ -76,47 +72,46 @@ helm install rel1-backend-server ./tmp/backend_server
...
-
 ## Deploy the module
 
 In our case, the module as we described earlier is responsible for intercepting the HTTP requests received 
-from the user. Once a request is received a decision must be made regarding the request, it should be either allowed, 
-columns redacted or blocked depending on the role of the user. The decision is made using 
-[OpenPolicyAgent](https://www.openpolicyagent.org), and applying the policy described in 
-[About the dashboard application](#About the dashboard application) and specified 
+from the user. Once a request is received a decision must be made regarding the request based on the role of the user and the characterization of the data being requested; it should be either allowed, potentially with fields redacted, or blocked. The decision is made using 
+[OpenPolicyAgent](https://www.openpolicyagent.org), and applying the policy specified
```
ronenkat, line 111: Please add in-line an example policy that will relate to the tag in the in-line asset you have added earlier.

site/docs/samples/dashboard-sample.md
elsalant: I don't know - Robert added this code.  I need to investigate

site/docs/samples/dashboard-sample.md
ronenkat: True, but having the policy inline will help understand how the policy affects the data.
A link by itself does not explain the policy.
The text here should be helpful in explaining how to write policies as well.

site/docs/samples/dashboard-sample.md
elsalant: I will fix this code.  A secret key is required for JWT creation and verification.  The GUI will require the key for JWT creation and the module for verification.  The current code creates two copies of this key, one in the fogprotect ns (for the GUI) and the other in the fybrik-blueprint ns for the module.  The code uses the serviceaccount for the Manager, and then due to ns issues, creates a global binding.  This of course is not the way to go.

Instead, the code should be fixed as follows:
1. The module creates a serviceaccount to handle the JWT secret.  This will be created in the fybrik-system ns
2. A single secret is created for the JWT
3. Define a ClusterRole for a jwt-secret-reader-role which only gives verbs for reading (get, list...)
4. Define a RoleBinding associated with the JWT secret in both the fybrik-blueprint ns and in the fogprotect ns to give the GUI and module access to the secret.  Note that the RoleBinding restricts the ClusterRole definitions to the specified namespace.

roytman:
@elsalant could you please rebase the PR

site/docs/samples/dashboard-sample.md
```diff
@@ -65,7 +61,7 @@ This enables easy [cleanup](#cleanup) once you're done experimenting with the sa
 ## Backend data server
 
 The backend data server is responsible for reading or writing data to a database or any other resource. In this 
-example the backend data server is a simple server that returns the mock data for the dashboard.    
+example the backend data server is a simple server that returns the mock container data.
```
roytman, line 64: the new helm (3.7) doesn't support `helm chart` commands, the code snippet below should be updated

site/docs/samples/dashboard-sample.md
```diff
@@ -76,63 +72,59 @@ helm install rel1-backend-server ./tmp/backend_server
...
 kubectl -n fybrik-blueprints port-forward svc/rest-read 5559:5559 &
 ```
+## Deploy the sample application  
 
-Afterwards, deploy the dashboard application:  
+We now deploy the dashboard that will display a table containing the safety data of the factory, along with two 
+buttons to emulate starting and stopping the manufacturing robot. One can change the role of the user using a pull down menu.   
 ```bash
 helm chart pull ghcr.io/fybrik/factory-gui-chart:v0.0.1
 helm chart export --destination=./tmp ghcr.io/fybrik/factory-gui-chart:v0.0.1
```
roytman, line 130: the same with `helm chart export`

PR #858, opened 2021-10-17
Default config policies evaluator

shlomitk1:
Refers to https://github.com/fybrik/fybrik/issues/794

This PR implements the following tasks:
- definition of the interfaces between the manager and the config policy evaluator
- definition of the config policy evaluator interface
- implementing the default config policy evaluator interface (no OPA, no written policies, replacement of the hard-coded manager logic by hard-coded evaluator decisions)
- separating the code of the plotter generation into 1. module selection and 2. plotter spec generation


manager/controllers/app/config/default_evaluator.go
```diff
@@ -0,0 +1,66 @@
...
+func NewDefaultConfig() *DefaultConfig {
+	return &DefaultConfig{}
+}
+
+// Evaluate replaces hard-coded decisions in manager by default configuration
+// The following logic is implemented:
+/* 	Read capability is deployed in a read-type scenario.
+    Read capability is deployed at the workload scope.
+	Write capability is not yet supported and thus won’t be deployed.
+	Transform is deployed if a list of governance actions is not empty.  Transforms should always be performed close to the data.
```
revit13, line 27: is this condition reflected in the code? transform is always allow.

manager/controllers/app/config/evaluator_input.go
```diff
@@ -0,0 +1,81 @@
...
+	dataset "fybrik.io/fybrik/manager/controllers/app/dataset"
+	"fybrik.io/fybrik/pkg/multicluster"
+	model "fybrik.io/fybrik/pkg/taxonomy/model/base"
+)
+
+// WorkloadInfo holds workload details such as cluster/region, type, etc.
+type WorkloadInfo struct {
+	// Cluster where the user workload is running
+	Cluster multicluster.Cluster
+}
```
simanadler, line 18: @shlomitk1 I assume in the future there will be additional information and that this will be a taxonomy driven structure declared somewhere else.  Is this a correct assumption?

manager/controllers/app/config/evaluator_input.go
```diff
@@ -0,0 +1,81 @@
...
+	Interface api.InterfaceDetails
+	// requested usage, e.g. "read": true, "write": false
+	Usage map[api.DataFlow]bool
+}
+
+// EvaluatorInput is an input to Configuration Policies Evaluator.
+// Used to evaluate configuration policies.
+type EvaluatorInput struct {
+	// A list of available clusters
+	Clusters []multicluster.Cluster
```
simanadler, line 34: @shlomitk1 In the future the list of available clusters should be provided as part of the infrastructure metadata, rather than being passed in as part of the EvaluatorInput.  Right? Perhaps until we have something final this information should just be in some configmap?

manager/controllers/app/dataset/data_info.go
```diff
@@ -0,0 +1,52 @@
...
+	// Geography is the geo-location of the asset
+	Geography string
+	// Connection is the connection details in raw format as received from the connector
+	Connection serde.Arbitrary
+	// Metadata
+	Metadata *pb.DatasetMetadata
+}
+
+// Transforms a CatalogDatasetInfo into a DataDetails struct
+// TODO Think about getting rid of one or the other and reuse
```
simanadler, line 32: Agree.  We shouldn't be doing such transforms.  There should be a single structure used (and validated via the taxonomy) throughout.

manager/controllers/app/dataset/data_info.go
```diff
@@ -0,0 +1,52 @@
...
+
+// DataDetails is the asset metadata received from the catalog connector
+// This structure is in use by the manager and other components, such as policy manager and config policies evaluator
+type DataDetails struct {
+	// Name of the asset
+	Name string
+	// Interface is the protocol and format
+	Interface app.InterfaceDetails
+	// Geography is the geo-location of the asset
+	Geography string
```
simanadler, line 24: Isn't this part of the metadata?  Why is it called out separately?

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -13,8 +13,11 @@ import (
 
 	"fybrik.io/fybrik/manager/controllers"
 	"fybrik.io/fybrik/pkg/environment"
+	local "fybrik.io/fybrik/pkg/multicluster/local"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 
+	config "fybrik.io/fybrik/manager/controllers/app/config"
```
simanadler, line 19: Do we want it here or in pkg?

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -239,6 +243,20 @@ func (r *FybrikApplicationReconciler) reconcileFinalizers(applicationContext *ap
 	return nil
 }
 
+// GetLocalCluster returns the local cluster metadata
+func (r *FybrikApplicationReconciler) GetLocalCluster() (multicluster.Cluster, error) {
```
simanadler, line 247: Should this perhaps be in pkg/multicluster?

manager/controllers/app/config/default_evaluator.go
```diff
@@ -0,0 +1,66 @@
+// Copyright 2020 IBM Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+package config
```
simanadler, line 4: I would suggest calling it adminconfig (or something similar) so it's not confused with confused with fybrik/config

manager/controllers/app/config/evaluator_interface.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2020 IBM Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+package config
+
+import corev1 "k8s.io/api/core/v1"
+
+// EvaluatorInterface is an interface for config policies' evaluator
+type EvaluatorInterface interface {
+	Evaluate(in *EvaluatorInput) (EvaluatorOutput, error)
```
simanadler, line 10: Should there also be an interface for loading the relevant config policies?  For now its implementation would do nothing.

manager/controllers/app/config/evaluator_input.go
shlomitk1: Yes.

manager/controllers/app/dataset/data_info.go
shlomitk1: There was some technical issue with using grpc response structure as is. I don't remember all the details but 1. it should eventually go away as we get rid of the grpc structures and 2. it is not relevant for this PR - I just moved the structure to a new package because it was causing import loops.

manager/controllers/app/config/default_evaluator.go
shlomitk1: OK.

manager/controllers/app/config/evaluator_interface.go
shlomitk1: It should, I thought to add when it's ready and replaces the default implementation in manager.

manager/controllers/app/config/evaluator_input.go
simanadler: @shlomitk1 I think clusters will be part of the metadata describing the infrastructure available and its costs, capabilities, etc.  I don't think it should be in the parameters passed to the evaluator, but rather stored with the rest of the infra metadata.  Where/how we store that is currently TBD (OPA data, part of the taxonomy, configmap, etc).  I would suggest adding an interface for obtaining infra metadata.  Initital implementation of that could be to obtain the clusters using ClusterManager.

manager/controllers/app/dataset/data_info.go
simanadler: OK.  Please open an issue to address it though

manager/controllers/app/dataset/data_info.go
simanadler: Please open an issue for this

manager/controllers/app/metadata/data_info.go
```diff
@@ -0,0 +1,52 @@
+// Copyright 2020 IBM Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+package metadata
```
simanadler, line 4: Not a great name.  Metadata about what?  workload, infra, data?  Perhaps datametadata?  
Why isn't all of this part of the catalog connector?  I know in this PR you are just moving code around, so perhaps open an issue about this.

Issue #863, opened 2021-10-18
Resolving endpoint in FybrikApplication fails if no labels


laurieaw63:
If no labels are present resolving the template endpoint for FybrikApplication fails with this error:
```
2021-10-18T17:30:38.319Z	INFO	controllers.FybrikApplication	Reconciled with errors: An error was received for asset {"asset_id": "101ae141-d5de-4f9d-8707-14a7ff2eb6f1", "context_type": "catalogs", "context_id": "86297613-3de1-453a-aaa6-21a74b9310a3"} . If the error persists, please contact an operator.Error description: could not process template {{ .Release.Name}}.{{ get .Values.labels "app.fybrik.io/dataAccessModuleNamespace" | default .Release.Namespace }}: template: hostname:1:33: executing "hostname" at <.Values.labels>: wrong type for value; expected map[string]interface {}; got interface {}	{"fybrikapplication": "fybrik-lawcloud/lawapp-nolabel"}

```

PR #864, opened 2021-10-18
Fix issue with endpoint in fybrikApplication #863

laurieaw63:
Fix issue with endpoint when no labels exist in fybrikApplication #863

Signed-off-by: Laurie Williams <laurieaw@us.ibm.com>

### DIFF
--- a/manager/controllers/app/moduleinstance.go
+++ b/manager/controllers/app/moduleinstance.go
@@ -461,25 +461,32 @@ func moduleAPIToService(api *app.ModuleAPI, scope app.CapabilityScope, appContex
 	releaseName := utils.GetReleaseName(appContext.Name, appContext.Namespace, instanceName)
 	blueprintNamespace := utils.GetBlueprintNamespace()
 
-	values := make(map[string]interface{})
-	if len(appContext.Labels) == 0 {
-		values["Release"] =
-			map[string]interface{}{
-				"Name":      releaseName,
-				"Namespace": blueprintNamespace,
-			}
-	} else {
-		values["Release"] = map[string]interface{}{
-			"Name":      releaseName,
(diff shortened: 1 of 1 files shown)


manager/controllers/app/moduleinstance.go
```diff
@@ -461,14 +461,21 @@ func moduleAPIToService(api *app.ModuleAPI, scope app.CapabilityScope, appContex
...
 			"Name":      releaseName,
 			"Namespace": blueprintNamespace,
-		},
-		"Values": map[string]interface{}{
+		}
+		values["Values"] = map[string]interface{}{
 			"labels": appContext.Labels,
-		},
+		}
 	}
```
roee88, line 479: Looking forward we may add other attributes to Values so we may want to improve this to be a bit cleaner.
Consider either initializing as an empty map when the labels is nil, or using structs with `omitempty` in the json tag.

laurieaw63: updated code

shlomitk1 (approved):

manager/controllers/app/moduleinstance.go
```diff
@@ -461,18 +461,33 @@ func moduleAPIToService(api *app.ModuleAPI, scope app.CapabilityScope, appContex
...
-			"labels": appContext.Labels,
+		Values: Values{
+			Labels: appContext.Labels,
 		},
 	}
 
 	// the following is required for proper types (e.g., labels must be map[string]interface{})
-	values, err = utils.StructToMap(values)
+	var values map[string]interface{}
+	values, err = utils.StructToMap(args)
```
roee88, line 490: Why the declaration and not just `values, err := utils.StructToMap(args)`

PR #872, opened 2021-10-19
Create blueprints in the system namespace

shlomitk1:

Refers to https://github.com/fybrik/fybrik/issues/793

This PR differentiates between a blueprint namespace and a modules' namespace. It allocates blueprints in the system namespace (as plotters) and deploys modules in the predefined namespace.
Removed duplicates in role definitions.
BLUEPRINT_NAMESPACE is replaced by MODULES_NAMESPACE env var.


manager/controllers/motion/suite_test.go
```diff
@@ -100,9 +99,9 @@ var _ = BeforeSuite(func(done Done) {
 	} else {
 		fmt.Printf("Setup fake environment... \n")
 
-		blueprintNamespace := utils.GetBlueprintNamespace()
-		fmt.Printf("Motion test suite using blueprint namespace: %s\n", blueprintNamespace)
-		workerNamespaceSelector := fields.SelectorFromSet(fields.Set{"metadata.namespace": blueprintNamespace})
+		dataAccessModuleNamespace := utils.GetDataAccessModuleNamespace()
+		fmt.Printf("Motion test suite using blueprint namespace: %s\n", dataAccessModuleNamespace)
```
laurieaw63, line 103: nit change comment blueprint===data access module namespace (to avoid confusion)

manager/controllers/app/fybrik_notebook_test.go
```diff
@@ -146,15 +146,15 @@ func TestS3Notebook(t *testing.T) {
...
 	}, timeout, interval).Should(Equal(true))
 
-	blueprintNamespace := utils.GetBlueprintNamespace()
-	fmt.Printf("blueprint namespace notebook test: %s\n", blueprintNamespace)
+	dataAccessModuleNamespace := utils.GetDataAccessModuleNamespace()
+	fmt.Printf("data access module namespace notebook test: %s\n", dataAccessModuleNamespace)
 
 	// Forward port of arrow flight service to local port
-	svcName := strings.Replace(application.Status.AssetStates["fybrik-notebook-sample/data-csv"].Endpoint.Hostname, "."+blueprintNamespace+".svc.cluster.local", "", 1)
+	svcName := strings.Replace(application.Status.AssetStates["fybrik-notebook-sample/data-csv"].Endpoint.Hostname, "."+dataAccessModuleNamespace+".svc.cluster.local", "", 1)
```
laurieaw63, line 153: Unrelated to this change, the .svc.cluster.local should be removed. This is the cluster domain name and is configurable. So may not be svc.cluster.local. Besides it is not needed when accessing the endpoint.

manager/controllers/app/blueprint_controller.go
```diff
@@ -233,6 +234,7 @@ func (r *BlueprintReconciler) reconcile(ctx context.Context, log logr.Logger, bl
 		}
 		module.Arguments.Labels[app.BlueprintNameLabel] = blueprint.Name
 		module.Arguments.Labels[app.BlueprintNamespaceLabel] = blueprint.Namespace
+		module.Arguments.Labels[app.DataAccessModuleNamespaceLabel] = dataAccessModuleNamespace
```
laurieaw63, line 237: I believe the module.Arguments.Labels will end up being passed to the data access module in the .Values.labels. If that is true then we cannot add this label if dataAccessModuleNamespace == default blueprints namespace, until we fully implement pass 2. For pass 1, the chart is creating the namespace. And if this is the default blueprints namespace then the helm install will fail if this label is set.

manager/controllers/app/blueprint_controller.go
```diff
@@ -132,7 +132,7 @@ func (r *BlueprintReconciler) deleteExternalResources(blueprint *app.Blueprint)
 
 func (r *BlueprintReconciler) applyChartResource(log logr.Logger, chartSpec app.ChartSpec, args map[string]interface{}, blueprint *app.Blueprint, releaseName string) (ctrl.Result, error) {
 	log.Info(fmt.Sprintf("--- Chart Ref ---\n\n%v\n\n", chartSpec.Name))
-	kubeNamespace := blueprint.Namespace
+	kubeNamespace := utils.GetDataAccessModuleNamespace()
```
laurieaw63, line 135: Just a suggestion to avoid confusion, I think this method should be named GetDefaultDataAccessModuleNamespace(). The data access module namespace will either be the default configure when Fybrik is deployed OR what is specified with the label. For pass 1, when the label is specified these will be different. The helm release will be created in the default data access module but the artifacts for the data access module will be created in the namespace specified in the label.  When we get to pass 2, this will no longer be the case.

charts/fybrik/templates/fybrik-blueprints-role.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
```
laurieaw63, line 3: Not related to this review, but we will need to address and change using verb '*' and resources '*' at some point. This is be a security issue to use *.

laurieaw63 (commented):
This change will also require some changes to custom pipeline. We will need to coordinate this change going in.

Makefile
```diff
@@ -20,7 +20,7 @@ deploy: $(TOOLBIN)/kubectl $(TOOLBIN)/helm
                --namespace $(KUBE_NAMESPACE) --wait --timeout 120s
 
 .PHONY: test
-test: export BLUEPRINT_NAMESPACE?=fybrik-blueprints
+test: export DATA_ACCESS_MODULE_NAMESPACE?=fybrik-blueprints
```
roee88, line 23: I think that we should drop "data access" and just say modules. So `MODULES_NAMESPACE`, `.Values.modulesNamespace`, etc. would be my recommendation.

manager/apis/app/v1alpha1/blueprint_types.go
```diff
@@ -211,6 +211,7 @@ func CreateMetaBlueprintWithoutState(blueprint *Blueprint) MetaBlueprint {
 }
 
 const (
-	BlueprintNamespaceLabel = "app.fybrik.io/blueprintNamespace"
-	BlueprintNameLabel      = "app.fybrik.io/blueprintName"
+	BlueprintNamespaceLabel        = "app.fybrik.io/blueprintNamespace"
+	BlueprintNameLabel             = "app.fybrik.io/blueprintName"
+	DataAccessModuleNamespaceLabel = "app.fybrik.io/dataAccessModuleNamespace"
```
roee88, line 216: AFAIK usually dashed-case is used so I suggest to use `app.fybrik.io/modules-namespace` here and in a future PR do the same for all of our labels.

manager/apis/app/v1alpha1/blueprint_types.go
```diff
@@ -211,6 +211,7 @@ func CreateMetaBlueprintWithoutState(blueprint *Blueprint) MetaBlueprint {
 }
 
 const (
-	BlueprintNamespaceLabel = "app.fybrik.io/blueprintNamespace"
-	BlueprintNameLabel      = "app.fybrik.io/blueprintName"
+	BlueprintNamespaceLabel        = "app.fybrik.io/blueprintNamespace"
+	BlueprintNameLabel             = "app.fybrik.io/blueprintName"
```
roee88, line 215: Why do we need these if it's always the system namespace?

manager/controllers/app/blueprint_controller.go
roee88: I agree that this line should be removed, we don't need it for neither pass 1 nor pass 2.

charts/fybrik/templates/fybrik-admin-role.yaml
```diff
@@ -32,90 +32,5 @@ rules:
   - patch
```
roee88, line 32: admin-role was supposed to be a role applied to a human administrator (to be able to create modules, storage accounts, do debugging, etc.). It currently has nothing to do with it so why not just move whatever is left to the system-role and open an issue for creating a new fybrik-admin-role for humans?

manager/apis/app/v1alpha1/blueprint_types.go
roee88: I take it back. The information might be useful for debugging. The `blueprintNamespace` name is misleading after this change tough.

manager/controllers/app/blueprint_controller.go
shlomitk1: Renamed to GetDefaultModulesNamespace()

manager/controllers/app/blueprint_controller.go
shlomitk1: Removed.

manager/controllers/app/fybrik_notebook_test.go
shlomitk1: Removed.

manager/controllers/utils/environmentVariable.go
```diff
@@ -7,18 +7,18 @@ import (
 	"os"
 )
 
-// BlueprintNamespace defines a namespace where blueprints and associated resources will be allocated
-const DefaultBlueprintNamespace = "fybrik-blueprints"
+// DefaultDataAccessModuleNamespace defines a namespace where module resources will be allocated
```
laurieaw63, line 10: Nit use DefaultModuleNamespace

manager/controllers/app/blueprint_controller.go
laurieaw63: Actually we need this for pass 1. If the , if the dataAccessModuleNamespace label is specified and is != defaultModuleNamespace then set it. if the dataAccessModuleNamespace label is specified == defaultModuleNamespace ; don't set it. We need this label to be conditionally set in the values.labels so data access module can create it if NOT the default.

manager/apis/app/v1alpha1/blueprint_types.go
laurieaw63: If we change this label's name we also need to coordinate with the data control plane developers as they are now using this label; and any developers that have fybrikapplication yamls using this value. I would vote to leave it.

laurieaw63 (changes requested):

manager/apis/app/v1alpha1/blueprint_types.go
roee88: I don't see why it makes any impact on existing users. This is an unreleased feature, if you set app.fybrik.io/dataAccessModuleNamespace and not the new label then the existing behavior is preserved: module will be installed to default namespace and the label will still be passed as-is to the module.

shlomitk1:
@laurieaw63 I'm taking out the dataAccessModuleNamespace from this PR. There is another related task for this: "Propagate FybrikApplication label for managed configurations" - the label will be added in that PR. As for the label name and conditions on when to use it, let's coordinate inside the issue.

laurieaw63:
One final comment, since we are trying to distinguish between blueprints namespace and modules namespace I think we should change fybrik-blueprints to fybrik-modules?

shlomitk1:
> 
> 
> One final comment, since we are trying to distinguish between blueprints namespace and modules namespace I think we should change fybrik-blueprints to fybrik-modules?

Makes sense to me. @roee88 - what do you think?

roee88:
> > One final comment, since we are trying to distinguish between blueprints namespace and modules namespace I think we should change fybrik-blueprints to fybrik-modules?
> 
> Makes sense to me. @roee88 - what do you think?



> > One final comment, since we are trying to distinguish between blueprints namespace and modules namespace I think we should change fybrik-blueprints to fybrik-modules?
> 
> Makes sense to me. @roee88 - what do you think?

I tend to keep it as is, it's not about what CRDs are deployed to the namespace but rather what is logically there. A group of modules linked together is logically an instance of a blueprint not a module/modules.

PR #893, opened 2021-10-21
support retrieving helm charts from private helm registry

cdoron:

In `manager/main.go`, we temporarily made the following change

```
&corev1.Secret{}:                {Field: systemNamespaceSelector},
//&corev1.Secret{}:                {Field: workerNamespaceSelector},
```
We will need to figure out which secrets we need to place in the cache.

Following Roee's suggestion, we add an explicit field for the secret next to the chart field in FybrikModule.  That way you would know that login is required and what secret to use.

We use https://github.com/vdemeester/k8s-pkg-credentialprovider
to overcome the problem with using `k8s.io/kubernetes` as a dependency

### DIFF
--- a/manager/main.go
+++ b/manager/main.go
@@ -80,10 +80,11 @@ func run(namespace string, metricsAddr string, enableLeaderElection bool,
 		&motionv1.StreamTransfer{}:      {Field: workerNamespaceSelector},
 		&kbatch.Job{}:                   {Field: workerNamespaceSelector},
 		&kbatch.CronJob{}:               {Field: workerNamespaceSelector},
-		&corev1.Secret{}:                {Field: workerNamespaceSelector},
 		&corev1.Pod{}:                   {Field: workerNamespaceSelector},
 		&kapps.Deployment{}:             {Field: workerNamespaceSelector},
 		&corev1.PersistentVolumeClaim{}: {Field: workerNamespaceSelector},
+		&corev1.Secret{}:                {Field: systemNamespaceSelector},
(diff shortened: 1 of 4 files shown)


shlomitk1 (changes requested):
First, you need to rebase.
Second, `&corev1.Secret{}:                {Field: systemNamespaceSelector},` should be used because the secret is still taken from the system namespace. 
Do we have a role + rolebinding for get/list/watch secrets in the system namespace?

manager/main.go
```diff
@@ -80,10 +80,10 @@ func run(namespace string, metricsAddr string, enableLeaderElection bool,
 		&motionv1.StreamTransfer{}:      {Field: workerNamespaceSelector},
 		&kbatch.Job{}:                   {Field: workerNamespaceSelector},
 		&kbatch.CronJob{}:               {Field: workerNamespaceSelector},
-		&corev1.Secret{}:                {Field: workerNamespaceSelector},
 		&corev1.Pod{}:                   {Field: workerNamespaceSelector},
 		&kapps.Deployment{}:             {Field: workerNamespaceSelector},
 		&corev1.PersistentVolumeClaim{}: {Field: workerNamespaceSelector},
```
shlomitk1, line 85: These are no longer correct. Please rebase.

manager/main.go
```diff
@@ -80,10 +80,10 @@ func run(namespace string, metricsAddr string, enableLeaderElection bool,
...
+		&corev1.Secret{}:                {Field: workerNamespaceSelector},
```
shlomitk1, line 86: We no longer require watching  for secrets in the module namespace.

PR #925, opened 2021-10-31
Add job to the build workflow for the notebook sample.

revit13:
This PR adds a job to the build workflow for the notebook sample.
It also contains additional two fixes:
- added `SHELL := /bin/bash` to the manager/Makefile to avoid failing on `pushd` command
- calls `NewGomegaWithT` in the sample test to make gomega functions work properly.

### DIFF
--- a/manager/Makefile
+++ b/manager/Makefile
@@ -1,3 +1,4 @@
+SHELL := /bin/bash
 ROOT_DIR := ..
 include $(ROOT_DIR)/Makefile.env
 #include Makefile.env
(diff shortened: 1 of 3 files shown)


froesef (approved):

manager/Makefile
```diff
@@ -1,3 +1,4 @@
+SHELL := /bin/bash
```
roee88, line 1: Please revert. 

There is no need to use pushd. Just use cd  (no need to cd back too, it doesn't affect the next command).

roee88 (changes requested):

manager/Makefile
revit13: Done. Thanks

roee88 (approved):

PR #933, opened 2021-11-02
OpenAPI based API for catalog connectors

### LINKED ISSUE #939: Define OpenAPI based API for catalog connectors
This issue is regarding defining schemas, base and layers related files (as introduced by https://github.com/fybrik/fybrik/issues/693) for catalog connectors to fybrik.

rohithdv:
This PR defines json schema spec files for catalog connector API (with request and response objects). 
This PR fixes #939.


charts/fybrik/files/taxonomy/data_catalog_request.json
```diff
@@ -0,0 +1,19 @@
+{
+	"$schema": "http://json-schema.org/draft-04/schema#",
+	"title": "Data Catalog Request",
+	"type": "object",
+	"properties": {
+	  "datasetid": {
```
shlomitk1, line 6: Is there a naming convention? In go struct it was datasetID

charts/fybrik/files/taxonomy/data_catalog_response.json
```diff
@@ -0,0 +1,71 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Data Catalog Response",
+  "type": "object",
+  "properties": {
+    "name": {
+      "type": "string"
+    },
+    "Asset": {
```
shlomitk1, line 9: why start with a capital letter?

charts/fybrik/files/taxonomy/data_catalog_response.json
```diff
@@ -0,0 +1,71 @@
...
+        "connection": {
+          "$ref": "taxonomy.json#/definitions/Connection"
+        },
+        "dataFormat": {
+          "type": "string"
+        }
+      },
+      "required": [
+        "connection"
+      ]
```
shlomitk1, line 54: dataFormat should be required as well

charts/fybrik/files/taxonomy/data_catalog_response.json
```diff
@@ -0,0 +1,71 @@
...
+    },
+    "AssetMetadata": {
+      "type": "object",
+      "properties": {
+        "owner": {
+          "type": "string"
+        },
+        "geography": {
+          "type": "string"
```
shlomitk1, line 63: geography should be required

charts/fybrik/files/taxonomy/taxonomy.json
```diff
@@ -15,9 +15,33 @@
     "AppInfo": {
       "type": "object"
     },
+    "Authentication": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "type": "string"
+        }
```
shlomitk1, line 23: what about namespace? We need both to access to a secret

charts/fybrik/files/taxonomy/taxonomy.json
```diff
@@ -15,9 +15,33 @@
...
+    "Connection": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "type": "string"
+        }
+      },
+      "additionalProperties": true,
+      "required": [
+        "name"
```
shlomitk1, line 42: Wasn't it connection type that was required (s3/db2/...)?

config/taxonomy/codegen.spec.yaml
```diff
@@ -34,3 +34,31 @@ paths:
...
+        responses:
+          '200':
+            description: successful operation
+            content:
+              application/json:
+                schema:
+                  type: object
+                  $ref: "../../charts/fybrik/files/taxonomy/data_catalog_response.json"
+          '400':
+            description: Invalid status value
```
shlomitk1, line 64: I believe the description should indicate that an error happened, 400 is a valid status value.

charts/fybrik/files/taxonomy/data_catalog_request.json
```diff
@@ -0,0 +1,19 @@
+{
+	"$schema": "http://json-schema.org/draft-04/schema#",
+	"title": "Data Catalog Request",
+	"type": "object",
+	"properties": {
```
ronenkat, line 5: should we add an operation type for the request? i.e., read, create, delete assets? 
no need to implement for now, but how would it be reflected in the request (just the type for now).

charts/fybrik/files/taxonomy/data_catalog_response.json
ronenkat: Not sure that is correct for all cases, the control plane should be able to work even if it is not mentioned.

charts/fybrik/files/taxonomy/taxonomy.json
ronenkat: don't think authentication have to be a secret. That is one option.

charts/fybrik/files/taxonomy/data_catalog_request.json
shlomitk1: Right, we'll need the correct credentials based on the operation. Today, we get only "read" credentials.

charts/fybrik/files/taxonomy/data_catalog_response.json
shlomitk1: The control plane makes many decisions based on the geography of the dataset: from governance actions to config policies and deployment optimization. If this field is not required, people who write governance policies and admin policies need to be aware of this and write the policies appropriately, otherwise predicates will fail on an empty geography value.

roee88: > If this field is not required, people who write governance policies and admin policies need to be aware of this and write the policies appropriately, otherwise predicates will fail on an empty geography value.

This is still reasonable. I can only guess that a geography field is hardly supported in existing catalogs. It's also not required by all use cases. In my opinion we should be able to support this case as if geography is specified as "unknown" and the different personas need to be aware of it. However for this to work I would actually accept your suggestion to make this required so there will be control through taxonomy of whether "unknown" is a valid value.

charts/fybrik/files/taxonomy/taxonomy.json
roee88: @shlomitk1 In taxonomy `name` is the common name for a discriminator field (like in a Connection that name would be `s3`, `db2`, etc.). 

@rohithdv I don't get this one. The example taxonomy you added does not match this definition at all.

config/taxonomy/codegen.spec.yaml
```diff
@@ -20,7 +20,7 @@ paths:
         description: Policy Manager Request Object.
```
roee88, line 20: Instead of having a `codegen.spec.yaml` start creating OpenAPI specs for the different connector types.

roee88 (changes requested):

charts/fybrik/files/taxonomy/data_catalog_request.json
rohithdv: changed to assetID. Thanks.

charts/fybrik/files/taxonomy/data_catalog_response.json
rohithdv: fixed. Thanks.

config/taxonomy/codegen.spec.yaml
rohithdv: fixed. Now we have polcymanager.spec.yaml and catalogconnector.spec.yaml to separate out the APIs. I have fixed the description in the  catalogconnector.spec.yaml.

charts/fybrik/files/taxonomy/data_catalog_request.json
```diff
@@ -0,0 +1,19 @@
+{
+	"$schema": "http://json-schema.org/draft-04/schema#",
+	"title": "Data Catalog Request",
```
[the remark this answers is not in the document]
rohithdv: changed titles to "Data Catalog Service - Asset Details". Please review. Thanks.

charts/fybrik/files/taxonomy/data_catalog_request.json
```diff
@@ -0,0 +1,19 @@
...
+	},
+	"additionalProperties": false,
+	"required": [
+	  "datasetid"
+	],
+	"definitions": {
+        "datasetid": {
+            "type": "string"
+        }
+	}
```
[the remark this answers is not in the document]
rohithdv: fixed. Thanks.

config/taxonomy/codegen.spec.yaml
rohithdv: created policymanager.spec.yaml and catalogconnector.spec.yaml now. Please review. Thanks.

charts/fybrik/files/taxonomy/taxonomy.json
rohithdv: @roee88 Updated the taxonomy.json to have the output of example taxonomy. Please review. Thanks.

charts/fybrik/files/taxonomy/data_catalog_response.json
```diff
@@ -0,0 +1,71 @@
...
+    "name",
+    "Asset"
+  ],
+  "definitions": {
+    "AssetSpec": {
+      "type": "object",
+      "properties": {
+        "secretRef": {
+          "$ref": "#/definitions/SecretRef"
+        },
```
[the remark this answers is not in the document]
rohithdv: Renamed to secret for now. Please suggest any alternatives if needed. Basically this field provides the credentials to access the remote data asset. This is different from the credentials (obtained through a kubernetes secret) to access the data catalog itself (which is needed in the input request as shown [here](https://github.com/rohithdv/fybrik/blob/415d2f6f3c8d06bf2239d6763e4a40151799ddea/config/taxonomy/catalogconnector.spec.yaml#L15)). Thanks.

charts/fybrik/files/taxonomy/data_catalog_response.json
```diff
@@ -0,0 +1,71 @@
...
+        "assetMetadata": {
+          "$ref": "#/definitions/AssetMetadata"
```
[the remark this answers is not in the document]
rohithdv: We had [this](https://github.com/rohithdv/fybrik/blob/415d2f6f3c8d06bf2239d6763e4a40151799ddea/charts/fybrik/files/taxonomy/policy_manager_request.json#L44) for the policy manager. Are you referring to using the `resource` field ? Shouldn't the catalog info be more generic than the resource field ? Please suggest. Thanks.

charts/fybrik/files/taxonomy/data_catalog_response.json
```diff
@@ -0,0 +1,71 @@
...
+    "AssetDetails": {
+      "type": "object",
+      "description": "Asset details",
+      "properties": {
+        "connection": {
+          "$ref": "taxonomy.json#/definitions/Connection"
+        },
+        "dataFormat": {
+          "type": "string"
+        }
```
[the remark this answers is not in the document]
rohithdv: Updated this. Please check. Thanks.

charts/fybrik/files/taxonomy/data_catalog_response.json
```diff
@@ -0,0 +1,71 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Data Catalog Response",
+  "type": "object",
+  "properties": {
+    "name": {
+      "type": "string"
+    },
```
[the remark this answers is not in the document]
rohithdv: added description. Please check. Thanks.

charts/fybrik/files/taxonomy/data_catalog_request.json
rohithdv: added `operationType` field for this. Please review.

charts/fybrik/files/taxonomy/taxonomy.json
rohithdv: sorry I didnt understand..are you referring to the `Authentication` field ? If yes, then in the layers `authentication.yaml` file, do you mean we should have `oneOf` construct like we have for the `Connection` field ?

rohithdv: sorry I didnt understand. Are you referring to `Authentication` field? If yes, then do you mean that we have to use the `oneOf` construct like we do for the `Connection` field in the layers file ? @roee88

charts/fybrik/files/taxonomy/data_catalog_response.json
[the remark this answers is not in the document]
roee88: That's exactly the part of this PR that needs a design. Avoid redundancy, if it's not resource as-is then it's probably that includes resource.

charts/fybrik/files/taxonomy/data_catalog_request.json
ronenkat: @rohithdv I think the we need concerete operation types. we can start with "read" or "get", but expect we would need "create, "update" and "delete" in the future.
Since at this time only the manager talks with the catalog connector, it should be in the base taxonomy.

charts/fybrik/files/taxonomy/data_catalog_response.json
```diff
@@ -0,0 +1,72 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Data Catalog Response",
```
ronenkat, line 3: response should also have operation type, and the content would be for the type.
Can you generalize? 
i also suggest moving the name, type and asset to common definition as in the future they will be used by operationtype "create"/"update" so we should architecture for future use.
Even as we start only with "get"/"read".

charts/fybrik/files/taxonomy/data_catalog_response.json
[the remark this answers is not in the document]
rohithdv: Updated data catalog metadata to reuse resource from policy manager request object. Please review. Thanks. @roee88

charts/fybrik/files/taxonomy/taxonomy.json
rohithdv: Updated assetCredentials to have vaultPluginPath as is currently used in the protos as discussed with Revital. Please review. Thanks. @roee88 @revit13

charts/fybrik/files/taxonomy/data_catalog_response.json
rohithdv: Updated the base taxonomy to have the operationType enums. Please review. Thanks. 
Also moved name, type and asset to common definition under the definitions so that it can be reused later. I hope this was your suggested change. Please review. Thanks. @ronenkat

charts/fybrik/files/taxonomy/data_catalog_request.json
```diff
@@ -0,0 +1,17 @@
+{
+	"$schema": "http://json-schema.org/draft-04/schema#",
+	"title": "Data Catalog Request - GetDatasetDetails",
+	"type": "object",
+	"properties": {
+	  "assetID": {
+		"type": "string"
+	  },
+	  "operationType": {
+		"$ref": "taxonomy.json#/definitions/OperationType"
```
roee88, line 10: No need to reference taxonomy.json here because it's not configurable at runtime.

charts/fybrik/files/taxonomy/data_catalog_request.json
```diff
@@ -0,0 +1,17 @@
...
+	  }
+	},
+	"additionalProperties": false,
+	"required": [
+	  "assetID"
```
roee88, line 15: Should operation type also be required?

charts/fybrik/files/taxonomy/data_catalog_request.json
```diff
@@ -0,0 +1,17 @@
...
+	]
+}
```
roee88, line 17: Please fix formatting on all files (missing empty line at end of file)

charts/fybrik/files/taxonomy/data_catalog_response.json
roee88: I disagree about the need to echo request data in the response. If it's in the request we don't need it in the response.

charts/fybrik/files/taxonomy/data_catalog_response.json
```diff
@@ -0,0 +1,82 @@
...
```
roee88, line 3: Title should reflect that this is for get asset details.

charts/fybrik/files/taxonomy/data_catalog_request.json
roee88: @ronenkat I don't understand your goal. This json schema is for the get dataset details call. My understanding from the other comment is that operation type is for the get asset scenario, but required for knowing which credential permissions to obtain.

If your aim is to be future proof for other API calls like creating and deleting an asset in the catalog then I agree that we should factor out common definitions to a different schema. My suggestion is to create `data_catalog.json` with common definitions related to catalog assets and refer to that file from different request/response schemas.

charts/fybrik/files/taxonomy/data_catalog_response.json
```diff
@@ -0,0 +1,82 @@
...
+    "name",
+    "asset"
+  ],
+  "additionalProperties": false,
+  "definitions": {
+    "assetName": {
+      "type": "string",
+      "description": "Name of the asset"
+    },
+    "asset": {
```
roee88, line 26: Unused definition?

charts/fybrik/files/taxonomy/data_catalog_response.json
```diff
@@ -0,0 +1,82 @@
...
+      "type": "object",
+      "properties": {
+        "owner": {
+          "type": "string"
+        },
+        "geography": {
+          "type": "string"
+        },
+        "metadata": {
+          "$ref":"policy_manager_request.json#/definitions/Resource"
```
roee88, line 77: This should be owned by the catalog not the policy manager. I suggest to add data_catalog.json with common definitions like that. Should probably also rename Resource accordingly.

charts/fybrik/files/taxonomy/data_catalog_request.json
rohithdv: fixed. Please check. Thanks.

charts/fybrik/files/taxonomy/data_catalog_request.json
rohithdv: done formatting with vscode. However it does not insert empty blank line at the end of file. Hope it is ok. Thanks.

charts/fybrik/files/taxonomy/data_catalog_request.json
rohithdv: fixed. Thanks.

charts/fybrik/files/taxonomy/data_catalog_response.json
rohithdv: fixed. Thanks.

charts/fybrik/files/taxonomy/data_catalog_response.json
rohithdv: fixed. Thanks.

charts/fybrik/files/taxonomy/data_catalog_response.json
rohithdv: I have added data_catalog.json to have Resource definition. However change the name "Resource" to something else would result in generated code for policy manager to change which may have other dependencies. Can we postpone the renaming to another PR at a later stage ? Please review. Thanks.

charts/fybrik/files/taxonomy/data_catalog_request.json
rohithdv: removed from base taxonomy and added the definition in data_catalog_request.json. Please review. Thanks.

charts/fybrik/files/taxonomy/data_catalog_response.json
rohithdv: removed the field from the response object. Please check. Thanks.

charts/fybrik/files/taxonomy/data_catalog_response.json
```diff
@@ -0,0 +1,76 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Data Catalog Response - GetAssetDetails",
+  "type": "object",
+  "properties": {
+    "name": {
+      "$ref": "#/definitions/assetName"
```
roee88, line 7: What's the difference between this field and asset.assetMetadata.metadata.name?

charts/fybrik/files/taxonomy/data_catalog_response.json
roee88: @rohithdv currently there are no required fields here. The above discussion is about `geography` but we should think if `metadata` should be required too.

charts/fybrik/files/taxonomy/data_catalog_response.json
```diff
@@ -0,0 +1,76 @@
...
+    "assetMetadata": {
+      "type": "object",
+      "properties": {
+        "owner": {
+          "type": "string"
+        },
+        "geography": {
+          "type": "string"
+        },
+        "metadata": {
```
roee88, line 70: I find it confusing to have a `metadata` field inside `assetMetadata`. It might be better to add `owner` and `geography` directly to `data_catalog.json#/definitions/Resource` instead of wrapping (they might be valuable in the policy manager context as well).

Overall it seems like `data_catalog.json#/definitions/Resource` can represent `ResourceMetadata` and be a top level property of the response object and its only sibling is the details property with connection information and format.

charts/fybrik/files/taxonomy/data_catalog_request.json
roee88: Consider setting up your IDE with a better formatter as you also have inconsistencies in space indentation.

charts/fybrik/files/taxonomy/data_catalog_response.json
rohithdv: Changed the structure of the data catalog response.json based on the above comments. Please review and let me know if there are any changes needed. Thanks.

charts/fybrik/files/taxonomy/data_catalog_response.json
rohithdv: removed the `name` field as it is covered by the `resourceMetadata.name` field. Thanks.

charts/fybrik/files/taxonomy/data_catalog_request.json
rohithdv: changed the formatting. Please review. Thanks.

roee88 (approved):

charts/fybrik/files/taxonomy/data_catalog_request.json
rohithdv: As discussed, moved commond definitions of `details`, `credentials` and `resource` to `data_catalog.json` and referred to them in `data_catalog_response.json`. @ronenkat @roee88

ronenkat (approved):

PR #944, opened 2021-11-09
Avoid using ClusterRole if clusterScope value is false in Vault helm …

revit13:
This PR avoids using ClusterRole if clusterScope value is false in Vault helm.


charts/vault/templates/vault-cluster-role.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if eq .Values.plugins.vaultPluginSecretsKubernetesReader.clusterScope true }}
```
roee88, line 1: No need to eq true a boolean variable

roee88 (approved):

charts/vault/templates/vault-cluster-role.yaml
revit13: Fixed. Thanks

PR #955, opened 2021-11-14
support retrieving helm charts from private helm registry

cdoron:

Following Roee's suggestion, we add an explicit field for the secret next to the chart field in FybrikModule. That way you would know that login is required and what secret to use.

We use https://github.com/vdemeester/k8s-pkg-credentialprovider
to overcome the problem with using `k8s.io/kubernetes` as a dependency

### DIFF
--- a/manager/controllers/app/blueprint_controller.go
+++ b/manager/controllers/app/blueprint_controller.go
@@ -184,6 +184,7 @@ func (r *BlueprintReconciler) applyChartResource(ctx context.Context, log logr.L
 				err := r.Helmer.RegistryLogin(repoToPull, cred.Username, cred.Password, false)
 				if err == nil {
 					registrySuccessfulLogin = repoToPull
+					break
 				} else {
 					log.Info("Failed to login to helm registry: " + repoToPull)
 				}
@@ -196,7 +197,10 @@ func (r *BlueprintReconciler) applyChartResource(ctx context.Context, log logr.L
 	err := r.Helmer.ChartPull(chartSpec.Name)
 	// if we logged into a registry, let us try to logout
 	if registrySuccessfulLogin != "" {
-		_ = r.Helmer.RegistryLogout(registrySuccessfulLogin)
+		logoutErr := r.Helmer.RegistryLogout(registrySuccessfulLogin)
+		if logoutErr != nil {
+			return ctrl.Result{}, errors.WithMessage(err, "failed to logout from helm registry: "+registrySuccessfulLogin)
+		}
 	}
 	if err != nil {
 		return ctrl.Result{}, errors.WithMessage(err, chartSpec.Name+": failed chart pull")
--- a/manager/controllers/app/blueprint_controller.go
+++ b/manager/controllers/app/blueprint_controller.go
@@ -196,7 +196,7 @@ func (r *BlueprintReconciler) applyChartResource(ctx context.Context, log logr.L
 	err := r.Helmer.ChartPull(chartSpec.Name)
 	// if we logged into a registry, let us try to logout
 	if registrySuccessfulLogin != "" {
-		r.Helmer.RegistryLogout(registrySuccessfulLogin)
+		_ = r.Helmer.RegistryLogout(registrySuccessfulLogin)
 	}
 	if err != nil {
 		return ctrl.Result{}, errors.WithMessage(err, chartSpec.Name+": failed chart pull")
--- a/manager/controllers/app/blueprint_controller.go
+++ b/manager/controllers/app/blueprint_controller.go
@@ -24,15 +24,20 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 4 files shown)


manager/controllers/app/blueprint_controller.go
```diff
@@ -141,6 +155,42 @@ func (r *BlueprintReconciler) applyChartResource(log logr.Logger, chartSpec app.
 	nbytes, _ := yaml.Marshal(args)
 	log.Info(fmt.Sprintf("--- Values.yaml ---\n\n%s\n\n", nbytes))
 
+	if chartSpec.ChartPullSecret != "" {
+		// obtain ChartPullSecret
+		pullSecret := corev1.Secret{}
+		pullSecrets := []corev1.Secret{}
+
+		if err := r.Get(ctx, types.NamespacedName{Namespace: utils.GetSystemNamespace(), Name: chartSpec.ChartPullSecret},
+			&pullSecret); err == nil {
```
shlomitk1, line 164: Do you have RBAC for secrets in system namespace? Is corev1.Secret mapped to the system namespace in manager cache?

cdoron: hi @shlomitk1 : thank you for helping me confirm that we have RBAC for secrets in system namespace

manager/controllers/app/blueprint_controller.go
```diff
@@ -141,6 +155,42 @@ func (r *BlueprintReconciler) applyChartResource(log logr.Logger, chartSpec app.
...
+		keyring, _ = credentialprovidersecrets.MakeDockerKeyring(pullSecrets, keyring)
+		repoToPull, err := getDomainFromImageName(chartSpec.Name)
+		if err != nil {
+			return ctrl.Result{}, errors.WithMessage(err, chartSpec.Name+": failed to parse image name")
+		}
+
+		creds, withCredentials := keyring.Lookup(chartSpec.Name)
+		if withCredentials {
+			for _, cred := range creds {
+				err := r.Helmer.RegistryLogin(repoToPull, cred.Username, cred.Password, false)
```
revit13, line 182: Does RegistryLogout is also needed at the end?

cdoron: good point. I will address this issue.

manager/controllers/app/blueprint_controller.go
```diff
@@ -130,7 +135,16 @@ func (r *BlueprintReconciler) deleteExternalResources(blueprint *app.Blueprint)
 	return errors.New(strings.Join(errs, "; "))
 }
 
-func (r *BlueprintReconciler) applyChartResource(log logr.Logger, chartSpec app.ChartSpec, args map[string]interface{}, blueprint *app.Blueprint, releaseName string) (ctrl.Result, error) {
+func getDomainFromImageName(image string) (string, error) {
+	named, err := distributionref.ParseNormalizedNamed(image)
+	if err != nil {
+		return "", fmt.Errorf("couldn't parse image name: %v", err)
```
shlomitk1, line 141: Perhaps use errors.WithMessage for consistency?

manager/controllers/app/blueprint_controller.go
```diff
@@ -141,7 +155,53 @@ func (r *BlueprintReconciler) applyChartResource(log logr.Logger, chartSpec app.
...
+
+		if err := r.Get(ctx, types.NamespacedName{Namespace: utils.GetSystemNamespace(), Name: chartSpec.ChartPullSecret},
+			&pullSecret); err == nil {
+			// if this is not a dockerconfigjson, ignore
+			if pullSecret.Type == "kubernetes.io/dockerconfigjson" {
+				pullSecrets = append(pullSecrets, pullSecret)
+			}
+		}
+
+		// create a keyring of all dockerconfigjson secrets, to be used for lookup
```
shlomitk1, line 173: At this point pullSecrets can be empty. You could try to pull chart anyway, but the credentialprovider functionality is not required in this case, right?

manager/controllers/app/blueprint_controller.go
```diff
@@ -141,7 +155,53 @@ func (r *BlueprintReconciler) applyChartResource(log logr.Logger, chartSpec app.
...
+
+		creds, withCredentials := keyring.Lookup(chartSpec.Name)
+		if withCredentials {
+			for _, cred := range creds {
+				err := r.Helmer.RegistryLogin(repoToPull, cred.Username, cred.Password, false)
+				if err == nil {
+					registrySuccessfulLogin = repoToPull
+					break
+				} else {
+					log.Info("Failed to login to helm registry: " + repoToPull)
```
shlomitk1, line 189: Is this a real error? It's just one cred from a range. Perhaps to check for non-empty registrySuccessfulLogin right after the for loop?

manager/controllers/app/blueprint_controller.go
```diff
@@ -141,7 +155,53 @@ func (r *BlueprintReconciler) applyChartResource(log logr.Logger, chartSpec app.
...
+				}
+			}
+		} else {
+			log.Info("Could not find ChartPullSecret. Will try to pull chart anyway")
```
shlomitk1, line 193: The printout is misleading. The error means that the required credentials were not supplied.

manager/controllers/app/blueprint_controller.go
cdoron: right. I will try to simplify

manager/controllers/app/blueprint_controller.go
cdoron: not an error, just an info message.

shlomitk1 (approved):

revit13 (approved):

PR #958, opened 2021-11-15
Fix integration-tests

shlomitk1:

This PR replaces ghcr.io with localhost:5000 for using fybrik-template module in the tests.


shlomitk1:
This does not affect tekton tests.

PR #960, opened 2021-11-16
Use revive instead of deprecated golint

froesef:

### DIFF
--- a/connectors/opa/lib/opa_reader_test.go
+++ b/connectors/opa/lib/opa_reader_test.go
@@ -10,7 +10,6 @@ import (
 	"time"
 
 	tu "fybrik.io/fybrik/connectors/opa/testutil"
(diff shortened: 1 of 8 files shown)


shlomitk1 (approved):

roee88:
We should consider using https://github.com/marketplace/actions/revive-action for its GitHub Annotations support

froesef:
There is also the general [golangci-lint](https://github.com/golangci/golangci-lint-action) action. This also generates annotations and should hopefully also be able to include revive. Our `make verify` code currently does a bit more than just run the linter action. (it also formats the code and tidies up modules) This is probably not correct as the user commit should already format the code correctly and tidy up modules. (So that other users have a clean experience when they check out the code and build and not some formatting from a previous PR)

PR #974, opened 2021-11-22
Add conditions to FybrikModule

shlomitk1:

Fixes https://github.com/fybrik/fybrik/issues/966

This PR defines a Ready condition with information such as status, message and observed generation.


charts/fybrik-crd/templates/app.fybrik.io_fybrikmodules.yaml
```diff
@@ -273,19 +273,32 @@ spec:
...
+                      description: Message contains the details of the current condition
+                      type: string
+                    observedGeneration:
+                      description: ObservedGeneration is the version of the resource
+                        for which the condition has been evaluated
+                      format: int64
+                      type: integer
+                    status:
+                      description: 'Status of the condition: true or false'
+                      type: string
```
roytman, line 293: If Condition status can have only 3 values: "True", "False" and "Unknown", should we add a string validation, something like: "pattern: "False|True|Unknown"

shlomitk1: For the best of my knowledge, we do not validate fields inside the status, only in spec.
I've used condition status similar to described in https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md
I'll change the description to include Unknown.

roytman: Actually, there is a better option, here is an example from `issuers.cert-manager.io`
```yaml
status:
  description: Status of the condition, one of (`True`, `False`, `Unknown`).
  enum:
  - "True"
  - "False"
  - Unknown
  type: string
```

manager/controllers/app/fybrikmodule_controller.go
```diff
@@ -54,22 +55,27 @@ func (r *FybrikModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 	observedStatus := moduleContext.Status.DeepCopy()
 	moduleVersion := moduleContext.GetGeneration()
+	if len(moduleContext.Status.Conditions) == 0 {
+		moduleContext.Status.Conditions = []api.Condition{{Type: api.ReadyCondition, Status: corev1.ConditionUnknown, ObservedGeneration: 0}}
+	}
 
 	// check if module has been validated before or if validated module is outdated
-	if string(moduleContext.Status.ValidModule) == "" || observedStatus.ValidatedGeneration != moduleVersion {
+	condition := moduleContext.Status.Conditions[0]
```
revit13, line 63: can `ReadyConditionIndex`  be used instead of 0?

revit13, line 63: where is the condition type is checked? it seems to be updated later without checking the type.

manager/apis/app/v1alpha1/condition_types.go
```diff
@@ -0,0 +1,32 @@
...
+	ErrorCondition ConditionType = "Error"
+	DenyCondition  ConditionType = "Deny"
+	ReadyCondition ConditionType = "Ready"
+)
+
+// Condition describes the state of a FybrikApplication at a certain point.
+type Condition struct {
+	// Type of the condition
+	Type ConditionType `json:"type"`
+	// Status of the condition: true or false
```
roytman, line 24: Should we add something like:
// +kubebuilder:validation:Optional
// +kubebuilder:validation:Enum=True,False,Unknown
 // +kubebuilder:default:=Unknown

manager/apis/app/v1alpha1/fybrikapplication_types.go
```diff
@@ -102,38 +102,6 @@ const (
...
-	// ReadyCondition means that access to a dataset is granted
-	ReadyCondition ConditionType = "Ready"
-)
-
-// Condition describes the state of a FybrikApplication at a certain point.
-type Condition struct {
-	// Type of the condition
-	Type ConditionType `json:"type"`
-	// Status of the condition: true or false
-	Status corev1.ConditionStatus `json:"status"`
```
roytman, line 131: as above

charts/fybrik-crd/templates/app.fybrik.io_fybrikmodules.yaml
shlomitk1: OK

manager/controllers/app/fybrikmodule_controller.go
shlomitk1: Condition type is not checked. Conditions are assumed to be static, i.e. having the same indices in the array over the run.

manager/apis/app/v1alpha1/condition_types.go
shlomitk1: Added

manager/controllers/app/fybrikmodule_controller.go
revit13: OK. does it make sense to use `ReadyConditionIndex`  instead of 0?

revit13 (approved):

PR #975, opened 2021-11-22
Fybrik Module Taxonomy Updates

eletonia:

- adds `FybrikModule` Status permissions to `fybrik-system-role`
- changes `additionalProperties` to `true` for `Interface` object so that Fybrik Module validation will be compliant with quickstart module and allow `endpoint` field
- changes `intent` value in notebook example `FybrikApplication` to be aligned with taxonomy values for `intent`
- removes kubebuilder validation for `CapabilityType` in `FybrikModuleTypes.go` since this is being validated by the FybrikModule taxonomy

### DIFF
--- a/manager/apis/app/v1alpha1/fybrikmodule_types.go
+++ b/manager/apis/app/v1alpha1/fybrikmodule_types.go
@@ -42,6 +42,20 @@ const (
 // ModuleCapability indicates at a high level what is being performed - often on a datset, but potentially other things as well
 type CapabilityType string
 
+const (
+	// Copy moves data from one location to another - i.e implicit copy
+	Copy CapabilityType = "copy"
+
+	// Write is accessed from within an application, typically through an SDK
+	Write CapabilityType = "write"
+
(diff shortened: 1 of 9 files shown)


manager/apis/app/v1alpha1/fybrikmodule_types.go
```diff
@@ -40,24 +40,8 @@ const (
...
-	// Write is accessed from within an application, typically through an SDK
-	Write CapabilityType = "write"
-
-	// Read is accessed from within an application, typically through an SDK
-	Read CapabilityType = "read"
-
-	// Transform processes and changes data
-	Transform CapabilityType = "transform"
-)
-
```
shlomitk1, line 60: Do not remove the constants.

eletonia: Ok I kept the constants and only removed the kubebuilder validation for `CapabilityType`

shlomitk1 (approved):

Issue #793, opened 2021-09-01
Ability to create Data Access modules in different namespaces


ronenkat:
### Discussed in https://github.com/fybrik/fybrik/discussions/774

<div type='discussions-op-text'>

<sup>Originally posted by **rachit1arora** August 11, 2021</sup>
A toggle option as a configuration parameter for fybrik which can determine where data access module can be deployed. Whether it can be deployed in a pre configured namespace or it can be deployed in a new namespace 

Motivation for the requirment: If fybrik is being used in a multi tenant kind of environment where multiple tenants will be creating fybrik applications and for each application we will be starting a data access module the resources used in each data access module should be isolated from each other and one way to isolate them is by creating them in separate tenant specific namespaces and each namespace can have  some network polices also added for achieving right isolation. 

In current form of fybrik it can be data access modules can only be created in a single namespace. 

Some additional thought for this requirement
- Blueprint CRs to be created in one system namespace and not in the namespace where data access module will actually be started
- Blueprint Controller can rely on another component to get the require role and role binding to deploy resources in a new namespace which is created for the data access module
- We can use either a separate spec attribute or a label in fybrik application which can be used to determine the name of new namespace in which data access module need to be started. 

@roee88 @laurieaw63 Please add your thoughts and comments on this requirement
</div>

shlomitk1:
Summary of the discussion with @roee88 and @laurieaw63:
-  Blueprint will be applied in the system namespace 
-  Blueprint spec will include a field `module_namespace`. This field will be used as .Release.Namespace for installing helm charts of modules. 
- Fybrik configuration (fybrik_config) will include a boolean option `manage_module_namespace`.
- When this option is set to false:
	- Module namespace name is taken from the environment, as in the current code base.
	- It will propagate to the Blueprint spec.
	- Existing role/rolebindings for the blueprint and blueprint namespace should be modified accordingly (use the configuration option as a condition)
- When this option is set to true: 
	- Module namespace name will be defined in a FybrikApplication label and propagated into the Blueprint spec. 
	- Validation of FybrikApplication will check the label presence.
        - Blueprint controller will manage the life cycle of the module namespace, i.e. create a new namespace prior to installing modules and delete the namespace when Blueprint is deleted.    
        - Assuming RBAC permissions in the module namespace are set by another component (TBD) 
- Managing the namespace creation/deletion:
	-  Fybrik configuration will include a templated yaml resource, e.g. corev1.Namespace where the namespace name is templated. 
[...]

PR #978, opened 2021-11-22
Refers to #793. Add ModulesNamespace to Blueprint spec

YingchaoDu:
This PR adds a field ModulesNamespace to the Blueprint spec and uses it for helm chart installs.


manager/controllers/app/fybrikapplication_controller_test.go
```diff
@@ -139,6 +139,7 @@ var _ = Describe("FybrikApplication Controller", func() {
 			Eventually(func() error {
 				return k8sClient.Get(context.Background(), blueprintObjectKey, blueprint)
 			}, timeout, interval).Should(Succeed(), "Blueprint has not been created")
+			Expect(blueprint.Spec.ModulesNamespace).To(Equal("fybrik-blueprints"))
```
shlomitk1, line 142: Use utils.GetDefaultModulesNamespace() instead of "fybrik-blueprints"

PR #985, opened 2021-11-24
Change ClusterManager implementations to local package visability

### LINKED ISSUE #988: Code style of the multicluster package and its sub-packages
The multicluster package and its sub-packages potentially have several code-style issues. 

- All ClusterManager implementations are named ClusterManager, which complicates the packages reading. I understood that it was done to prevent name duplication between package name and struct name, e.g. razee.RazeeClusterManager. However all implementations (except one from the dummy package) should not be public, so by reducing their visibility, we add robustness and resolve name conflicts. 
All instances of ClusterManager can be created by dedicated `NewXXXClusterManager` functions

- Cluster creation from config map is done in Razee and in local sub-packages, so both implementations should be combined into a single function.

- Several constant can be defined

- imports can be rearanged according to common open source convention: standard libraries; third -party modules, local repo imports.

roytman:
- Rename ClusterManager implementation constructs to explicit implementation names

with package visibility, e.g. for Razee from ClusterManager to razeeCluaterManager. Add methods to create different ClusterMnager implementations, NewXXXClusterManager.

- Add a method to create a cluster from ConfigMap. (it's used by several ClusterManager implementations)

- Add several constants

- Fix order of imports

/fixes #988

### DIFF
--- a/pkg/multicluster/local/local_cluster_manager.go
+++ b/pkg/multicluster/local/local_cluster_manager.go
@@ -8,49 +8,42 @@ import (
 	"fmt"
 
 	"emperror.dev/errors"
-	"fybrik.io/fybrik/manager/apis/app/v1alpha1"
-	"fybrik.io/fybrik/pkg/multicluster"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"fybrik.io/fybrik/manager/apis/app/v1alpha1"
+	"fybrik.io/fybrik/pkg/multicluster"
 )
 
 const (
 	clusterMetadataConfigmapName string = "cluster-metadata"
 )
 
-// ClusterManager for local cluster configuration
-type ClusterManager struct {
+// localClusterManager for local cluster configuration
+type localClusterManager struct {
 	Client    client.Client
 	Namespace string
 }
 
 // GetClusters returns a list of registered clusters
-func (cm *ClusterManager) GetClusters() ([]multicluster.Cluster, error) {
-	clusterMetadataConfigmap := corev1.ConfigMap{}
+func (cm *localClusterManager) GetClusters() ([]multicluster.Cluster, error) {
+	cmcm := corev1.ConfigMap{}
 	namespacedName := client.ObjectKey{
 		Name:      clusterMetadataConfigmapName,
 		Namespace: cm.Namespace,
 	}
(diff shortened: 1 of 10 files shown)


roee88:
FYI the original naming was due to linting error (since the prefix is redundant with the package name). Not sure if that still applies to your changes, I only scattered quickly.

pkg/multicluster/local/local_cluster_manager.go
```diff
@@ -8,49 +8,42 @@ import (
...
 	"emperror.dev/errors"
-	"fybrik.io/fybrik/manager/apis/app/v1alpha1"
-	"fybrik.io/fybrik/pkg/multicluster"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"fybrik.io/fybrik/manager/apis/app/v1alpha1"
+	"fybrik.io/fybrik/pkg/multicluster"
```
shlomitk1, line 17: Why is the order of imports changing? Shouldn't it be sorted lexicographically?

roytman: In most of OpenSource projects that I've seen, there is an imports separation between standard golang libraries, third parties and imports from the same repository.
For example, her is a file from  kubernetes-sigs /kubefed  https://github.com/kubernetes-sigs/kubefed/blob/master/pkg/controller/schedulingmanager/controller.go.

roytman:
> FYI the original naming was due to linting error (since the prefix is redundant with the package name). Not sure if that still applies to your changes, I only scattered quickly.

golint is happy now

pkg/multicluster/local/local_cluster_manager.go
```diff
@@ -8,49 +8,42 @@ import (
...
 )
 
 const (
 	clusterMetadataConfigmapName string = "cluster-metadata"
 )
 
-// ClusterManager for local cluster configuration
-type ClusterManager struct {
+// localClusterManager for local cluster configuration
+type localClusterManager struct {
```
froesef, line 25: In the past we refrained from using names like `localClusterManager` because the name is already in the package name and should not be repeated in the type definition.

roytman: In general you're right, but here it is a package visibility name, so it will not be used from any other packages with its package name. Therefore golint doesn't complain about it.

When all ClusterManager implementations named ClusterManager it reminds Java obfuscation ;-)

ronenkat (commented):
@roytman looks ok, but I want @tomersolomon1 to review prior to merging.

tomersolomon1 (approved):

ronenkat (approved):

Issue #983, opened 2021-11-23
Data Plane: Generate log information in a standardized manner


simanadler:
### Discussed in https://github.com/fybrik/fybrik/discussions/890

Create a single log structure to be used by both the control plane and the data plane components as is described in the [visibility high level design document](https://docs.google.com/document/d/1x3Jnp3CIN6YhJKtWM1uIimaw9-q6Hb9M7LqkazmBRBo/edit?usp=sharing).

- [ ] Change the arrow flight service to generate log entries according to the new structure
- [ ] Change the data movement operator to generate log entries according to the new structure

Dependent on #926

roytman:
@simanadler, @cdoron, can we close it?

PR #989, opened 2021-11-28
DRAFT: Fybrik Logging

simanadler:
Related to #983.  Introduces the logging structure and helper functions.
No changes yet to the manager to use these.

**Update:** Moved to zerolog which provides structured logging in json format.  Updated existing logging in fybrikapplication_controller.go to show how to use zerolog for fybrik logging.


manager/controllers/utils/utils.go
```diff
@@ -153,3 +153,15 @@ func Intersection(set1 []string, set2 []string) []string {
...
+
+const FYBRIKAPPGUID = "FybrikApplicationGUID"
+
+// GetFybrikApplicationGUID returns a globally unique ID for the FybrikApplication instance.
+// It must be unique over time and across clusters, even after the instance has been deleted, because this ID will be used for logging purposes.
+func GetFybrikApplicationGUID(fapp *app.FybrikApplication) string {
+
+	// Use the clusterwise unique kubernetes id.
+	// No need to add cluster because FybrikApplication instances can only be created on the coordinator cluster.
+	return string(string(fapp.GetObjectMeta().GetUID()))
```
shlomitk1, line 165: Why double string conversion?

manager/controllers/utils/utils.go
```diff
@@ -153,3 +153,15 @@ func Intersection(set1 []string, set2 []string) []string {
 	}
 	return res
 }
+
+const FYBRIKAPPGUID = "FybrikApplicationGUID"
```
roytman, line 157: should we replace GUID to UUID ? 
GUID is more Microsoft style, golnag usually uses UUID.

pkg/logging/logging.go
```diff
@@ -0,0 +1,256 @@
...
+	"github.com/go-logr/logr"
+)
+
+// LogType indicates the types of log entries supported.  Logging packages typically provide separate function
+// calls for the different log types.
+type LogType string
+
+const (
+	ERROR LogType = "error" // Errors encountered in the code
+	INFO  LogType = "info"  // Info logs are things you want to tell the user or administrator which are not errors.
```
roytman, line 18: We need more log types, e.g. "audit"

roytman (commented):
A couple of questions:

- why do we need to reimplement logger, there are several golang loggers that support JSON format, see https://betterprogramming.pub/how-to-use-structured-json-logging-in-golang-applications-7fc5e2751dbd

- There is a very small number of debug log levels, most of golang applications, including K8s itself support a wide range of Info (Debug) levels. (for 0 to 8, and even more)

- Currently fybrik controllers use controller-runtime logger, do you plan to replace it, with this logger component ?

- BTW for developers I'd suggest using an existing golang logger, so developer will not have to work with different log formats (fybrik and K8s),  and a developer can define any number of log levels.
"github.com/go-logr/logr" has a method `WithValues` it allows to add for example `FybrikApplicationGUID` to a new logger, and all its messages will contain the FybrikApplicationGUID.

pkg/logging/logging.go
```diff
@@ -0,0 +1,256 @@
...
+	MANDATORY VerbosityLevel = 0
+	DEBUG     VerbosityLevel = 1
+	TRACE     VerbosityLevel = 2
+)
+
+// LogEntry defines the information for a single entry stored in the system logs.
+// It is assumed that the timestamp is added automatically and thus not included in the log entry structure.
+// Verbosity and log type are added by the log library via which the entries are sent to stdout or stderr.
+type LogEntry struct {
+	// Caller indicates the cluster, component and function that generated the message in the format "cluster/component/function"
```
shlomitk1, line 34: I think that logging the function is not necessary. This makes the filtering by component more difficult, depends on whether the logging was done by the inner function or the outer function receiving the response, might break after code refactoring and makes the caller string very long. It does not add extra value for developers either.

pkg/logging/logging.go
```diff
@@ -0,0 +1,256 @@
...
+const (
+	ERROR LogType = "error" // Errors encountered in the code
+	INFO  LogType = "info"  // Info logs are things you want to tell the user or administrator which are not errors.
+)
+
+// VerbosityLevel allows one to decide how "chatty" you want your logs to be.
+type VerbosityLevel int
+
+const (
+	MANDATORY VerbosityLevel = 0
```
shlomitk1, line 25: Why not just int? It will be less conversions to make...

pkg/logging/logging.go
```diff
@@ -0,0 +1,256 @@
...
+	Message string `json:"message"`
+
+	// Unique id of the kubernetes fybrikapplication (Needs to be unique over time in order to support history and not just current workloads)
+	// +required
+	FybrikAppGUID string `json:"fybrikAppGuid"`
+
+	// ForUser should be true if the message should be shared with the end user in the FybrikApplication status.
+	// Assumed false if not indicated.
+	// +optional
+	ForUser bool `json:"forUser,omitempty"`
```
shlomitk1, line 50: Not sure I understand. If the message should be shared with the end user in the status, manager will update the status. It is not as if manager will collect the logs, analyze them and update the status when ForUser is true.

pkg/logging/logging.go
```diff
@@ -0,0 +1,256 @@
...
+
+	// ForArchive should be true if the message should be archived long term.
+	// For example, contains full contents of FybrikApplication and its status and should be stored for auditing purposes.
+	// Assumed false if not indicated
+	// +optional
+	ForArchive bool `json:"forArchive,omitempty"`
```
shlomitk1, line 56: I suggest to rename to Archive.

pkg/logging/logging.go
```diff
@@ -0,0 +1,256 @@
...
+
+	// ResponseTime of the current operation in milliseconds.
+	// This is an optional parameter that can use in monitoring dashboards such as Kibana.
+	// +optional
+	ResponseTime int64 `json:"responseTime,omitempty"`
+
+	// Action is the current operation being called.
+	// For example, “create_plotter” or “update_blueprint”.
+	// +optional
+	Action string `json:"action,omitempty"`
```
shlomitk1, line 71: I think it is redundant and will be a part of the message. For example, "failed to update blueprint: <error>"

pkg/logging/logging.go
shlomitk1: Shouldn't ARCHIVE be a logType like INFO?

roytman: I don't think it is fybrik developer's decision to archive log or not. Probably we should flag or log messages or create a separate log target for auditing. The retention policy is out of our scope.

manager/controllers/utils/utils.go
simanadler: typo - thanks for catching!

simanadler:
> A couple of questions:
> 
>     * why do we need to reimplement logger, there are several golang loggers that support JSON format, see https://betterprogramming.pub/how-to-use-structured-json-logging-in-golang-applications-7fc5e2751dbd

@roytman We are not re-implementing the logger but rather wrapping an existing one (github.com/go-logr/logr) in order to generate consistent log entries across components (control plane and data plane)

> 
>     * There is a very small number of debug log levels, most of golang applications, including K8s itself support a wide range of Info (Debug) levels. (for 0 to 8, and even more)

I adopted the approach recommended by github.com/go-logr/logr, which is based on [Dave Cheney's approach](https://dave.cheney.net/2015/11/05/lets-talk-about-logging) as described in their readme.
> 
>     * Currently fybrik controllers use controller-runtime logger, do you plan to replace it, with this logger component ?

Currently the fybrik manager uses github.com/go-logr/logr.  The connectors use the standard golang log library.  I proposing to have all control plane golang components use the wrapper I created for go-logr.

> 
>     * BTW for developers I'd suggest using an existing golang logger, so developer will not have to work with different log formats (fybrik and K8s),  and a developer can define any number of log levels.
>       "github.com/go-logr/logr" has a method `WithValues` it allows to add for example `FybrikApplicationGUID` to a new logger, and all its messages will contain the FybrikApplicationGUID.

Yes, WithValues allows you to create key/value pairs.  It's one of the things I considered.  However, based on input from IBM product colleagues and our RedHat colleagues the recommendation was to use json format log entries.

manager/controllers/utils/utils.go
simanadler: sure

pkg/logging/logging.go
simanadler: I adopted the approach recommended by github.com/go-logr/logr, which is based on [Dave Cheney's approach](https://dave.cheney.net/2015/11/05/lets-talk-about-logging) as described in their readme.

pkg/logging/logging.go
simanadler: @shlomitk1 Don't you think it has value for tracing purposes?

pkg/logging/logging.go
simanadler: @shlomitk1 True, but it's less clear when reading the code.

pkg/logging/logging.go
simanadler: @shlomitk1 This is not meant to replace the status returned in fybrikapplication.status.  It duplicates it, so that if a customer deploys logging tools like logDNA or fluentd, etc they can filter and act on things at the log level.  See the different uses of the logs from the perspective of the different actors as defined in https://docs.google.com/document/d/1x3Jnp3CIN6YhJKtWM1uIimaw9-q6Hb9M7LqkazmBRBo/edit#heading=h.z3iyvac259h6

pkg/logging/logging.go
simanadler: I have to admit that I debated the approach to take here as well.  There are 3 options: (a) have a flag in the log entry structure as I've done here, (b) have a verbosity level specific to archive, or (c) have a separate log for auditing.  

All 3 options require the developer to indicate that it's something that needs to be archived.  He/she of course does not deal with retention policies.  That's out of scope for fybrik, and is handled by those deploying log processing tools.

I did not choose (c) because I think it's preferable that fybrik continue to write (via go-logr) only to std_err and std_out and let the log aggregator filter and route entries as desired rather than putting that logic in fybrik.  

I did not choose (b) because I thought we might want to have the flexibility to combine an archive field and verbosity, meaning that in essence you can also change the verbosity of what gets archived if desired.

@shlomitk1 I'm fine with changing to Archive

pkg/logging/logging.go
simanadler: @shlomitk1 By action I meant things like create, update, delete.  These are relevant especially from within reconcile.

roytman:
> Yes, WithValues allows you to create key/value pairs. It's one of the things I considered. However, based on input from IBM product colleagues and our RedHat colleagues the recommendation was to use json format log entries.

I'm not sure that they mean the debug logs. All logs in the OpenShift CNI is a plain text. Sould we revalidate it ?

pkg/logging/logging.go
roytman: @simanadler I agree with the chosen option (a), but my point that it is not fybrik responsibility to decide what should be archived or not, therefore I just suggest using a flag "Audit" or something similar. 
Somebody else will decide if he wants to archive it or not.

pkg/logging/logging.go
shlomitk1: I don't. First, it is not a full stack, as printed for segmentation fault. Second, for tracing you'd usually need to go to other logs (of the connectors, for example) to trace the problem.

pkg/logging/logging.go
shlomitk1: OK.

pkg/logging/logging.go
```diff
@@ -0,0 +1,256 @@
...
+// ------ User Facing Log Functions ---------
+// Components (control plane and data plane) that interface with the user / user workload should log interactions their USER LEVEL log entries via user level logging functions. LogFAError, LogFAStruct, LogFAUpdate
+// This ensures that they are logged with ForUser and ForArchive set to true
+// Log entries not relevant to users should be logged with the internal logging functions
+
+// LogFAStruct writes to the log file a log entry containing for example, the FybrikApplication structure or a subset (spec or status) of it.
+// The log entry indicates it's a user level entry and should be archived.
+// This could be used by data path components as well to log user level structures.
+// It is logged as INFO/MANDATORY
+func LogFAStruct(log logr.Logger, arg interface{}, argName string, caller string, guid string, cluster string, function string) {
```
shlomitk1, line 83: @simanadler In this and next functions `log` is used. Do you intend to implement Logger interface? Standard Logger implementations, including the one currently used, are not printing info as JSON. 
What Logger implementation will be used for the main manager process? My concern is that we will have a mixed structure in logs that will be not identifiable by external tools.

pkg/logging/logging.go
simanadler: Changed ForArchive to Audit

pkg/logging/logging.go
shlomitk1: @simanadler Do you have an example when ForUser is set to true but ForArchive/Audit isn't? In the wrapper functions either both are true or both are false, unless I missed something.

pkg/logging/logging.go
simanadler: @shlomitk1 Based on the comments to this PR and my discussion with @roytman this morning, I'm looking at moving to https://github.com/rs/zerolog instead of go-logr, since it supports json based structured logging.

pkg/logging/logging.go
simanadler: @shlomitk1 When we add support for reconciling based on policy, metadata or infrastructure updates then we will have situations where Audit (ForArchive) will be true but ForUser will be false.

pkg/logging/logging.go
```diff
@@ -0,0 +1,256 @@
...
+	logEntry := LogEntry{
+		Caller:        buildCallerPath(cluster, caller, function),
+		Message:       "",
+		FybrikAppGUID: guid,
+	}
+	printStructure(arg, argName, log, &logEntry, verbosity)
+}
+
+// ----- INTERNAL FUNCTIONS -------
+// Called only by functions in this package
```
revit13, line 213: consider putting these functions under `internal` sub package.

pkg/logging/logging.go
simanadler: Removed

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -537,18 +568,20 @@ func ownerLabels(id types.NamespacedName) map[string]string {
...
 
 // GetAllModules returns all CRDs of the kind FybrikModule mapped by their name
-func (r *FybrikApplicationReconciler) GetAllModules() (map[string]*api.FybrikModule, error) {
+func (r *FybrikApplicationReconciler) GetAllModules(fybrikappuuid string) (map[string]*api.FybrikModule, error) {
 	ctx := context.Background()
 
 	moduleMap := make(map[string]*api.FybrikModule)
 	var moduleList api.FybrikModuleList
 	if err := r.List(ctx, &moduleList, client.InNamespace(utils.GetSystemNamespace())); err != nil {
-		r.Log.V(0).Info("Error while listing modules: " + err.Error())
```
roytman, line 546: log := r.Log.Str(logging.FYBRIKAPPUUID, fybrikappuuid)

manager/controllers/utils/environmentVariable.go
```diff
@@ -13,6 +13,12 @@ const DefaultBlueprintNamespace = "fybrik-blueprints"
 // Controller namespace defines a namespace where
 const DefaultControllerNamespace = "fybrik-system"
 
+// Returns a verbosity level as an integer based on
+// https://github.com/rs/zerolog#leveled-logging
+func GetLoggingVerbosity() string {
```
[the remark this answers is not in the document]
simanadler: will fix

manager/controllers/utils/utils.go
```diff
@@ -154,14 +154,14 @@ func Intersection(set1 []string, set2 []string) []string {
 	return res
 }
 
-const FYBRIKAPPGUID = "FybrikApplicationGUID"
+const FYBRIKAPPUUID = "FybrikApplicationUUID"
 
 // GetFybrikApplicationGUID returns a globally unique ID for the FybrikApplication instance.
 // It must be unique over time and across clusters, even after the instance has been deleted, because this ID will be used for logging purposes.
-func GetFybrikApplicationGUID(fapp *app.FybrikApplication) string {
+func GetFybrikApplicationUUID(fapp *app.FybrikApplication) string {
```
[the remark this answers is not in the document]
simanadler: The UUID is automatically generated for all kubernetes objects.
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -5,23 +5,26 @@ package app
...
 	"os"
 	"strings"
 	"time"
 
+	"github.com/rs/zerolog"
+
 	"fybrik.io/fybrik/manager/controllers"
 	"fybrik.io/fybrik/pkg/adminconfig"
 	"fybrik.io/fybrik/pkg/environment"
+	"fybrik.io/fybrik/pkg/logging"
```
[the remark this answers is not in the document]
simanadler: @roytman how would you like them organized?

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -59,17 +62,34 @@ type FybrikApplicationReconciler struct {
...
+		annotations := applicationContext.GetAnnotations()
+		if annotations == nil {
+			annotations = map[string]string{}
+			annotations[utils.FYBRIKAPPUUID] = uuid
+			applicationContext.SetAnnotations(annotations)
+		} else {
+			annotations[utils.FYBRIKAPPUUID] = utils.GetFybrikApplicationUUID(applicationContext)
+		}
+	}
+
```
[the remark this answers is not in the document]
simanadler: sure

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -303,9 +323,17 @@ func setReadModulesEndpoints(applicationContext *api.FybrikApplication, flows []
 // reconcile receives either FybrikApplication CRD
 // or a status update from the generated resource
 func (r *FybrikApplicationReconciler) reconcile(applicationContext *api.FybrikApplication) (ctrl.Result, error) {
-	utils.PrintStructure(applicationContext.Spec, r.Log, "FybrikApplication")
-	// Data User created or updated the FybrikApplication
+	uuid := utils.GetFybrikApplicationUUID(applicationContext)
```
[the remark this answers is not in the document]
simanadler: sure

manager/controllers/app/fybrikapplication_controller.go
simanadler: sure

manager/controllers/utils/utils.go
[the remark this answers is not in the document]
roee88: I have missed the discussion where it was decided to use the object uuid. Can you share a reference? Or is it just a placeholder for now?

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -59,17 +62,34 @@ type FybrikApplicationReconciler struct {
...
+	if applicationContext.GetGeneration() == 1 {
+		annotations := applicationContext.GetAnnotations()
+		if annotations == nil {
+			annotations = map[string]string{}
+			annotations[utils.FYBRIKAPPUUID] = uuid
+			applicationContext.SetAnnotations(annotations)
+		} else {
+			annotations[utils.FYBRIKAPPUUID] = utils.GetFybrikApplicationUUID(applicationContext)
+		}
```
roee88, line 88: - Missing a call to applicationContext.SetAnnotations?
- You already got the uuid in line 79, no need to recalculate

manager/controllers/utils/utils.go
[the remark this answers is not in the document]
simanadler: @roee88 Please see the discussion here: https://github.com/fybrik/fybrik/discussions/851

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -85,13 +105,14 @@ func (r *FybrikApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		log.Trace().Str(logging.FYBRIKAPPUUID, uuid).Msg("Reconciler validating Fybrik application")
 		err := applicationContext.ValidateFybrikApplication("/tmp/taxonomy/fybrik_application.json")
-		log.V(0).Info("Reconciler validating Fybrik application")
 		applicationContext.Status.ValidatedGeneration = appVersion
+
 		// if validation fails
 		if err != nil {
 			// set error message
-			log.V(0).Info("Fybrik application validation failed " + err.Error())
+			log.Fatal().Err(err).Str(logging.FYBRIKAPPUUID, uuid).Bool(logging.FORUSER, true).Bool(logging.AUDIT, true).Msg("FybrikApplication valdiation failed")
```
shlomitk1, line 115: Lack of consistency. If a webhook rejects the resource, no audit entry is created.

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -266,7 +286,7 @@ func (r *FybrikApplicationReconciler) deleteExternalResources(applicationContext
 		return nil
 	}
 
-	r.Log.V(0).Info("Reconcile: FybrikApplication is deleting the generated " + applicationContext.Status.Generated.Kind)
+	r.Log.Trace().Str(logging.FYBRIKAPPUUID, utils.GetFybrikApplicationUUID(applicationContext)).Str(logging.ACTION, logging.DELETE).Msg("Reconcile: FybrikApplication is deleting the generated " + applicationContext.Status.Generated.Kind)
```
shlomitk1, line 289: If we register the action (delete, update, create), perhaps we could add Kind as well and reduce this long message in this case, i.e. DELETE Plotter, UPDATE FybrikApplication, etc.

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -303,9 +323,17 @@ func setReadModulesEndpoints(applicationContext *api.FybrikApplication, flows []
...
-	// Data User created or updated the FybrikApplication
+	uuid := utils.GetFybrikApplicationUUID(applicationContext)
 
+	// Log the request received - i.e. the fybrikapplication.spec
+	jsonStruct, err := json.Marshal(applicationContext.Spec)
+	if err != nil {
+		r.Log.Warn().Str(logging.FYBRIKAPPUUID, uuid).Msg("Failed converting fybrikapplication.spec to json")
+	} else {
+		r.Log.Trace().Bool(logging.FORUSER, true).Bool(logging.AUDIT, true).Str(logging.FYBRIKAPPUUID, uuid).RawJSON("FybrikApplication.Spec", jsonStruct)
+	}
```
shlomitk1, line 334: Can this functionality be extracted into a function, making the resource general? Something in the lines of utils.PrintStructure that received interface{} and a string.

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -316,26 +344,26 @@ func (r *FybrikApplicationReconciler) reconcile(applicationContext *api.FybrikAp
...
+		r.Log.Info().Str(logging.FYBRIKAPPUUID, uuid).Msg("No plotter will be generated since no datasets are specified")
 		return ctrl.Result{}, nil
 	}
 
 	// create a list of requirements for creating a data flow (actions, interface to app, data format) per a single data set
 	// workload cluster is common for all datasets in the given application
 	workloadCluster, err := r.GetWorkloadCluster(applicationContext)
 	if err != nil {
-		r.Log.V(0).Info("could not determine in which cluster the workload runs: " + err.Error())
+		r.Log.Fatal().Err(err).Bool(logging.FORUSER, true).Bool(logging.AUDIT, true).Str(logging.FYBRIKAPPUUID, uuid).Str(logging.ACTION, logging.CREATE).Msg("Could not determine in which cluster the workload runs")
```
shlomitk1, line 355: why logging.CREATE? It is not a failure to create a resource.

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -316,26 +344,26 @@ func (r *FybrikApplicationReconciler) reconcile(applicationContext *api.FybrikAp
...
 		return ctrl.Result{}, err
 	}
 	var requirements []DataInfo
 	for _, dataset := range applicationContext.Spec.Data {
 		req := DataInfo{
 			Context: dataset.DeepCopy(),
 		}
-		r.Log.V(0).Info("Preparing requirements for " + req.Context.DataSetID)
+		r.Log.Trace().Str(logging.FYBRIKAPPUUID, uuid).Str(logging.DATASETID, req.Context.DataSetID).Str(logging.ACTION, logging.CREATE).Msg("Preparing requirements for dataset")
```
shlomitk1, line 363: It is not a create action. All controller functionality is about constructing a Plotter, does not seem right to have a create action for all printouts.

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -364,16 +392,18 @@ func (r *FybrikApplicationReconciler) reconcile(applicationContext *api.FybrikAp
 	ownerRef := &api.ResourceReference{Name: applicationContext.Name, Namespace: applicationContext.Namespace, AppVersion: applicationContext.GetGeneration()}
 
 	resourceRef := r.ResourceInterface.CreateResourceReference(ownerRef)
-	if err := r.ResourceInterface.CreateOrUpdateResource(ownerRef, resourceRef, plotterSpec, applicationContext.Labels); err != nil {
-		r.Log.V(0).Info("Error creating " + resourceRef.Kind + " : " + err.Error())
+	if err := r.ResourceInterface.CreateOrUpdateResource(ownerRef, resourceRef, plotterSpec, applicationContext.Labels, uuid); err != nil {
+		r.Log.Fatal().Err(err).Str(logging.FYBRIKAPPUUID, uuid).Bool(logging.FORUSER, true).Bool(logging.AUDIT, true).Msg("Error creating " + resourceRef.Kind)
```
shlomitk1, line 396: Action is missing here. CREATE + plotter

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -441,8 +471,8 @@ func (r *FybrikApplicationReconciler) GetWorkloadCluster(application *api.Fybrik
 		if application.Spec.Selector.WorkloadSelector.Size() == 0 {
 			return multicluster.Cluster{}, nil
 		}
-		// the workload runs in a local cluster
-		r.Log.V(0).Info("selector.clusterName field is not specified for an existing workload - a local cluster is assumed")
+		r.Log.Warn().Err(errors.New("selector.clusterName field is not specified")).Str(logging.FYBRIKAPPUUID, utils.GetFybrikApplicationUUID(application)).Str(logging.ACTION, logging.CREATE).Msg("No workload cluster indicated, so a local cluster is assumed")
```
shlomitk1, line 474: It is not an error. Cluster of the workload does not have to be specified if the workload runs on a local cluster.

manager/controllers/app/fybrikapplication_controller.go
simanadler: Fixed

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -467,13 +497,14 @@ func (r *FybrikApplicationReconciler) GetWorkloadCluster(application *api.Fybrik
...
 // NewFybrikApplicationReconciler creates a new reconciler for FybrikApplications
+// A logger is created that logs the standard
 func NewFybrikApplicationReconciler(mgr ctrl.Manager, name string,
 	policyManager connectors.PolicyManager, catalog connectors.DataCatalog, cm multicluster.ClusterLister,
 	provision storage.ProvisionInterface, configEvaluator adminconfig.EvaluatorInterface) *FybrikApplicationReconciler {
 	return &FybrikApplicationReconciler{
 		Client:            mgr.GetClient(),
 		Name:              name,
-		Log:               ctrl.Log.WithName("controllers").WithName(name),
+		Log:               logging.LogInit(logging.CONTROLLER, name, "Coordinator Cluster"), // TODO - get the real cluster name
```
shlomitk1, line 507: You can't know the cluster name at this point. The main process has not finished yet to initialize cache, config, etc. so trying to read a configmap at this point won't work. Cluster can be initialized later, at the first reconcile.

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -537,18 +568,20 @@ func ownerLabels(id types.NamespacedName) map[string]string {
...
 // GetAllModules returns all CRDs of the kind FybrikModule mapped by their name
-func (r *FybrikApplicationReconciler) GetAllModules() (map[string]*api.FybrikModule, error) {
+func (r *FybrikApplicationReconciler) GetAllModules(fybrikappuuid string) (map[string]*api.FybrikModule, error) {
 	ctx := context.Background()
 
 	moduleMap := make(map[string]*api.FybrikModule)
 	var moduleList api.FybrikModuleList
 	if err := r.List(ctx, &moduleList, client.InNamespace(utils.GetSystemNamespace())); err != nil {
-		r.Log.V(0).Info("Error while listing modules: " + err.Error())
+		r.Log.Error().Err(err).Str(logging.FYBRIKAPPUUID, fybrikappuuid).Bool(logging.AUDIT, true).Msg("Error while listing modules")
```
shlomitk1, line 577: Action == LIST

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -59,17 +62,34 @@ type FybrikApplicationReconciler struct {
...
 func (r *FybrikApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	log := r.Log.WithValues("fybrikapplication", req.NamespacedName)
+	log := r.Log.With().Str("fybrikapplication", req.NamespacedName.String()).Logger()
+
 	// obtain FybrikApplication resource
 	applicationContext := &api.FybrikApplication{}
 	if err := r.Get(ctx, req.NamespacedName, applicationContext); err != nil {
-		log.V(0).Info("The reconciled object was not found")
+		// TODO - Should this be fatal rather than error level?
+		log.Error().Err(err).Str(logging.FYBRIKAPPUUID, "Unknown").Msg("The reconciled object was not found")
```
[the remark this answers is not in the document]
simanadler: done

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -593,7 +626,7 @@ func (r *FybrikApplicationReconciler) updateProvisionedStorageStatus(application
 			return false, nil
 		}
 		if !res.Provisioned {
-			r.Log.V(0).Info("No bucket has been provisioned for " + id)
+			r.Log.Error().Err(errors.New(res.ErrorMsg)).Str(logging.FYBRIKAPPUUID, utils.GetFybrikApplicationUUID(applicationContext)).Str(logging.ACTION, logging.CREATE).Str(logging.DATASETID, id).Msg("No bucket has been provisioned for " + id)
```
shlomitk1, line 629: No need for action here.

shlomitk1, line 629: At this point we don't know if it is an error, the request can be in pending state.

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -593,7 +626,7 @@ func (r *FybrikApplicationReconciler) updateProvisionedStorageStatus(application
...
 			// TODO(shlomitk1): analyze the error
 			if res.ErrorMsg != "" {
```
shlomitk1, line 631: Now it's time to log the error.

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -605,13 +638,17 @@ func (r *FybrikApplicationReconciler) updateProvisionedStorageStatus(application
 }
 
 func (r *FybrikApplicationReconciler) buildSolution(applicationContext *api.FybrikApplication, requirements []DataInfo) (map[string]NewAssetInfo, *api.PlotterSpec, error) {
+	uuid := utils.GetFybrikApplicationUUID(applicationContext)
+
 	// get deployed modules
-	moduleMap, err := r.GetAllModules()
+	moduleMap, err := r.GetAllModules(uuid)
```
shlomitk1, line 644: uuid should be a member field of the controller and be updated on each reconcile, rather than be passed to all functions.

manager/controllers/app/fybrikapplication_controller.go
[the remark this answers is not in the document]
roytman: Usually (or at least in the Open Source project where I participated) there is the following imports separating:
```
import ( 
standard go libraries

third-party projects

local repo imports
)
```

manager/controllers/app/fybrikapplication_controller.go
roytman: @shlomitk1 , can the same controller (`FybrikApplicationReconciler`) be concurrently called to reconcile events from different fybrik apps ?

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -59,17 +62,34 @@ type FybrikApplicationReconciler struct {
...
 	if err := r.Get(ctx, req.NamespacedName, applicationContext); err != nil {
-		log.V(0).Info("The reconciled object was not found")
+		// TODO - Should this be fatal rather than error level?
+		log.Error().Err(err).Str(logging.FYBRIKAPPUUID, "Unknown").Msg("The reconciled object was not found")
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+
+	// Add the globally unique identifier for the FybrikApplication instance to the annotations for use in logging
+	// Only need to do this once, and not on every update to this resource
+	uuid := utils.GetFybrikApplicationUUID(applicationContext)
```
shlomitk1, line 79: Why is uuid added to annotations? Where are annotations used and why is this field duplicated in the resource?

manager/controllers/app/fybrikapplication_controller.go
shlomitk1: It can. Right, uuid can't be a global member then.

manager/controllers/app/fybrikapplication_controller.go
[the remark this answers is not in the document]
roytman: @simanadler , sorry zerolog has a slightly different API, see [Sub-loggers](https://github.com/rs/zerolog/blob/0d2f0be4ee595b0f227574d22310abb749e31319/log.go#L27),
or you can consider to use [zapr](https://github.com/go-logr/zapr), which is logr wrapper over logr. My point is to prevent adding Str(logging.FYBRIKAPPUUID, uuid)  to every call.

`import (
	"fmt"

	"github.com/go-logr/logr"
	"github.com/go-logr/zapr"
	"go.uber.org/zap"
)

func main() {
	var log logr.Logger

	zapLog, err := zap.NewProduction()
	if err != nil {
		panic(fmt.Sprintf("who watches the watchmen (%v)?", err))
	}
	log = zapr.NewLogger(zapLog)
	log = log.WithValues("FYBRIKAPPUUID", "myApp")
	log.Info("Logr in action!", "the answer", 42)
	log.Error(fmt.Errorf("myError"), "error test")
	log = log.WithCallDepth(1)
	log.Error(fmt.Errorf("myError2"), "error test 2")
}`
output:

> {"level":"info","ts":1638291742.615371,"caller":"pkg/test.go:21","msg":"Logr in action!","FYBRIKAPPUUID":"myApp","the answer":42}
{"level":"error","ts":1638291742.615975,"caller":"pkg/test.go:22","msg":"error test","FYBRIKAPPUUID":"myApp","error":"myError","stacktrace":"main.main\n\tC:/Users/ALEXEYROYTMAN/goWork/src/github.com/ibm/ovsdb-etcd/pkg/test.go:22\nruntime.main\n\tC:/Go/src/runtime/proc.go:225"}
{"level":"error","ts":1638291742.615975,"caller":"runtime/proc.go:225","msg":"error test 2","FYBRIKAPPUUID":"myApp","error":"myError2","stacktrace":"runtime.main\n\tC:/Go/src/runtime/proc.go:225"}

roytman:
@simanadler , should module developers use our logger, if yes, I'd suggest defining an interface (it can be similar to zerlog, or logr, or something in the middle), an instance of which will be provided to the module. 
Therefore we will be able to change the logger implementation without changing the external API.

shlomitk1:
> @simanadler , should module developers use our logger, if yes, I'd suggest defining an interface (it can be similar to zerlog, or logr, or something in the middle), an instance of which will be provided to the module. Therefore we will be able to change the logger implementation without changing the external API.

I don't see how this is technically possible. Module developers can use any language for implementation, the module logs are printed to a different pod. Unless there is some global component which can be accessed via REST that does the logging, but this will be an overkill, I think.

manager/controllers/app/fybrikapplication_controller.go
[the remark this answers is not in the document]
simanadler: @roytman I will fix in the files I touch, but perhaps this should be a separate issue because it affects all the code.

manager/controllers/app/fybrikapplication_controller.go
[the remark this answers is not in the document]
simanadler: @roytman I fixed so we don't add the uuid on every call.

manager/controllers/app/fybrikapplication_controller.go
simanadler: Yes, done

manager/controllers/app/fybrikapplication_controller.go
simanadler: It's a failure while creating the data plane

manager/controllers/app/fybrikapplication_controller.go
simanadler: I guess we need to come up with guidelines about when to use CREATE, UPDATE, DELETE in the log messages.  Do you have thoughts about this?

manager/controllers/app/fybrikapplication_controller.go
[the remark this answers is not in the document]
simanadler: fixed

manager/controllers/app/fybrikapplication_controller.go
simanadler: Agreed.  That's why it's a warning

manager/controllers/app/fybrikapplication_controller.go
simanadler: Correct.  I removed it completely.  I spoke with @roee88 and we decided not to pass cluster at all.  The logging agent will add the information when it collects the logs.

manager/controllers/app/fybrikapplication_controller.go
simanadler: Changed to warn

manager/controllers/app/fybrikapplication_controller.go
simanadler: uuid will be added to annotations of fybrikapplication, plotter, blueprint and then it needs to be passed to each of the modules in the data plane.

shlomitk1: My point is that uuid can be added to annotations of plotter, blueprint and modules without changing FybrikApplication. 
Two technical concerns I have: 
1. Where do you update FybrikApplication itself, not the status? The only place it's done is inside reconcileFinalizers where finalizers are added for the first generation. 
2. Any update to the resource will cause an extra reconcile loop.

manager/controllers/app/fybrikapplication_controller.go
shlomitk1: My thought was that the action is a k8s action applied to a resource or resources, such as CREATE, UPDATE, DELETE, GET, LIST. This is an important information for admin, developers, audit - each one for a different reason.

manager/controllers/app/fybrikapplication_controller.go
simanadler: @shlomitk1 We could add logging in the webhook

manager/controllers/app/fybrikapplication_controller.go
simanadler: We could do that I guess.  Inside the reconcile functions though we often don't know if we are creating or updating though.  Even the k8s call is createorupdate.

simanadler:
Closing this draft PR, and will submit implementation PR based on the discussions here.

PR #1002, opened 2021-12-05
Avoid pushing fybrik-template chart to public registry.

revit13:
With the recent move of the implicit copy charts to the  [data-movement-operator](https://github.com/fybrik/data-movement-operator/tree/master/modules) repo only the fybrik-template chart is left as a chart we push into the public registry. As this chart is used for testing there is no point in pushing it to the public registry thus this PR removes the use of `helm-push-public`.


roytman (approved):

shlomitk1:
@revit13 Did you replace all occurrences of ghcr.io/fybrik/fybrik-template:0.1.0 in the test yamls?

PR #1003, opened 2021-12-05
Revert "Avoid pushing fybrik-template chart to public registry."

roytman:
We need this almost empty helm chart in a public repository for integration testing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -215,9 +215,9 @@ jobs:
       if: ${{ github.event_name != 'pull_request' }}
       env:
         DOCKER_PUBLIC_TAGNAME:  ${{ steps.version.outputs.version }}
-      run: make docker-retag-and-push-public
+      run: make docker-retag-and-push-public && make helm-push-public
     - name: Publish latest image tag on push tag event
       if: github.event_name != 'pull_request' && steps.version.outputs.push_tag_event == 'true'
       env:
         DOCKER_PUBLIC_TAGNAME: 'latest'
-      run: make docker-retag-and-push-public
+      run: make docker-retag-and-push-public && make helm-push-public
--- a/Makefile
+++ b/Makefile
@@ -118,6 +118,10 @@ endef
 docker-retag-and-push-public:
 	$(call do-docker-retag-and-push-public)
 
+.PHONY: helm-push-public
+helm-push-public:
+	DOCKER_HOSTNAME=${DOCKER_PUBLIC_HOSTNAME} DOCKER_NAMESPACE=${DOCKER_PUBLIC_NAMESPACE} DOCKER_TAGNAME=${DOCKER_PUBLIC_TAGNAME} make -C modules helm-chart-push
+
 .PHONY: save-images
 save-images:
 	docker save -o images.tar ${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}/manager:${DOCKER_TAGNAME} \


revit13:
@roytman I assume you are refering to `name: ghcr.io/fybrik/fybrik-template:0.1.0` from https://github.com/fybrik/fybrik/blob/master/manager/testdata/blueprint.yaml#L56 . I indeed missed that sorry. Can we change it to `name: localhost:5000/fybrik-system/fybrik-template:0.1.0` similar to https://github.com/fybrik/fybrik/blob/master/manager/testdata/unittests/module-read-csv.yaml#L14 so we could avoid pushing this chart to public repo? Thanks

shlomitk1:
> @roytman I assume you are refering to `name: ghcr.io/fybrik/fybrik-template:0.1.0` from https://github.com/fybrik/fybrik/blob/master/manager/testdata/blueprint.yaml#L56 . I indeed missed that sorry. Can we change it to `name: localhost:5000/fybrik-system/fybrik-template:0.1.0` similar to https://github.com/fybrik/fybrik/blob/master/manager/testdata/unittests/module-read-csv.yaml#L14 so we could avoid pushing this chart to public repo? Thanks

@revit13 Yes, it can be renamed to localhost:5000/fybrik-system/fybrik-template:0.1.0

revit13:
@roytman @shlomitk1 ok Thanks I will create a pr for that.

roytman:
@revit13 , @shlomitk1 I understood from @roee88, that this chart is used for the integration tests, and better to check helm chart deployment from a public repository. 
Did I miss something ?

revit13:
@roytman OK I see. @shlomitk1 can we use arrow-flight-module-chart in the integration tests instead of fybrik-template? not sure what version though - maybe `latest`

shlomitk1:
> @roytman OK I see. @shlomitk1 can we use arrow-flight-module-chart in the integration tests instead of fybrik-template? not sure what version though - maybe `latest`

Not latest. Any change in arrow-flight repo will break testing. The last official version that is compatible with Fybrik release, perhaps.

revit13:
> > @roytman OK I see. @shlomitk1 can we use arrow-flight-module-chart in the integration tests instead of fybrik-template? not sure what version though - maybe `latest`
> 
> Not latest. Any change in arrow-flight repo will break testing. The last official version that is compatible with Fybrik release, perhaps.

OK Thanks. I guess  `fybrik-template` chart was created to avoid such compatibility issues...

roee88:
> > > @roytman OK I see. @shlomitk1 can we use arrow-flight-module-chart in the integration tests instead of fybrik-template? not sure what version though - maybe `latest`
> > 
> > 
> > Not latest. Any change in arrow-flight repo will break testing. The last official version that is compatible with Fybrik release, perhaps.
> 
> OK Thanks. I guess `fybrik-template` chart was created to avoid such compatibility issues... on the other hand, this is the purpose of integration tests to reveal such problems. no?

Not in this repo. There are no integration tests with external modules in this repo. The ideas discussed so far were to eventually setup a way for external repos to run their own integration tests against fybrik. If this decision needs to be revisited then we can open a separate discussion.

revit13:
> > > > @roytman OK I see. @shlomitk1 can we use arrow-flight-module-chart in the integration tests instead of fybrik-template? not sure what version though - maybe `latest`
> > > 
> > > 
> > > Not latest. Any change in arrow-flight repo will break testing. The last official version that is compatible with Fybrik release, perhaps.
> > 
> > 
> > OK Thanks. I guess `fybrik-template` chart was created to avoid such compatibility issues... on the other hand, this is the purpose of integration tests to reveal such problems. no?
> 
> Not in this repo. There are no integration tests with external modules in this repo. The ideas discussed so far were to eventually setup a way for external repos to run their own integration tests against fybrik. If this decision needs to be revisited then we can open a separate discussion.

@roee88 OK, I guess https://github.com/fybrik/fybrik/blob/master/manager/testdata/blueprint.yaml#L18 should be changed after the implicit copy modules are moved to their own repo right? I can create an issue for that.

revit13 (approved):

shlomitk1:
> Also, these are used in unit tests not in integration tests. @shlomitk1 after getting back to the previous state (i.e. after this PR is merged) would one be forced to push anything to the local registry before running those unit tests from within an IDE?

Unit tests are not supposed to use helm, so the images do not matter.

roee88:
If the value is ignored, i.e., we have no test that actually helm installs the dummy, then this PR is not needed. I was under the probably wrong impression that we do have some test in the blueprint controller that works with the non-fake helm implementation. Apologies for the confusion.

roytman:
> If the value is ignored, i.e., we have no test that actually helm installs the dummy, then this PR is not needed. I was under the probably wrong impression that we do have some test in the blueprint controller that works with the non-fake helm implementation. Apologies for the confusion.

So we don't have any integration test with a module installation from a real public repository ?

revit13:
a minor comment: `fybrik-template` is pushed to public repo with tag `latest`, `master` or release version tag. I guess that in the tests that use the public chart one of them should be used instead of `0.1.0` tag to avoid making the public push redundant.

shlomitk1:
> > If the value is ignored, i.e., we have no test that actually helm installs the dummy, then this PR is not needed. I was under the probably wrong impression that we do have some test in the blueprint controller that works with the non-fake helm implementation. Apologies for the confusion.
> 
> So we don't have any integration tests with a module installation from a real public repository ?

We have unit tests that run in the fake environment and integration tests that run in the real environment. Tests in xxx_controller_test.go run as both unit tests and integration tests. Blueprint.yaml is not applied in the integration test, blueprint_read.yaml is applied in the integration test as well, and it includes a public image for a module: ghcr.io/fybrik/fybrik-template:0.1.0
On the other hand, e2e test in fybrikapplication_controller-test.go applies the same module image from localhost:5000...

PR #1016, opened 2021-12-08
Katalog connector openapi

tomersolomon1:
WIP - Auto-generate data_catalog.json from the Asset CRD. This PR will also contain the implementation of the openAPI protocol for the Katalog connector (WIP).


connectors/katalog/example/asset.yaml
```diff
@@ -3,27 +3,23 @@ kind: Asset
...
+    owner: Alice
     geography: us-south
     tags:
-    - finance
-    componentsMetadata:
-      nameOrig: 
+      tag: PII
+    columns:
+      - name: c1
         tags:
```
shlomitk1, line 24: Why do tags appear twice?

connectors/katalog/pkg/connector/catalog.go
```diff
@@ -6,6 +6,8 @@ import (
 	"context"
 	"encoding/json"
 
+	"fybrik.io/fybrik/connectors/katalog/pkg/v1alpha1"
+
 	"log"
 
 	connectors "fybrik.io/fybrik/pkg/connectors/protobuf"
```
shlomitk1, line 13: Protobuf used by an open-api version? I thought the idea was to get rid of protos and move to open-api scheme instead.

pkg/slices/contains.go
```diff
@@ -0,0 +1,13 @@
+// Copyright 2021 IBM Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+package slices
+
+func ContainsString(str string, list []string) bool {
```
shlomitk1, line 6: There is a similar function in https://github.com/fybrik/fybrik/blob/ac8ea647b849bb028cbea636e86422507197c734/manager/controllers/utils/utils.go#L50
I agree it belongs in the package but please delete the duplicated one and replace the references accordingly.

connectors/katalog/example/asset.yaml
tomersolomon1: I suppose it's different tags: tags for the entire asset vs tags for specific columns.. But maybe @rohithdv @ritwikc can elaborate, I was following the schema they defined :)

PR #1024, opened 2021-12-12
Fix golint and formatting warnings

roytman:
Fix golint warnings, such as double imports, redundant aliases, e.t.c.

### DIFF
--- a/manager/controllers/app/blueprint_controller.go
+++ b/manager/controllers/app/blueprint_controller.go
@@ -9,36 +9,33 @@ import (
 	"strings"
 	"time"
 
-	"fybrik.io/fybrik/manager/controllers"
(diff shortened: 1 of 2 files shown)


revit13 (approved):

roee88:
I think that the data catalog related files are auto generated and we're not done with that line of work yet.  The plan for those files was to get things working end to end based on generated files and then move to own the code (e.g., fix gosec issues but the same holds for linting). Just FYI. @rohithdv.

roytman:
> I think that the data catalog related files are auto generated and we're not done with that line of work yet. The plan for those files was to get things working end to end based on generated files and then move to own the code (e.g., fix gosec issues but the same holds for linting). Just FYI. @rohithdv.

@roee88 , I changed only controllers files

PR #1036, opened 2021-12-15
Fix for issue 1028

laurieaw63:
Patch default service account for non-kind clusters with registry credentials.
Set modulesNamespace in blueprint testcases.

Signed-off-by: laurieaw laurieaw@us.ibm.com

### DIFF
--- a/manager/controllers/app/blueprint_controller_test.go
+++ b/manager/controllers/app/blueprint_controller_test.go
@@ -24,7 +24,7 @@ func deployBlueprint(namespace string, shouldSucceed bool) {
 
 	// Set the correct namespace
 	blueprint.SetNamespace(namespace)
-	blueprint.Spec.ModulesNamespace = namespace
+	blueprint.Spec.ModulesNamespace = utils.GetDefaultModulesNamespace()
 	fmt.Printf("Blueprint controller unit test - blueprint namespace : %s\n", namespace)
 	blueprintKey := client.ObjectKeyFromObject(blueprint)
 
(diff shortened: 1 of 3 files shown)


manager/controllers/app/blueprint_controller_test.go
```diff
@@ -24,7 +24,8 @@ func deployBlueprint(namespace string, shouldSucceed bool) {
 
 	// Set the correct namespace
 	blueprint.SetNamespace(namespace)
-	fmt.Printf("Blueprint controller unit test - blueprint namespace: %s\n", namespace)
+	blueprint.Spec.ModulesNamespace = namespace
```
shlomitk1, line 27: Use  `utils.GetDefaultModulesNamespace() `here. The `namespace` variable indicates where the blueprint is deployed (should be system namespace for positive behavior, something else for negative behavior). It is not modules' namespace.

manager/controllers/app/blueprint_controller_unit_test.go
```diff
@@ -49,6 +49,7 @@ func TestBlueprintReconcile(t *testing.T) {
 	blueprint, err := readBlueprint("../../testdata/blueprint.yaml")
 	g.Expect(err).To(gomega.BeNil(), "Cannot read blueprint file for test")
 	blueprint.Namespace = blueprintNamespace
+	blueprint.Spec.ModulesNamespace = blueprintNamespace
```
shlomitk1, line 52: Same comment.

laurieaw63: Yup good catch; changed

manager/controllers/app/blueprint_controller_test.go
laurieaw63: Yup good catch; changed

shlomitk1 (approved):

Issue #1031, opened 2021-12-15
[Razee] Missing razee/watch-resource label in cluster-metadata ConfigMap 


revit13:
As noted in https://github.com/razee-io/Razee/blob/master/README.md#architecture-overview the razee/watch-resource label should be added to all resources that we want to monitor. Currently cluster-metadata ConfigMap is deployed without such label when using Razee as cluster manager.

PR #1044, opened 2021-12-20
Add razee/watch-resource label to cluster-metadata ConfigMap

revit13:
Fix issue #1031 
As noted in https://github.com/razee-io/Razee/blob/master/README.md#architecture-overview the razee/watch-resource label should be added to all resources that we want to monitor. Currently, cluster-metadata ConfigMap is deployed without such label when using Razee as the cluster manager.
This PR adds such label.

### DIFF
--- a/charts/fybrik/templates/cluster-metadata-config.yaml
+++ b/charts/fybrik/templates/cluster-metadata-config.yaml
@@ -3,10 +3,8 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: cluster-metadata
-  {{- if eq .Values.global.multiclusterOrchestrator "razee" }}
   labels:
     razee/watch-resource: "debug"
-  {{- end }}
 data:
   ClusterName: {{ required "cluster name must be set" .Values.cluster.name | quote }}
   Region: {{ required "cluster region must be set" .Values.cluster.region | quote }}
--- a/charts/fybrik/templates/cluster-metadata-config.yaml
+++ b/charts/fybrik/templates/cluster-metadata-config.yaml
@@ -3,6 +3,10 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: cluster-metadata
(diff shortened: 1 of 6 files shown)


charts/fybrik/templates/cluster-metadata-config.yaml
```diff
@@ -3,6 +3,10 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: cluster-metadata
+  {{- if eq .Values.global.multiclusetrOrchestrator "razee" }}
```
roee88, line 6: What's wrong with just adding the label if isRazeeEnabled? Why the new value

shlomitk1, line 6: typo: multicluster

charts/fybrik/values.yaml
```diff
@@ -44,6 +44,8 @@ global:
   # zerolog verbosity level 
   # ref: https://github.com/rs/zerolog#leveled-logging
   loggingVerbosity: -1
+  # Multicluster orchestrator used.
+  multiclusetrOrchestrator: ""
```
shlomitk1, line 48: multicluster

charts/fybrik/templates/cluster-metadata-config.yaml
revit13: I thought it might be misleading as it is true only when deploying the coordinator cluster.

revit13: Thanks will fix that.

roee88: I'm not a fan of adding another global attribute. Consider if just always adding the label is harmless, or otherwise if such flag is more appropriate in the `worker` section or even as `enabled` flag in `razee`.

revit13: OK I think that adding the label is harmless.

roytman (approved):

PR #1046, opened 2021-12-21
Add logging messages to setErrorCondition() and setDenyCondition().

cdoron:
These methods already construct error messages to place in application status.
We add these error message to log.

### DIFF
--- a/manager/controllers/app/fybrikapplication_conditions.go
+++ b/manager/controllers/app/fybrikapplication_conditions.go
@@ -4,96 +4,693 @@
 package app
 
 import (
+	"context"
+	"fmt"
+
+	"os"
 	"strings"
+	"time"
+
+	"fybrik.io/fybrik/manager/controllers"
+	"fybrik.io/fybrik/pkg/adminconfig"
+	"fybrik.io/fybrik/pkg/environment"
+	local "fybrik.io/fybrik/pkg/multicluster/local"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+
+	connectors "fybrik.io/fybrik/pkg/connectors/clients"
+	pb "fybrik.io/fybrik/pkg/connectors/protobuf"
+
+	"emperror.dev/errors"
+	"github.com/rs/zerolog"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -203,11 +203,11 @@ func (r *FybrikApplicationReconciler) checkReadiness(applicationContext *api.Fyb
 			if !found {
 				message := "No copy has been created for the asset " + assetID + " required to be registered"
 				log.Info().Str(logging.DATASETID, assetID).Msg(message)
```
shlomitk1, line 205: Redundant.

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -356,7 +356,7 @@ func (r *FybrikApplicationReconciler) reconcile(applicationContext *api.FybrikAp
 			Context: dataset.DeepCopy(),
 		}
 		if err := r.constructDataInfo(&req, applicationContext, workloadCluster); err != nil {
-			AnalyzeError(applicationContext, req.Context.DataSetID, err)
+			AnalyzeError(r.Log, applicationContext, req.Context.DataSetID, err)
 			log.Error().Err(err).Str(logging.DATASETID, req.Context.DataSetID).Msg("Error constructing data info")
```
shlomitk1, line 360: Redundant logging

manager/controllers/app/fybrikapplication_conditions.go
```diff
@@ -43,21 +46,27 @@ func resetAssetState(application *api.FybrikApplication, assetID string) {
 	application.Status.AssetStates[assetID] = api.AssetState{Conditions: conditions}
 }
 
-func setErrorCondition(application *api.FybrikApplication, assetID string, msg string) {
+func setErrorCondition(log zerolog.Logger, application *api.FybrikApplication, assetID string, err error, msg string) {
```
shlomitk1, line 49: In case of an error, msg already contains it. Why do we need both err and msg?

manager/controllers/app/fybrikapplication_controller.go
cdoron: Thanks. I removed this log line. I hope it is okay that we lose the "Error constructing data info" message

manager/controllers/app/fybrikapplication_conditions.go
cdoron: in checkReadiness(), there are cases where there is just a message and no err.
In those cases the err parameter is nil.

shlomitk1: Right. What I mean is that err is not required. Or you could go over the messages and remove err.Error() from them. Otherwise we'd see the err.Error() printed twice in the log.

shlomitk1 (approved):

shlomitk1 (approved):

Issue #1049, opened 2021-12-23
Add taxonomy and API structures as code 


roee88:
During the work on taxonomy we hit some drawbacks of using the generated code, like the inability to properly use these structures in CRDs and difficulties in defining a single source of truth for the definitions.

To address these and simplify the taxonomy related parts of the repo:

- [x] Add the taxonomy and API structures as code to `pkg/model` (i.e. the model code) #1051
- [x] Use the model code to generate the JSON schemas (base taxonomy.json, catalog and policy manager)  https://github.com/fybrik/fybrik/pull/1060
- [x] Use the model code in manager #1054
- [x] Use the model code in opa connector
- [x] Use the model code in katalog connector
- [x] Remove the old generated code https://github.com/fybrik/fybrik/pull/1061
- [ ] Update docs
- [x] Change the `taxonomy validate` command to validate custom `taxonomy.json` files (and replace the legacy unit test for it) https://github.com/fybrik/fybrik/pull/1065
- [x] Remove the use of openapi2crd and archive its repo (partially https://github.com/fybrik/fybrik/pull/1065)
- [x] Remove the generate target from manager/Makefile https://github.com/fybrik/fybrik/pull/1082
- [ ] Review all taxonomy model definitions

Later we can do the same for CRDs
- #1091

Note: We may in the future try to use something like https://github.com/kubernetes/kube-openapi for the schema generation but for now to keep things more aligned with our current schemas  https://github.com/fybrik/json-schema-generator will be used.

simanadler:
Is there clear documentation in the website about this - both in terms of they why and the how?

roee88:
There is minimal documentation in https://fybrik.io/v0.5/tasks/custom-taxonomy/.
There are related open issues in https://github.com/fybrik/fybrik/issues/792 and https://github.com/fybrik/fybrik/issues/779.
There is an `Update docs` task in this issue. 

Please be clearer on the ask.

simanadler:
@roee88 I suggest we add a section in https://fybrik.io/v0.5/concepts/introduction/ about taxonomy - i.e. why it's needed, and how it is used.

roee88:
Documentation will be addressed as part of #779

Issue #886, opened 2021-10-21
Remove a duplicate structure for storing asset metadata


shlomitk1:
The current code base has the following structure for storing asset metadata:
```
type DataDetails struct {
	// Name of the asset
	Name string
	// Interface is the protocol and format
	Interface app.InterfaceDetails
	// Geography is the geo-location of the asset
	Geography string
	// Connection is the connection details in raw format as received from the connector
	Connection serde.Arbitrary
	// Metadata
	TagMetadata *pb.DatasetMetadata
}
```
It is obtained by converting the catalog connector response. 
**Goal:** to use the same structure both in manager and catalog connector without doing conversions.
**Dependency**: open-api model for catalog connector (similar to policy manager). 
@roee88 @rohithdv Do we have an issue for this?

shlomitk1:
Depends on https://github.com/fybrik/fybrik/issues/972

PR #1054, opened 2021-12-23
Use the revisited taxonomy model files 

roee88:
See #1049

#1051 introduced model files in `pkg/models`. This PR changes code in pkg and manager to use the new model files. Future PRs can improve upon that (e.g., to address #886. I put some TODOs for obvious things I encountered but wanted to keep this PR sane).

Minor changes are also made to the OPA connector so it would compile but a future PR will revisit the OPA connector implementation to remove any dependency on older model files or grpc definitions.


config/taxonomy/datacatalog.spec.yaml
```diff
@@ -12,7 +12,7 @@ paths:
         summary: getAssetInfo
         parameters:
           - in: header
-            name: X-Request-DataCatalog-Cred
+            name: X-Request-Datacatalog-Cred
```
roee88, line 15: FYI this is because header names should be in canonical form

shlomitk1 (approved):

Issue #926, opened 2021-11-01
Manager: Generate log information in a standardized manner


simanadler:
### Discussed in https://github.com/fybrik/fybrik/discussions/890

Create a single log structure to be used by both the control plane and the data plane components as is described in the [visibility high level design document](https://docs.google.com/document/d/1x3Jnp3CIN6YhJKtWM1uIimaw9-q6Hb9M7LqkazmBRBo/edit?usp=sharing).

Steps:

- [x] Define the log entry structure and make it available to all components 
- [x] Update the contributing documenation to include a section about logging
- [x] Change the fybrik manager to generate log entries according to the new structure
- [x] Change the fybrik manager to log all changes to fybrikapplication and fybrikapplication status
- [x] Change the fybrik manager to log all config maps used when it is deployed - i.e. log config map read when manager starts up

Connectors and data plane component updates are addressed in separate issues. #982 and #983

@laurieaw63 @ronenkat @roee88 @revit13 @cdoron @shlomitk1 @rohithdv @ritwikc Please comment on both the newly defined structure and the task list.

roee88:
> (taxonomy?)

I think that it would be more appropriate to start with a code style section in the developer's guide and add logging guidelines to it (alongside other guidelines in the future).  For go code we can add something to pkg/logging but for 3rd parties we got the guidelines covering it. Otherwise, the complexity and overhead of logging might be too high.

simanadler:
> > (taxonomy?)
> 
> I think that it would be more appropriate to start with a code style section in the developer's guide and add logging guidelines to it (alongside other guidelines in the future). For go code we can add something to pkg/logging but for 3rd parties we got the guidelines covering it. Otherwise, the complexity and overhead of logging might be too high. Is there anything in opentelemetry that can help here in the future?

OK.  Makes sense.  Updated the tasks in the issue.

shlomitk1:
@simanadler We don't have credential manager connectors. Please remove from the task list.

ronenkat:
> @simanadler We don't have credential manager connectors. Please remove from the task list.

We have Vault back-end plug-ins. They should provide logging as well.

roee88:
> > @simanadler We don't have credential manager connectors. Please remove from the task list.
> 
> We have Vault back-end plug-ins. They should provide logging as well.

I think that this might be out of scope / different issue. It is inherently different than anything else in this list (does have the same context) and vault probably has its own logging for tracking who accessed what secret.

[...]

PR #1058, opened 2021-12-23
Follow-up logging tasks

shlomitk1:

More updates to #926:

    - Changed label to app-uuid instead of FybrikApplicationUUID.
    - Added uuid to unit tests for easier debugging
    - Added uuid to adminconfig policy evaluator input
    - Replaced logging methods in pkg/adminconfig


roee88 (approved):
Consider keeping the logger as a field if it's always the same logger.

PR #1060, opened 2021-12-26
 Use the taxonomy models to generate JSON schemas

roee88:
Part of #1049

1. Removes config/taxonomy. Moving the example taxonomy to samples/taxonomy.
2. From the `pkg/model` code generate JSON schemas for base taxonomy.json, catalog definitions, policy manager definitions 
3. Adds a `generate` make target to the top level Makefile. It overlaps with the functionality of `generate` in manager which will be removed in a future PR. 
4. To align with the example taxonomy and the fact that `dataformat` is an optional field this field was removed from any resource that used it with a placeholder format value ("table", "arrow").

The notebook test will pass once https://github.com/fybrik/arrow-flight-module/pull/109 is merged. Requires both PRs to be approved first.


rohithdv (approved):

charts/fybrik/files/taxonomy/policymanager.json
```diff
@@ -0,0 +1,81 @@
...
+      ],
+      "properties": {
+        "actionType": {
+          "$ref": "#/definitions/RequestActionType"
+        },
+        "destination": {
+          "type": "string"
+        },
+        "processingLocation": {
+          "$ref": "taxonomy.json#/definitions/ProcessingLocation"
```
shlomitk1, line 52: Why aren't both destination and processingLocation of the same type?

charts/fybrik/files/taxonomy/datacatalog.json
```diff
@@ -0,0 +1,113 @@
...
+        "resourceMetadata": {
+          "$ref": "#/definitions/ResourceMetadata"
+        }
+      }
+    },
+    "OperationType": {
+      "description": "OperationType Type of operation requested for the asset",
+      "type": "string",
+      "enum": [
+        "read"
```
shlomitk1, line 43: Perhaps to add also write and delete for future uses?

shlomitk1 (approved):

charts/fybrik/files/taxonomy/datacatalog.json
roee88: Keeping it as is for now because that's the only value in the existing taxonomy, but before closing the issue a full final review of the taxonomy definitions is required.

PR #1065, opened 2021-12-28
Taxonomy validate command changes

roee88:
Part of https://github.com/fybrik/fybrik/issues/1049

- Removed dependencies openapi2crd  and kin-openapi
- Fixed appinfo schema (was not structural)
- Error handling fixes in taxonomy validation


pkg/taxonomy/model/additionalprops.go
```diff
@@ -41,3 +45,22 @@ func (a AdditionalPropertiesType) MarshalJSON() ([]byte, error) {
...
+
+	if a.Schema != nil {
+		return &apiextensions.JSONSchemaPropsOrBool{Allows: true, Schema: a.Schema.ToJSONSchemaProps(flattenBy)}
+	}
+
+	// additionalProperties: false is forbidden in Kubernetes
+	// x-kubernetes-preserve-unknown-fields replaces the functionality of additionalProperties: true
+	if a.Allowed != nil {
+		// return &apiextensions.JSONSchemaPropsOrBool{Allows: *a.Allowed}
+		return nil
```
shlomitk1, line 62: Shouldn't it be `return &apiextensions.JSONSchemaPropsOrBool{Allows: true}` ?

roee88: For validation we remove additionalProperties unless it's set to a concrete schema (because of the reasons in the comment above this if condition)

shlomitk1 (approved):

PR #1072, opened 2021-12-29
doc: fix copy command for running manager locally

ityuhui:
The command fails to execute because the directory "/tmp/adminconfig" does not exist. 
```shell
cp -R ../charts/fybrik/files/adminconfig/* /tmp/adminconfig/
```

```shell
cp: target '/tmp/adminconfig/' is not a directory
```

So I try to change this command to
```shell
cp -R ../charts/fybrik/files/adminconfig /tmp/
```

### DIFF
--- a/manager/README.md
+++ b/manager/README.md
@@ -64,8 +64,8 @@ set -a; . .env; set +a
 
 ### Copy taxonomy JSON files and config policies locally
 ```bash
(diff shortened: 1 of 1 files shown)


roytman:
Hi @ityuhui , thanks for the fix.

roytman (approved):

PR #1078, opened 2021-12-29
Katalog connector openapi

tomersolomon1:
Part of https://github.com/fybrik/fybrik/issues/1049

Using the new model structures in Katalog


charts/fybrik/values.yaml
```diff
@@ -314,7 +314,7 @@ katalogConnector:
 
   service:
     type: ClusterIP
-    port: 80
+    port: 8080
```
shlomitk1, line 317: I think the deployment will look for katalog-connector:80.
` CATALOG_CONNECTOR_URL: {{ .Values.coordinator.catalogConnectorURL | default (printf "%s-connector:80" .Values.coordinator.catalog) | quote }}`
You need to set coordinator.catalogConnectorURL to start with "http://" and point to the right port for notebook tests and the actual deployment.

pkg/logging/logging.go
```diff
@@ -47,7 +47,6 @@ const (
 )
 
 // Log Entry Params - those listed in the constants below and
-// FybrikApplicationUUID defined in fybrik.io/manager/utils/utils.go
```
shlomitk1, line 50: Why commented out?

connectors/katalog/example/asset.yaml
```diff
@@ -2,28 +2,24 @@ apiVersion: katalog.fybrik.io/v1alpha1
 kind: Asset
```
shlomitk1, line 2: BTW, there is another asset definition in https://github.com/fybrik/fybrik/blob/master/manager/testdata/notebook/example-asset.yaml which should be updated as well.

shlomitk1 (approved):

PR #1081, opened 2021-12-30
Fix to local multi-cluster setup

pugangxa:
1. Fix the make cluster-prepare-wait since there's no vault to wait
2. Fix the url for razee to out of box usage
3. Add the http prefix since this setup uses mock implementation

Signed-off-by: Gang Pu <pugangxa@cn.ibm.com>

### DIFF
--- a/charts/fybrik/kind-control.values.yaml
+++ b/charts/fybrik/kind-control.values.yaml
@@ -26,6 +26,7 @@ cluster:
 
 # Configuration when deploying to a coordinator cluster.
 coordinator:
+  catalogConnectorURL: "http://katalog-connector:80"
   vault:
     # vault service in local setup is exposed via nodeport
     address: http://control-control-plane:31752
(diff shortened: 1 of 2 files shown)


charts/fybrik/kind-control.values.yaml
```diff
@@ -26,14 +26,15 @@ cluster:
 
 # Configuration when deploying to a coordinator cluster.
 coordinator:
+  catalogConnectorURL: "http://katalog-connector:80"
```
revit13, line 29: why does the default `catalogConnectorURL` value need to be changed?

shlomitk1: This is not a mock setup. `make deploy` is used. Value of catalogConnectorURL should not change, it is defined in charts/fybrik/values.yaml. Since katalog-connector is not open-api based yet, adding "http://" won't work.

pugangxa: Refer to https://github.com/fybrik/fybrik/blob/ce6d8e7cdf27b20d7b683126aeea3af583c2dc97/charts/fybrik/kind-control.values.yaml#L54-L56 , in this setup mock is used in the same file and the mock does not have grpc endpoint, we should make them consistent.

pugangxa: If not add the "http://" prefix the manager will keep restarting.

shlomitk1: Ok, you are right, I missed the service image generation.

shlomitk1 (approved):

PR #1082, opened 2021-12-30
Move makefile targets from manager to root

roee88:
* Moved manifests target
* Moved pre-test target
* Removed redundant generate target
* Removed unused run-deploy-tests target

Side effect of #1049

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ include Makefile.env
 export DOCKER_TAGNAME ?= master
 export KUBE_NAMESPACE ?= fybrik-system
 
+.PHONY: all
 all: generate manifests verify
 
 .PHONY: license
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@ include Makefile.env
 export DOCKER_TAGNAME ?= master
 export KUBE_NAMESPACE ?= fybrik-system
 
+all: generate manifests verify
+
 .PHONY: license
 license: $(TOOLBIN)/license_finder
 	$(call license_go,.)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -25,13 +38,25 @@ deploy: $(TOOLBIN)/kubectl $(TOOLBIN)/helm
...
+		$(shell find samples/taxonomy/example -type f -name '*.yaml')
+	cp manager/testdata/unittests/sampletaxonomy/taxonomy.json /tmp/taxonomy/taxonomy.json
+
 .PHONY: test
 test: export MODULES_NAMESPACE?=fybrik-blueprints
 test: export CONTROLLER_NAMESPACE?=fybrik-system
-test:
-	$(MAKE) -C manager pre-test
+test: pre-test
 	go test -v ./...
```
shlomitk1, line 59: In manager test there were some environment variables defined (e.g. modules namespace).

manager/Makefile
```diff
@@ -1,72 +1,31 @@
...
-#include Makefile.env
 
 .PHONY: all
 all: manager
 
 include $(ROOT_DIR)/hack/make-rules/tools.mk
+include $(ROOT_DIR)/hack/make-rules/docker.mk
+include $(ROOT_DIR)/hack/make-rules/verify.mk
+
 DOCKER_NAME ?= manager
```
shlomitk1, line 11: Don't the previous includes define already DOCKER_NAME?

roee88: This makes `IMG` be defined correctly. You commented before my last commit so maybe it's clearer now.

shlomitk1 (approved):

PR #1086, opened 2021-12-30
Add golint workflow

roytman:
- Add a golint workflow. Currently, it checks only new code, eventually, we will validate all code. 
- In addition, this PR adds an option to configure goimpotrts with the separation of local imports.  in order to do it run `hack/tools/bin/fixgoimports.sh <fileName>. if filename is not provided, imports in te entire project will be corrected.
- Finally, it have an option to automatically verify code before commit. In order to do it, install `pre-commit` utility:
```
pip install pre-commit
pre-commit install
```


.github/workflows/build.yml
```diff
@@ -24,6 +24,8 @@ jobs:
       with:
         go-version: ${{ env.GO_VERSION }}
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
```
roee88, line 28: Why is this change required?

.github/workflows/golang-lint.yml
```diff
@@ -0,0 +1,24 @@
...
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: golangci-lint
+        uses: reviewdog/action-golangci-lint@v2
+        with:
+          reviewdog_version: 'v0.13.1'
+          golangci_lint_version: 'v1.43.0'
+          golangci_lint_flags:  '--new-from-rev=HEAD~'
+          level: warning # GitHub Status Check won't become failure with this level.
```
roee88, line 24: I have been using the official action in other repos. It has good integration with GitHub. 
https://github.com/golangci/golangci-lint-action

.golangci.yml
```diff
@@ -0,0 +1,148 @@
+linters-settings:
```
roee88, line 1: What is the original source of this configuration file? Asking because there are some stuff I don't understand (like why it has gci configuration)

.pre-commit-config.yaml
```diff
@@ -0,0 +1,12 @@
+repos:
```
roee88, line 1: Need to add to development docs how to set it up. Also, for those of us who aren't fans of pre-commit hooks, I recommend that we make sure that the top IDEs can be setup to do all of these checks automatically and provide instructions on how to configure it so (not in this PR though)

hack/make-rules/tools.mk
```diff
@@ -86,6 +86,13 @@ $(TOOLBIN)/misspell:
 	GOBIN=$(ABSTOOLBIN) go install github.com/client9/misspell/cmd/misspell@v0.3.4
 	$(call post-install-check)
 
+INSTALL_TOOLS += $(TOOLBIN)/goimports
+$(TOOLBIN)/goimports:
+	GOBIN=$(ABSTOOLBIN) go install golang.org/x/tools/cmd/goimports@v0.1.8
+	$(call post-install-check)
+	echo $(ABSTOOLBIN)/goimports -local fybrik.io/fybrik -w '$$'1 > $(ABSTOOLBIN)/fixgoimports.sh
+	chmod 755 $(ABSTOOLBIN)/fixgoimports.sh
+
```
roee88, line 95: I personally never liked the fact that tools.ml installs `dlv`. Now `goimports` falls into the same category in my opinion. The original idea was to have tools installed just for the project so they won't compete with a global install of the user. But in my opinion that doesn't hold for tools that I already use in the project because my IDE does (like `dlv` and `goimports`). Just a personal opinion though.

hack/make-rules/verify.mk
```diff
@@ -38,43 +38,25 @@ CODE_MAINT += tidy
...
-.PHONY: lint-todo
-lint-todo: $(TOOLBIN)/golangci-lint
-	$(TOOLBIN)/golangci-lint run --enable=godox ${GOLINT_LINTERS} ./...
+.PHONY: lint-updated
+lint-updated: $(TOOLBIN)/golangci-lint
+	$(TOOLBIN)/golangci-lint run --new-from-rev=HEAD~
+
+.PHONY: lint-godox
+lint-godox: $(TOOLBIN)/golangci-lint
+	$(TOOLBIN)/golangci-lint run --enable=godox
```
roee88, line 59: Why is godox disabled in main config? Why does it need its own target? Also, godox is opiniated and we do use TODO comments in out code, like many other projects. You can change that of course :-)

roee88 (commented):
1. It feels to me like there are too many lint related make targets and that can confuse people
2. revive can probably be removed from tools now

hack/make-rules/tools.mk
roytman: indeed, but I wanted to allow to configure/fix imports separately. 
Plan to remove it later.

hack/make-rules/verify.mk
roytman: it was one of my questions to you ;-)

.github/workflows/golang-lint.yml
roee88: > but I want to add the reviewdog GitHub commiter

May I ask why?

roytman: sure, it simplifies to see why lint fails, but as I said it is a WIP, will see

roee88: What I meant is that the official action also shows this.

roytman:
was done by #1340 and several other golint related PRs. Closing.

PR #1099, opened 2022-01-03
Add .PHONY: to make rules

roytman:
Add  .PHONY: to make rules

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,7 @@ deploy: $(TOOLBIN)/kubectl $(TOOLBIN)/helm
 	$(TOOLBIN)/helm install fybrik charts/fybrik --values $(VALUES_FILE) \
                --namespace $(KUBE_NAMESPACE) --wait --timeout 120s
 
+.PHONY: pre-test
 pre-test: generate manifests $(TOOLBIN)/etcd $(TOOLBIN)/kube-apiserver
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -40,6 +40,7 @@ deploy: $(TOOLBIN)/kubectl $(TOOLBIN)/helm
 	$(TOOLBIN)/helm install fybrik charts/fybrik --values $(VALUES_FILE) \
                --namespace $(KUBE_NAMESPACE) --wait --timeout 120s
 
+.PHONY: pre-test
```
shlomitk1, line 43: According to https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html, 
`a phony target should not be a prerequisite of a real target file`.
`pre-test` is a prerequisite of `test`

shlomitk1:
Is there an issue this PR is related to? Why is it required?

roytman:
> 



> Is there an issue this PR is related to? Why is it required?

No, I did not open an issue, but in Makefiles all targets that are not connected to real files need the `.PHONY` annotation. https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

Makefile
shlomitk1: @roytman ^^^

shlomitk1 (approved):

PR #1106, opened 2022-01-04
Change default appVersion to 0.0.0 for fybrik and fybrik-crd charts

revit13:


revit13:
Closing this as running [quick-start ](https://fybrik.io/dev/get-started/quickstart/) with `Install latest development version from GitHub` makes the control-plane pods fail due to Imagepull error.
The reason is that the Chart.yaml of fybrik is `0.0.0` which is also the image tag that does not exist in the public repo.

roytman (approved):

PR #1118, opened 2022-01-05
Correct the namespace of the deployed modules when deleting a fybrik application

ityuhui:
When I delete a fybrik application

```
kubectl delete fybrikapplications.app.fybrik.io my-notebook
```

The plotter and blueprint can be deleted successfully.
but the deployed module helm release can not be deleted.

```
fybrik-blueprints        my-notebook-fybrik-notebook-sample-arrow-flight-aef23-arrobszj6   1/1     Running     0          16h
```

The log of manager shows the "helm release" cannot be found
```
{"level":"error","Controller":"Blueprint","error":"uninstall: Release not loaded: my-notebook-fybrik-notebook-sample-arrow-flight-aef23: release: not found","time":"2022-01-05T01:35:26Z","caller":"/root/fybrik/manager/controllers/app/blueprint_controller.go:105","message":"Error while deleting owned resources"}
```

I think the root cause is the namespaces of module do not match:

`blueprint.Namespace` in `deleteExternalResources` is "fybrik-system" 
but actually the namespace of deployed modules is "fybrik-blueprints"

The right namespace should be `blueprint.Spec.ModulesNamespace`

### DIFF
--- a/manager/controllers/app/blueprint_controller.go
+++ b/manager/controllers/app/blueprint_controller.go
@@ -126,7 +126,7 @@ func (r *BlueprintReconciler) reconcileFinalizers(blueprint *fapp.Blueprint) (ct
 func (r *BlueprintReconciler) deleteExternalResources(blueprint *fapp.Blueprint) error {
 	errs := make([]string, 0)
 	for release := range blueprint.Status.Releases {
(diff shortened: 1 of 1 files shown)


shlomitk1 (approved):
Good catch!

PR #1124, opened 2022-01-05
Adding taxonomy validation in manager

### LINKED ISSUE #1090: Validate connectors requests/responses against taxonomy
Validate all requests/responses to connectors. Can probably be part of the client code in manager.

rohithdv:
This PR fixes #1090 


shlomitk1:
1. Use constants for taxonomy file locations (see https://github.com/fybrik/fybrik/blob/41e12b47096558033026102a5914efb99e1fb0a8/manager/controllers/app/fybrikmodule_controller.go#L36)
2. I don't think we need to validate requests, only responses from the connectors.

rohithdv:
> 1. Use constants for taxonomy file locations (see https://github.com/fybrik/fybrik/blob/41e12b47096558033026102a5914efb99e1fb0a8/manager/controllers/app/fybrikmodule_controller.go#L36
>    )
> 2. I don't think we need to validate requests, only responses from the connectors.

fixed. Please check. Thanks.

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -412,6 +418,32 @@ func CreateDataRequest(application *api.FybrikApplication, dataCtx api.DataConte
...
+	}
+
+	// Return any error
+	if len(allErrs) == 0 {
+		return nil
+	}
+
+	return apierrors.NewInvalid(
+		schema.GroupKind{Group: "app.fybrik.io", Kind: "DataCatalog-AssetResponse"},
+		response.ResourceMetadata.Name, allErrs)
```
shlomitk1, line 444: ResourceMetadata.Name is optional, better use dataset id

manager/controllers/app/policy_compiler_interface.go
```diff
@@ -30,10 +40,44 @@ func ConstructOpenAPIReq(datasetID string, input *app.FybrikApplication, operati
...
+	log.Println(allErrs)
+	if err != nil {
+		log.Println("err: ")
+		log.Println(err)
+		return err
+	}
+
+	// Return any error
+	if len(allErrs) == 0 {
+		log.Println("returning nil")
```
shlomitk1, line 72: remove this printout

manager/controllers/app/policy_compiler_interface.go
```diff
@@ -30,10 +40,44 @@ func ConstructOpenAPIReq(datasetID string, input *app.FybrikApplication, operati
...
+		return err
+	}
+	log.Println("responseJSON (policy decisions):" + string(responseJSON))
+
+	// Validate Fybrik module against taxonomy
+	allErrs, err = validate.TaxonomyCheck(responseJSON, taxonomyFile)
+	log.Println("allErrs:")
+	log.Println(allErrs)
+	if err != nil {
+		log.Println("err: ")
```
shlomitk1, line 65: Use zerolog.Logger instead of log as in other manager files.

manager/controllers/mockup/policymanager.go
```diff
@@ -42,28 +42,69 @@ func (m *MockPolicyManager) GetPoliciesDecisions(input *policymanager.GetPolicyD
...
+
+		actionBytes, errJSON := json.MarshalIndent(action, "", "\t")
+		if errJSON != nil {
+			return nil, fmt.Errorf("error Marshalling External Catalog Connector Response: %v", errJSON)
+		}
+		err := json.Unmarshal(actionBytes, &actionOnDataset)
+		if err != nil {
+			return nil, fmt.Errorf("error in unmarshalling actionBytes : %v", err)
+		}
+
```
shlomitk1, line 58: 1. Does taxonomy require additional properties for Deny action?
2. This code is duplicated in two more places.

manager/controllers/app/fybrikapplication_controller.go
rohithdv: Actually dataset id is stored as response.ResourceMetadata.Name in the response object. There is no other field in the response object which explicitly stores the dataset id. Hope this is ok ?

manager/controllers/mockup/policymanager.go
rohithdv: 1. As of now, no. Please check https://github.com/fybrik/fybrik/blob/67d8886ea0bf58e80c2b1a6999f53c20d6bf2998/samples/taxonomy/example/module/actions.yaml#L27
2. Fixed code duplication. Please check if ok. Thanks.

manager/controllers/app/policy_compiler_interface.go
rohithdv: fixed. Please check. Thanks.

manager/controllers/app/policy_compiler_interface.go
rohithdv: fixed. Please check. Thanks.

manager/controllers/mockup/policymanager.go
```diff
@@ -40,39 +53,53 @@ func (m *MockPolicyManager) GetPoliciesDecisions(input *policymanager.GetPolicyD
...
 		// empty result simulates allow
 		// no need to construct any result item
+
 	case "deny-dataset":
 		actionOnDataset := taxonomy.Action{}
-		(&actionOnDataset).Name = "Deny"
+		action := make(map[string]interface{})
+		action["name"] = "Deny"
+		denyAction := map[string]interface{}{}
+		action["Deny"] = denyAction
```
shlomitk1, line 62: Is this required? Isn't the name for Deny action enough?

rohithdv: Since we have an explicit action named Deny (like for REdactAction, RemoveAction) as given [here](https://github.com/fybrik/fybrik/blob/67d8886ea0bf58e80c2b1a6999f53c20d6bf2998/samples/taxonomy/example/module/actions.yaml#L25), we need an explicit Deny action related field in the returned action. @roee88

manager/controllers/app/fybrikapplication_controller.go
shlomitk1: You could pass req.Context.DataSetID to ValidateAssetResponse and use that.
ResourceMetadata.Name is optional, thus we don't want to rely on it.

rohithdv: sure will do.

shlomitk1 (approved):
We'll need to change logging methods in some of the files. Leaving for a future PR.

PR #1136, opened 2022-01-06
Fix the image used by kind develop env

pugangxa:
Signed-off-by: Gang Pu <pugangxa@cn.ibm.com>

Image tages already changed so need change value for kind cluster. In my development it always pulls the image with "master" tag now.

### DIFF
--- a/charts/fybrik/kind-control.values.yaml
+++ b/charts/fybrik/kind-control.values.yaml
@@ -5,7 +5,7 @@
 # Global configuration applies to multiple components installed by this chart
 global:
   hub: localhost:5000/fybrik-system
-  tag: "master"
+  tag: "0.0.0"
   imagePullPolicy: "Always"
 
   prettyLogging: false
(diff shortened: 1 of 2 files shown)


revit13:
@pugangxa for local development the docker tag is `0.0.0` please see https://github.com/fybrik/fybrik/blob/master/charts/fybrik/integration-tests.values.yaml#L8. The images are pushed to public repo with tag `master`

pugangxa:
> @pugangxa for local development the docker tag is `0.0.0` please see https://github.com/fybrik/fybrik/blob/master/charts/fybrik/integration-tests.values.yaml#L8. The images are pushed to public repo with tag `master`

This is for the kind cluster and you can see https://github.com/fybrik/fybrik/blob/edc962c25e09f699d2b5bcfbfdd4b0d34acc828c/charts/fybrik/kind-control.values.yaml#L7-L9 it's in fact pushed to `localhost:5000/fybrik-system`. 
In my development everytime I build and pushed to my kind registry the image is with tag `0.0.0` but it always pulled the image w/ tag `master` so cause problem. I just changed the values for kind cluster for local development to make them consistent. For public repo user can still use `master` as `tag`.

revit13:
@pugangxa you are right I was confused. Thanks!

revit13 (approved):

PR #1143, opened 2022-01-07
Completing Operate First GitHub workflow

### LINKED ISSUE #523: Send a PR to operate-first/apps on each release automatically

eletonia:

- removes generated files from `integration/operate-first`
- changes README to point to `integration/operate-first` instead of `samples/operate-first`
- pushes updates to generated files in Fybrik repo to fork of `operate-first/apps` repo
- creates a pull request to the original `operate-first/apps` repo
- specifies which directory in apps repo to push generated files to (`apps/cluster-scope/base`)
- specifies which repo to make the pull request from (`fybrik/operate-first-apps`)
- tested GitHub workflow to see if pull request is being created as expected - see workflow run [here](https://github.com/eletonia/fybrik/runs/4926705037?check_suite_focus=true) and PR created [here](https://github.com/operate-first/apps/pull/1553)

Fixes #523


.github/workflows/operate-first-publish.yaml
```diff
@@ -8,42 +8,43 @@ on:
...
+          rsync -a --include='*/' --include='*.yaml' --exclude='*' fybrik-repo/integration/operate-first/ operate-first-repo/cluster-scope/base
       - uses: tibdex/github-app-token@v1
         id: generate-token
         with:
-          app_id: ${{ secrets.CHARTS_APP_ID }}
-          private_key: ${{ secrets.CHARTS_APP_PRIVATE_KEY  }}
-          repository: fybrik/fybrik
+          app_id: ${{ secrets.FYBRIK_BOT_APP_ID }}
+          private_key: ${{ secrets.FYBRIK_BOT_PRIVATE_KEY }}
+          repository: fybrik/charts
```
revit13, line 39: I think the repo needs to be `fybrik/operate-first-apps`

revit13:
@eletonia a general question - why do the generated files need to be part of the directory in fybrik? why they can not be generated on the fly upon executing the workflow and creating the pr?

.github/workflows/operate-first-publish.yaml
```diff
@@ -8,42 +8,43 @@ on:
...
           path: fybrik-repo
+      - name: Checkout original Operate First repository
+        uses: actions/checkout@v2
+        with:
+          repository: operate-first/apps
+          path: operate-first-repo
       - run: |
-          rm -r fybrik-repo/samples/operate-first/*/
-          rm fybrik-repo/samples/operate-first/*.yaml
+          rm -r fybrik-repo/integration/operate-first/*/
```
roee88, line 25: I don't think that we need the generated files to be committed at all (and then you won't need to delete them here)

.github/workflows/operate-first-publish.yaml
```diff
@@ -8,42 +8,43 @@ on:
...
           pushd fybrik-repo/hack/tools && ./install_yq.sh && ./install_helm.sh && popd
-          pushd fybrik-repo/samples/operate-first
+          pushd fybrik-repo/integration/operate-first
           make helm-template
           pip install pipenv
           pipenv install
           pipenv run python filterClusterScoped.py
+          rm ./*.yaml && popd
```
roee88, line 32:
```suggestion
          rm ./*.yaml
          popd
```

eletonia:
> @eletonia a general question - why do the generated files need to be part of the directory in fybrik? why they can not be generated on the fly upon executing the workflow and creating the pr?

@revit13 You're right, we don't need these anymore. I was committing them earlier to keep track of the latest version of the CRDs that were deployed to Operate First by manually making a PR. Now that this is automated, I removed the generated files.

.github/workflows/operate-first-publish.yaml
eletonia: @revit13 Thanks, just changed it.

.github/workflows/operate-first-publish.yaml
eletonia: @roee88 I removed the generated files and this line

.github/workflows/operate-first-publish.yaml
eletonia: Done

revit13 (approved):

roytman (approved):

PR #1147, opened 2022-01-10
doc: add link to the doc of DDC

ityuhui:
this adds link of "Architecture" to the doc "Data Distribution Controller"


roee88 (approved):

roee88:
Thanks. @roytman the docs about Data Distribution Controller should move to its repo, I think that they're a bit outdated too so we should consider creating a slim version in the README instead.

PR #1174, opened 2022-01-14
Update readme for running manager locally

ityuhui:
This PR adds environment variables about Razee to the readme.

### DIFF
--- a/manager/README.md
+++ b/manager/README.md
@@ -55,6 +55,14 @@ VAULT_MODULES_ROLE="module"
 ENABLE_WEBHOOKS="false"
 ```
 
+If the manager runs in the Razee environment, you also need to add the following environment variables:
+
+```bash
+RAZEE_USER="razee-dev@example.com"
+RAZEE_PASSWORD="password123"
+RAZEE_URL="http://localhost:3333/graphql"
+```
+
 If you plan to run manager from the command line,
 then run the following to export all of the variables:
 
--- a/manager/README.md
+++ b/manager/README.md
@@ -55,21 +55,15 @@ VAULT_MODULES_ROLE="module"
 ENABLE_WEBHOOKS="false"
 ```
 
-If the manager runs in a Razee environment, you also need to add the following environment variables:
+If the manager works with a Razee service, you also need to add the following environment variables:
 
 ```bash
-RAZEE_USER="razee-dev@example.com"
-RAZEE_PASSWORD="password123"
-RAZEE_URL="http://localhost:3333/graphql"
+RAZEE_URL=<Razee access point> # e.g. "http://localhost:3333/graphql"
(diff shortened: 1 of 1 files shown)


manager/README.md
```diff
@@ -55,6 +55,14 @@ VAULT_MODULES_ROLE="module"
 ENABLE_WEBHOOKS="false"
 ```
 
+If the manager runs in the Razee environment, you also need to add the following environment variables:
```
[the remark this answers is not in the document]
ityuhui: When a Razee environment is setup:

1. There are "razeedash" and "razeedash-api" pods running at fybrik control plane:
`kubectl get po -A | grep razeedash`

2. Can get response from NodePort service of "razeedash-api-np"
`curl http://localhost:3333/graphql`

manager/README.md
```diff
@@ -55,6 +55,14 @@ VAULT_MODULES_ROLE="module"
...
+
+```bash
+RAZEE_USER="razee-dev@example.com"
+RAZEE_PASSWORD="password123"
+RAZEE_URL="http://localhost:3333/graphql"
```
shlomitk1, line 63: Where is the port-forward from razee pod to localhost:3333?

ityuhui: Don't need a "port-forward".
There is a NodePort service "razeedash-api-np" created by https://github.com/fybrik/fybrik/blob/ee85999dd766b97feb0ab2f7e095247deccb580f/third_party/razee/razee-nodeports.yaml#L5

shlomitk1: OK, thanks!

manager/README.md
```diff
@@ -55,6 +55,22 @@ VAULT_MODULES_ROLE="module"
 ENABLE_WEBHOOKS="false"
 ```
 
+If the manager runs in a Razee environment, you also need to add the following environment variables:
```
roytman, line 58: Hi thanks for the PR, 
but it is not clear what do you mean by "a Razee environment", I assume that you mean Razee and not Satellite deployment. However, the Raze can be collocated on the same cluster as the manager, or can be a remote service. Manager can connect to Razee by using userName/password  as you explain here or by using oauth API key, as shown [here](https://fybrik.io/v0.6/tasks/multicluster/#coordinator-cluster).

When a user (developer)  runs  manager manually, he probably knows knows which Razee he plans to use. 
I would suggest the following changes:
If the manager will works with a Razee service,  you also need to add the following environment variables:
```bash
RAZEE_URL=<Razee access point> # e.g. "http://localhost:3333/graphql"
RAZEE_USER=<Razee user>
# you should define either RAZEE_PASSWORD or API_KEY, if both variables are defined RAZEE_PASSWORD will be used
RAZEE_PASSWORD=<Razee password> 
API_KEY=<Razee api key>
```
And the paragraph `You can know that the manager runs in a Razee environment by` (Lines 66- 73) can be removed.

ityuhui: Thanks! I will update it with your suggestions.

roytman (approved):

PR #1176, opened 2022-01-16
Fix Vault's “claim iss is invalid” error in Kubernetes 1.21. 

### LINKED ISSUE #1175: Vault error  [‘claim “iss” is invalid’] with Kubernertes 1.21
Vault error  [‘claim “iss” is invalid’] appears when running the notebook sample with Kubernetes 1.21

revit13:
This PR adds `disable_iss_validation=true` when configuring Kubernetes auth method in Vault. Following what is written in [here](https://www.vaultproject.io/docs/auth/kubernetes#discovering-the-service-account-issuer): `disable_iss_validation=true is the new recommended value for all versions of Vault`. and [here](https://www.vaultproject.io/docs/auth/kubernetes#kubernetes-auth-method): `If you are upgrading to Kubernetes v1.21+, ensure the config option disable_iss_validation is set to true. `

Closes #1175

### DIFF
--- a/charts/vault/templates/post-pod-creation-cm.yaml
+++ b/charts/vault/templates/post-pod-creation-cm.yaml
@@ -10,7 +10,7 @@ data:
 {{- if semverCompare "<1.21-0" .Capabilities.KubeVersion.Version }}
       vault write auth/kubernetes/config kubernetes_host="https://kubernetes.default.svc:443"
 {{- else }}
-      vault write auth/kubernetes/config issuer="https://kubernetes.default.svc" kubernetes_host="https://kubernetes.default.svc:443"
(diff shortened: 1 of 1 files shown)


charts/vault/templates/post-pod-creation-cm.yaml
```diff
@@ -8,9 +8,9 @@ data:
       # configuring kubernetes auth method and use it later to create roles.
       vault auth enable kubernetes
 {{- if semverCompare "<1.21-0" .Capabilities.KubeVersion.Version }}
-      vault write auth/kubernetes/config kubernetes_host="https://kubernetes.default.svc:443"
+      vault write auth/kubernetes/config kubernetes_host="https://kubernetes.default.svc:443" disable_iss_validation=true
```
roytman, line 11: why do you set it for Kubernetes versions less than 1.21? The Valut's recommendation is for "Kubernetes v1.21+"

roytman (approved):

PR #1177, opened 2022-01-16
Track changes in policy files and recompile when needed

shlomitk1:
Fixes https://github.com/fybrik/fybrik/issues/1010

- Only out of the box policies are deployed by default
- Instructions how to add/delete/modify policies are provided
- Policy files are recompiled if newer than the last compile time
- Added policies' version


pkg/adminconfig/rego_file_evaluator.go
```diff
@@ -78,7 +100,22 @@ func (r *RegoPolicyEvaluator) prepareQuery() (rego.PreparedEvalQuery, error) {
...
 // Evaluate method evaluates the rego files based on the dynamic input object
 func (r *RegoPolicyEvaluator) Evaluate(in *EvaluatorInput) (EvaluatorOutput, error) {
+	log := r.Log.With().Str(utils.FybrikAppUUID, in.Workload.UUID).Logger()
+	recompile := false
 	if !r.ReadyForEval {
+		recompile = true
+	} else {
+		recent, err := r.isRecent()
+		if err != nil {
+			log.Error().Err(err).Msg("")
```
simanadler, line 110: Need a meaningful message

site/docs/concepts/config-policies.md
```diff
@@ -135,3 +136,18 @@ config[{"copy": decision}] {
 }
 
 ```
+
+### How to add / modify / delete policies
```
simanadler, line 140: If we are doing lifecycle management of policies now, we need to add versions to policies.

simanadler (commented):
I think we need to add versions to config policies if we are doing lifecycle management of them now.

pkg/adminconfig/rego_file_evaluator.go
shlomitk1: Done.

site/docs/concepts/config-policies.md
shlomitk1: Added a global file with the version value, plus instructions to increase version upon each update.

pkg/adminconfig/rego_file_evaluator.go
```diff
@@ -58,27 +85,43 @@ func (r *RegoPolicyEvaluator) prepareQuery() (rego.PreparedEvalQuery, error) {
...
 		fileName := filepath.Join(RegoPolicyDirectory, name)
-		module, err := ioutil.ReadFile(filepath.Clean(fileName))
+		module, err := os.ReadFile(filepath.Clean(fileName))
 		if err != nil {
 			return rego.PreparedEvalQuery{}, err
 		}
 		modules[name] = string(module)
 	}
+	r.Timestamp = time.Now()
 	compiler, err := ast.CompileModules(modules)
```
simanadler, line 95: We recompile modules when we are doing a query and discover that the policy files have been updated according to the code here.  (1) I assume this could have a significant performance impact on this particular query.  Is this the best approach to take?  Perhaps this should be done outside of the query path?  (2) This same issue is being discussed about data governance policies.  There we are looking at creating an event when policies change, and having fybrikapplication reconcile run again to see if any changes are required to the data plane - assuming that changes should affect existing workloads and their data planes.  (3) Since changes to existing data planes may not always be required, perhaps we need some indication in the changed policies as to whether they should affect existing workloads.

simanadler (commented):
See comment.  I think there are some bigger issues that may need to be considered.

pkg/adminconfig/rego_file_evaluator.go
shlomitk1: (2) and (3) are about existing data planes. If a data plane exists and the spec of the application did not change, we won't do nothing. Whether this is the right approach or not is a bigger discussion but unrelated to this PR.
(1) Performance impact will be only if policies have changed, which does not happen often.

shlomitk1:
Closing this PR. I'll create a new PR that has customized policy files inside samples/adminconfig, introduces a version for policies, adds instructions how to add a customized policy file, and removes module compilation from the query path, i.e. does this in main.

PR #1178, opened 2022-01-16
Fixes: Missing details in log when module is not chosen by the Fybrik…

### LINKED ISSUE #1168: Missing details in log when module is not chosen by the Fybrik manager
I am trying to use a new module with Fybrik 0.6.0. fybrikapplication reports an error: `Data path could not be constructed` without further details why the path could not be constructed.

simanadler:
Fixes #1168
Makes the error message clearer when a data plane can't be constructed for a given data set. Also prints the catalogID/datasetID and its context as well as the deployed modules to the log.

### DIFF
--- a/manager/controllers/app/plotter_generator.go
+++ b/manager/controllers/app/plotter_generator.go
@@ -142,23 +142,22 @@ func (p *PlotterGenerator) AddFlowInfoForAsset(item DataInfo, appContext *app.Fy
 
 	// DataStore for destination will be determined if an implicit copy is required
 	var sinkDataStore *app.DataStore
-	dataitem := item.Context.CatalogService + "/" + item.Context.DataSetID
 
 	solutions := p.FindPaths(&item, appContext)
 	// No data path found for the asset
 	if len(solutions) == 0 {
 		msg := "Deployed modules do not provide the functionality required to construct a data path"
-		p.Log.Error().Str(logging.DATASETID, dataitem).Msg(msg)
+		p.Log.Error().Str(logging.DATASETID, item.Context.DataSetID).Msg(msg)
 		logging.LogStructure("Data Item Context", item, p.Log, true, true)
 		logging.LogStructure("Module Map", p.Modules, p.Log, true, true)
-		return errors.New(msg + " for " + dataitem)
+		return errors.New(msg + " for " + item.Context.DataSetID)
 	}
-	p.Log.Trace().Str(logging.DATASETID, dataitem).Msg("Generating a plotter")
+	p.Log.Trace().Str(logging.DATASETID, item.Context.DataSetID).Msg("Generating a plotter")
 	selection := solutions[0]
 	datasetID := item.Context.DataSetID
 	for _, element := range selection.DataPath {
 		moduleCapability := element.Module.Spec.Capabilities[element.CapabilityIndex]
-		p.Log.Trace().Str(logging.DATASETID, dataitem).Msgf("Adding module for %s", moduleCapability.Capability)
+		p.Log.Trace().Str(logging.DATASETID, item.Context.DataSetID).Msgf("Adding module for %s", moduleCapability.Capability)
 		actions := element.Actions
 		template := app.Template{
 			Name: string(moduleCapability.Capability),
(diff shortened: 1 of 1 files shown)


manager/controllers/app/plotter_generator.go
```diff
@@ -142,17 +142,21 @@ func (p *PlotterGenerator) AddFlowInfoForAsset(item DataInfo, appContext *app.Fy
 
 	// DataStore for destination will be determined if an implicit copy is required
 	var sinkDataStore *app.DataStore
+	dataitem := item.Context.CatalogService + "/" + item.Context.DataSetID
```
shlomitk1, line 145: CatalogService is an optional field. Since we don't have multiple catalog services in the same deployment, we don't use this field at all. Printing DataSetID is enough.

simanadler: If we store the logs and use them after the deployment no longer exists, this will enable us to know from which data catalog the dataset came.

shlomitk1: Then populate this field from https://github.com/shlomitk1/the-mesh-for-data/blob/5a90ce6490c151ab3741f257dfb3e95d66e07dcc/manager/controllers/utils/configuration.go#L27 in case it is empty.

revit13:
@simanadler @shlomitk1 is it possible to add more details when the module is failed to be selected in the flow starting from [FindPath](https://github.com/fybrik/fybrik/blob/master/manager/controllers/app/plotter_generator.go#L146) function? for example when the module does not [supportsSinkInterface ](https://github.com/fybrik/fybrik/blob/2d1a918a8990580cb3de0c22a38e98dfa64c8cfb/manager/controllers/app/select_modules.go#L207)

manager/controllers/app/plotter_generator.go
simanadler: @shlomitk1 Why do we have it in the item struct if it's not being populated?

shlomitk1: > @shlomitk1 Why do we have it in the item struct if it's not being populated?

As the first step to support multi-catalog setup. At least once we had an issue to support it. The idea was that if an asset is registered in another catalog (not the one in the setup), the catalog service will be populated. An empty field indicates that the asset is registered in the "official" catalog.
I am not aware of a plan to support more than one catalog, so maybe this field is not required. @roytman @roee88 - thoughts?

shlomitk1:
> @simanadler @shlomitk1 is it possible to add more details when the module that is failed to be selected in the flow starting from [FindPath](https://github.com/fybrik/fybrik/blob/master/manager/controllers/app/plotter_generator.go#L146) function? for example when the module does not [supportsSinkInterface ](https://github.com/fybrik/fybrik/blob/2d1a918a8990580cb3de0c22a38e98dfa64c8cfb/manager/controllers/app/select_modules.go#L207)

No. This will end up in misleading messages, such as " a module xxx does not support db2 sink" when the requested interface is s3. What is possible to do is to provide a message similar to "could not construct a data path: source interface: s3,csv; requested interface: fybrik-arrow-flight"

revit13:
> > @simanadler @shlomitk1 is it possible to add more details when the module that is failed to be selected in the flow starting from [FindPath](https://github.com/fybrik/fybrik/blob/master/manager/controllers/app/plotter_generator.go#L146) function? for example when the module does not [supportsSinkInterface ](https://github.com/fybrik/fybrik/blob/2d1a918a8990580cb3de0c22a38e98dfa64c8cfb/manager/controllers/app/select_modules.go#L207)
> 
> No. This will end up in misleading messages, such as " a module xxx does not support db2 sink" when the requested interface is s3. What is possible to do is to provide a message similar to "could not construct a data path: source interface: s3,csv; requested interface: fybrik-arrow-flight"

@shlomitk1 that would be helpful. Thanks!

simanadler:
> 

I added a print of the entire item's context, which includes the interface

revit13 (approved):

shlomitk1:
> > 
> 
> I added a print of the entire item's context, which includes the interface

Can you point to the code? The only LogStructure I saw was for modules, not the item context...

shlomitk1 (changes requested):
Check that item.Context.CatalogService is not empty, otherwise the identifier in the logs will start with "/".

simanadler:
> > > 
> > 
> > 
> > I added a print of the entire item's context, which includes the interface
> 
> Can you point to the code? The only LogStructure I saw was for modules, not the item context...

You're right.  I forgot to push the change to git.  You should see it now.

shlomitk1 (approved):

PR #1193, opened 2022-01-19
Config policies' management

shlomitk1:
This PR includes:
- Extraction customized policies to samples/adminconfig/ folder
- Policy versioning
- Compilation of modules outside reconciler
- Instructions how to add customized policies to Fybrik deployment


site/docs/concepts/config-policies.md
```diff
@@ -135,3 +151,16 @@ config[{"copy": decision}] {
 }
 
 ```
+
+### How to customize policies
```
simanadler, line 155: How to provide custom policies

manager/main.go
```diff
@@ -135,6 +136,12 @@ func run(namespace string, metricsAddr string, enableLeaderElection bool,
 			}
 		}()
 
+		query, err := adminconfig.PrepareQuery()
```
simanadler, line 139: Add comment please describing that this is pre-compiling the config policies

site/docs/concepts/config-policies.md
```diff
@@ -86,6 +87,21 @@ config[{"write": decision}] {
...
+### Versioning 
+
+Policy versioning is done by defining a `version` variable inside the `adminconfig` package. 
+
+charts/fybrik/files/adminconfig/version.rego defines the version for the policies:
+
+```
+package adminconfig
+
+version := "0.0.1"
```
simanadler, line 99: Does this mean that if we have lots of files containing rego policies, and each file has multiple policies in it ... they all have the same version?

Won't it be difficult to know what changed between the versions?  Is there a reason not to have a version in each policy?

simanadler (commented):
Main question is about versioning

manager/main.go
shlomitk1: Done.

site/docs/concepts/config-policies.md
shlomitk1: Modified.

simanadler (approved):

PR #1196, opened 2022-01-19
Fix example 2 in custom taxonomy.

Mohammad-nassar10:

### DIFF
--- a/site/docs/tasks/custom-taxonomy.md
+++ b/site/docs/tasks/custom-taxonomy.md
@@ -62,7 +62,7 @@ helm upgrade fybrik fybrik-charts/fybrik -n fybrik-system --wait --set-file taxo
 
 ## Examples of changing taxonomy
 
-- Example 1: Add new intent for FybrikApplication
+### Example 1: Add new intent for FybrikApplication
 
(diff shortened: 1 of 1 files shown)


site/docs/tasks/custom-taxonomy.md
```diff
@@ -240,18 +260,52 @@ EOF
 ```
 The expected error is `The FybrikModule "taxonomy-module-test" is invalid: spec.capabilities.0.actions.0.name: Invalid value: "RedactYAction": spec.capabilities.0.actions.0.name must be one of the following: "Deny", "RedactAction", "RemoveAction"`. Thus, no FybrikModule CRD was created.
 
-To fix this, a new action name with `RedactYAction` value should be added to the taxonomy. Add a new value of "RedactYAction" in `custom-taxonomy.json` file in `ActionName` property as follows:
+To fix this, a new action name with `RedactYAction` value should be added to the taxonomy. Modify the `taxonomy-layer.yaml` file by add a new value of "RedactYAction". the file should look as follows:
```
simanadler, line 263: Replace "... a new action name with RedactYAction value should be added ..." with "a new action RedactYAction should bbe added ... "

roee88:
@Mohammad-nassar10 please check if adding a separate layer for just RedactYAction works. 
Also, demonstrating on an action that isn't completely made up might be better (and showing action properties other than columns will better demonstrate what an action definition is)

Mohammad-nassar10:
@roee88 I added a separate taxonomy layer for just the new action. When using the taxonomy compiled CLI tool we need to add the two taxonomy layers.
I also changed the `RedactYAction` to `FilterAction` that has `threshold` and `operation` properties.

roee88 (approved):

simanadler (approved):

Issue #927, opened 2021-11-01
Specify user required data flow type in fybrikapplication


froesef:
A user should be able to specify the data flow type(s) that should be provided for the application. Currently the default assumption is always to read. But writing should also be an option as well as possibly specifying a copy that way.

The proposal is to add a new field that contains all data flows that the user is requesting.
```
apiVersion: app.fybrik.io/v1alpha1
kind: FybrikApplication
metadata:
  name: my-notebook
  namespace: fybrik-notebook-sample
  labels:
    app: my-notebook
spec:
  selector:
    clusterName: thegreendragon
    workloadSelector:
      matchLabels:
        app: my-notebook
  appInfo:
    intent: fraud-detection
  data:
    - dataSetID: 'fybrik-notebook-sample/data-csv'
      requirements:
        dataFlows: ["write"]                  <-- new field
        interface:
          protocol: fybrik-arrow-flight
          dataformat: arrow


```

shlomitk1:
This will make `copy -> required: bool ` not needed. As for `copy -> catalog: {}` that indicates how to register a new asset, I suggest to move `catalog` directly under requirements. 
For example:
```
data:
    - dataSetID: 'fybrik-notebook-sample/data-csv'
      requirements:
        dataFlows: 
        - copy
        interface:
          protocol: fybrik-arrow-flight
          dataformat: arrow
        catalog:
           id: myNewCatalog
           service: WKC
```

roee88:
I suggest to take a step back and write down what is really needed in terms of requirements before jumping to spec changes. There is currently no written proposal for write support that I'm aware of. If there is, please share.

"Write" is more than one thing. There is write to an existing dataset, write to a new asset (pre-cataloged by control plane), write to newly allocated storage (allocated by control plane but cataloged by application). It's also not tied to data copy on read in any way.

ronenkat:
Suggest we continue to discuss in #479 or open a new discussion for supporting application and flows that write data.

shlomitk1:
I believe that this issue is about being able to specify the requested data flow. Currently, we have three different scenarios: read, read+copy, ingest (copy). The decision what type is represented by the given fybrikapplication is based on two factors:
- workload selector: non-empty = read
- copy: {required: true/false} specifies the need for copy.
Thus, non-empty workload selector in combination with copy->required=true will generate read+copy scenario, while an empty workload selector together with copy->required=true represents an ingest scenario.
[...]

PR #1258, opened 2022-02-03
Specify user required data flow type in fybrikapplication #927

simanadler:
Addresses issue #927 
Foundation for new flows such as write.

Done so far:

* Spec.Data.DataContext.Flow added - single flow, not array
* Explicit copy "hack" removed.  Can be done via config policies now
* Ingest / Copy  => determined by dataflow instead of dataCtx.Requirements.Copy.Required
* Removed duplicate definition of data flow constants (read, write, delete) across manager and policy compiler
* Change usage to single value in config policy input, instead of map

FybrikApplication interface changes
* No change needed for pure read!!
* Copy has to be indicated in dataCtx.Flows instead of dataCtx.Requirements.Copy.Required
* CopyRequirements renamed to FlowParams, and is used by multiple flows

Config Policy changes
* Usage (indication of flow) parameter in the input is now a single value, so the base policies implementations have changed.  ==> Anyone who has written custom config policies will need to update them

* Updates to taxonomy

Out of scope:
* Implementation of new flows (these are separate issues)
 

shlomitk1:
Many auto-generated crds have been changed for no reason. Seems as if you are using out-dated tools (controller-gen?)
Please re-run 'make install-tools' after you remove all files from hack/tools/bin and recompile.

manager/apis/app/v1alpha1/fybrikapplication_types.go
```diff
@@ -9,26 +9,29 @@ import (
...
 
-// CatalogRequirements contain the specifics for catalogging the data asset
-type CatalogRequirements struct {
-	// CatalogService specifies the datacatalog service that will be used for catalogging the data into.
-	// +optional
-	CatalogService string `json:"service,omitempty"`
-
+// TargetCatalog contain the specifics for the catalog which needs to be updated.
+// Updates may be adding a new dataset to the catalog (currently supported) or deleting an existing dataset (future)
+type TargetCatalog struct {
```
shlomitk1, line 14: When deleting an existing dataset, the dataset id contains all relevant information including catalog id.
Regarding adding a new asset, are we assuming the catalog already exists? Do all catalogs support a catalog concept? For example, katalog - is this going to be a namespace? 
We might need more data for the registration. I suggest to keep it in the designated structure instead of globally.

manager/apis/app/v1alpha1/fybrikapplication_types.go
```diff
@@ -9,26 +9,29 @@ import (
...
+	// Catalog indicates that the data asset must be cataloged, and in which catalog to register it.
 	// +optional
-	Required bool `json:"required,omitempty"`
+	Catalog TargetCatalog `json:"catalog,omitempty"`
 
-	// Catalog indicates that the data asset must be cataloged.
+	// Storage estimate indicates the estimated amount of storage required when copying or writing new data.
 	// +optional
-	Catalog CatalogRequirements `json:"catalog,omitempty"`
+	StorageEstimate int `json:"storageEstimate"`
```
shlomitk1, line 29: It is optional, add omitempty.

manager/apis/app/v1alpha1/fybrikapplication_types.go
```diff
@@ -9,26 +9,29 @@ import (
...
+
+	// IsNewDataSet if true indicates that the DataContext.DataSetID is user provided and not a true catalog ID.  Relevant when writing.
+	// A unique ID from the catalog will be provided in the FybrikApplication Status after a new catalog entry is created.
+	// +optional
+	IsNewDataSet bool `json:"isNewDataSet"`
```
shlomitk1, line 34: So maybe allow an empty datasetID instead if there is no dataset?

manager/apis/app/v1alpha1/fybrikapplication_types.go
```diff
@@ -37,32 +40,34 @@ type DataRequirements struct {
...
+	// When writing a new dataset it is the name provided by the user or workload generating it, and should include the sub-catalog name where appropriate.
 	// +required
 	// +kubebuilder:validation:MinLength=1
 	DataSetID string `json:"dataSetID"`
 
-	// CatalogService represents the catalog service for accessing the requested dataset.
-	// If not specified, the enterprise catalog service will be used.
+	// Flows indicates what is being done with the particular dataset - ex: read, write, copy, delete
+	// If more than one flow is indicated, the order is respected.
+	// This is optional for the purpose of backward compatability.  If nothing is provided, read is assumed.
```
shlomitk1, line 59: So for the read scenario we have either [] or [read], and for ingest scenario we have [copy], correct?

manager/apis/app/v1alpha1/fybrikapplication_types.go
```diff
@@ -37,32 +40,34 @@ type DataRequirements struct {
...
-// and includes information about the data format and technologies used by the application
-// to access the data.
+// DataContext indicates data set being processed by the workload
+// and includes information about the data format and technologies used to access the data.
 type DataContext struct {
-	// DataSetID is a unique identifier of the dataset chosen from the data catalog for processing by the data user application.
+	// DataSetID is a unique identifier of the dataset chosen from the data catalog.  For data catalogs that support multiple sub-catalogs, it includes the catalog id and the dataset id.
+	// When writing a new dataset it is the name provided by the user or workload generating it, and should include the sub-catalog name where appropriate.
 	// +required
 	// +kubebuilder:validation:MinLength=1
```
shlomitk1, line 54: could be empty if only for write

manager/apis/app/v1alpha1/fybrikapplication_types.go
simanadler: Correct.

manager/apis/app/v1alpha1/fybrikapplication_types.go
simanadler: If it's write and a new dataset, the user puts here the catalogid/filename.  And the in FlowParams the flag IsNewDataSet should be true

manager/apis/app/v1alpha1/fybrikapplication_types.go
simanadler: See answer to following comment

manager/apis/app/v1alpha1/fybrikapplication_types.go
simanadler: Yes

manager/apis/app/v1alpha1/fybrikapplication_types.go
simanadler: We will have several modes for write:
(1) Write a new dataset - fybrik allocates storage and registers a new entry in the catalog.  Catalog and filename in datasetid, and flowparams.isnewdataset = true
(2) Write a new dataset where storage has already been allocated by the user.  User creates the new catalog entry and indicates it to Fybrik in DataSetID.  No need for Fybrik to not register anything in the catalog.
(3) Write to existing dataset - Same as #2 except that the user doesn't have to create a new catalog entry.

FlowParams is intentionally separate so we can expand it.  I'm hoping to avoid having structure per dataflow type

connectors/opa/connector_test.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/gin-gonic/gin"
 	"github.com/stretchr/testify/assert"
 
+	app "fybrik.io/fybrik/manager/apis/app/v1alpha1"
```
shlomitk1, line 19: This does not seem right. Common definitions should be defined in pkg/model/... and be used by different components.

manager/apis/app/v1alpha1/fybrikapplication_types.go
shlomitk1: (2) and (3) are similar from the spec perspective. DatasetID points to the asset metadata + connection, so all is set.
I don't understand (1). Fybrik will allocate storage. For registration, it will need a full/partial identifier of the catalog/asset. It will also need to register the metadata (credentials come from Fybrik). How? Do you mean that in this case users will register the new metadata in some asset yaml and point to this yaml in DatasetID while specifying that this is not the real dataset id? 
Copy (explicit) is even more challenging. DatasetID will point to a real asset that needs to be copied using allocated storage or not, and registered or not. How will you specify all this information?

manager/testdata/unittests/ingest.yaml
```diff
@@ -13,9 +13,10 @@ spec:
     intent: Fraud Detection
   data:
     - dataSetID: s3-external/xxx
+      flows:
+      - copy
       requirements:
         copy:
```
shlomitk1, line 19: why is copy under requirements?

manager/apis/app/v1alpha1/fybrikapplication_types.go
```diff
@@ -37,32 +40,34 @@ type DataRequirements struct {
...
 	DataSetID string `json:"dataSetID"`
 
-	// CatalogService represents the catalog service for accessing the requested dataset.
-	// If not specified, the enterprise catalog service will be used.
+	// Flows indicates what is being done with the particular dataset - ex: read, write, copy, delete
```
roytman, line 57: we separate `read` and `copy`, should we separate `write` (add new rows, append data ) and `update` (update existing rows, change the existing data)

manager/apis/app/v1alpha1/fybrikapplication_types.go
```diff
@@ -37,32 +40,34 @@ type DataRequirements struct {
...
 	// +required
 	Requirements DataRequirements `json:"requirements"`
 }
 
 // FybrikApplicationSpec defines data flows needed by the application, the purpose and other contextual information about the application.
-// Read flow - if selector is populated, fybrik builds a data plane for reading the specified data sets
-// Ingest flow - if no selector, and data/copy/required is true then the data specified is copied into a bucket allocated by fybrik and is cataloged in the data catalog
+// Read flow - if selector is populated, fybrik builds a data plane for reading the specified data sets.
+// Ingest flow - if no selector, and data/copy/required is true then the data specified is copied into a bucket allocated by fybrik and is cataloged in the data catalog.
 type FybrikApplicationSpec struct {
```
roytman, line 71: why do we still need this selector based separation?

manager/apis/app/v1alpha1/fybrikapplication_types.go
roytman: actually, in the case of v0.6.0 Flows will not be [], it will be nil

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -405,15 +405,18 @@ func (r *FybrikApplicationReconciler) reconcile(applicationContext ApplicationCo
...
 func CreateDataRequest(application *api.FybrikApplication, dataCtx api.DataContext, assetMetadata *datacatalog.ResourceMetadata) adminconfig.DataRequest {
-	usage := make(map[api.DataFlow]bool)
-	// request to read is determined by the workload selector presence
-	usage[api.ReadFlow] = (application.Spec.Selector.WorkloadSelector.Size() > 0)
-	// explicit request to copy
-	usage[api.CopyFlow] = dataCtx.Requirements.Copy.Required
+	var flows []api.DataFlow
+
+	// If a workload selector is provided but no flow, assume read - for backward compatability
+	if (application.Spec.Selector.WorkloadSelector.Size() > 0) && (len(dataCtx.Flows) == 0) {
```
roytman, line 411: we have to check that `Flows` is not nil.

pkg/model/policymanager/model.go
```diff
@@ -4,22 +4,13 @@
...
-// List of operationType
-const (
-	READ   RequestActionType = "read"
-	WRITE  RequestActionType = "write"
-	DELETE RequestActionType = "delete"
-)
-
 type RequestAction struct {
-	ActionType         RequestActionType           `json:"actionType"`
+	ActionType         api.DataFlow                `json:"actionType"`
```
roytman, line 13: `DataFlow` definitions should be moved to `pkg/model/taxonomy/module.go` and defined as 
```
// +kubebuilder:validation:Enum=read;write;delete;copy
type DataFlow string

const (
	// ReadFlow indicates a data set is being read
	ReadFlow DataFlow = "read"

	// WriteFlow indicates a data set is being written
	WriteFlow DataFlow = "write"

	// DeleteFlow indicates a data set is being deleted
	DeleteFlow DataFlow = "delete"

	// CopyFlow indicates a data set is being copied
	CopyFlow DataFlow = "copy"
)
```

charts/fybrik/files/taxonomy/external.json
```diff
@@ -0,0 +1,630 @@
+{
```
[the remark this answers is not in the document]
simanadler: Something generated by Fybrik.  @roee88 can you explain?

manager/apis/app/v1alpha1/fybrikapplication_types.go
simanadler: @roytman I have a separate issue for addressing adding metadata / tags.  It will have to be provided by the user ... unless we had a module that runs and analyzes the data.  That's out of scope though

simanadler: > (2) and (3) are similar from the spec perspective. DatasetID points to the asset metadata + connection, so all is set. I don't understand (1). Fybrik will allocate storage. For registration, it will need a full/partial identifier of the catalog/asset. It will also need to register the metadata (credentials come from Fybrik). How? Do you mean that in this case users will register the new metadata in some asset yaml and point to this yaml in DatasetID while specifying that this is not the real dataset id? 

No.  The user will provide in DatasetID something like ex: HRCatalog/JanSalaries.  This is not a unique identifier.  It tells Fybrik that HRCatalog is the catalog in which the file JanSalaries should be registered.  Fybrik knows this is not a unique ID because flowparams.isnewdataset = true


Copy (explicit) is even more challenging. DatasetID will point to a real asset that needs to be copied using allocated storage or not, and registered or not. How will you specify all this information?

DatasetID will be the catalog id of source data set (the same way we did it for ingest).  If the copy needs to be registered in the catalog then DataContext.FlowParams.Catalog will contain the name of the catalog in which to register the copy.  Name of the dataset will be copied from the source.  If DataContext.FlowParams.Catalog is not indicated, then the asset won't be cataloged (not a good idea though).

manager/testdata/unittests/ingest.yaml
simanadler: Mistake.  Ingest unit tests fixed.

manager/apis/app/v1alpha1/fybrikapplication_types.go
simanadler: Old comments removed

pkg/model/policymanager/model.go
simanadler: Done.  Thanks!

manager/apis/app/v1alpha1/fybrikapplication_types.go
simanadler: If we decide we want in the future a dedicated update flow the changes in this PR should be able to support it.

charts/fybrik/files/adminconfig/default_policies.rego
```diff
@@ -4,35 +4,35 @@ package adminconfig
...
-    input.request.usage.read == true
+    input.request.usage[_] == "read"
     policy := {"ID": "read-default-enabled", "description":"Read capability is requested for read workloads", "version": "0.1"}
     decision := {"policy": policy, "deploy": "True"}
 }
 
 # read capability deployment
 config[{"capability": "read", "decision": decision}] {
-    input.request.usage.read == false
+    input.request.usage[_] != "read"
```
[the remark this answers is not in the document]
simanadler: Will do.

manager/apis/app/v1alpha1/plotter_types.go
```diff
@@ -107,7 +107,7 @@ type SubFlow struct {
 
 	// Type of the flow (e.g. read)
 	// +required
-	FlowType DataFlow `json:"flowType"`
+	FlowType taxonomy.DataFlow `json:"flowType"`
```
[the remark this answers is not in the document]
simanadler: yes

manager/controllers/app/fybrikapplication_controller_unit_test.go
```diff
@@ -744,6 +745,7 @@ func TestCopyData(t *testing.T) {
 	application := &app.FybrikApplication{}
 	g.Expect(readObjectFromFile("../../testdata/unittests/ingest.yaml", application)).NotTo(gomega.HaveOccurred())
 	application.Spec.Data[0].DataSetID = assetName
+	application.Spec.Data[0].Flows = []taxonomy.DataFlow{taxonomy.CopyFlow}
```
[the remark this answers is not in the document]
simanadler: There are several read tests.  In one of them I changed the yaml file to have an empty flow list, rather than creating an additional test.

charts/fybrik/files/taxonomy/external.json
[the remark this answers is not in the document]
roytman: @simanadler @shlomitk1 , `json-schema-generator ` puts definitions of all objects that are not in the taxonomy model into this file. It was created when `DataFlow` was a part of the fybrik APIs. Now when it was moved into the `taxonomy.model`, the `extrnal.json` is not generated, and this one can/should be removed.

manager/apis/app/v1alpha1/fybrikapplication_types.go
```diff
@@ -10,26 +10,19 @@ import (
...
-
-// FlowRequirements include the requirements for the data copy operation when there is an explicit copy request
+// FlowRequirements include the requirements specific to the flow
 // Note: Implicit copies done for data plane optimization by Fybrik do not use these parameters
 type FlowRequirements struct {
-	// Catalog indicates that the data asset must be cataloged, and in which catalog to register it.
+	// Catalog indicates that the data asset must be cataloged, and in which catalog to register it
 	// +optional
-	Catalog TargetCatalog `json:"catalog,omitempty"`
+	Catalog string `json:"catalog,omitempty"`
```
shlomitk1, line 18: perhaps TargetCatalogID to avoid confusions?

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -408,10 +408,10 @@ func CreateDataRequest(application *api.FybrikApplication, dataCtx api.DataConte
 	var flows []taxonomy.DataFlow
```
shlomitk1, line 408: why array instead of a single object?

charts/fybrik/files/adminconfig/default_policies.rego
```diff
@@ -4,35 +4,35 @@ package adminconfig
 
 # read capability deployment
 config[{"capability": "read", "decision": decision}] {
-    input.request.usage.read == true
+    input.request.usage[_] == "read"
```
shlomitk1, line 7: Why are policies changed if this is a single flow?

simanadler: Leaving it as a map makes usage more complex to populate, especially when we want to allow custom flow definitions

manager/apis/app/v1alpha1/plotter_types.go
```diff
@@ -129,7 +129,7 @@ type Flow struct {
...
```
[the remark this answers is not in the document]
simanadler: Removed the array of flows.  Now there is a single flow per datasetid.  FlowRequirements.StorageEstimate  won't get propagated to the plotter.  Rather the manager will use it to decide which modules to choose.  I.e. It will be passed as a parameter to the config policies.   FlowRequirements.IsNewDataSet is used to decide whether DataSetID is an existing entry in the catalog, or whether it indicates only the dest catalog / object name ... and thus Fybrik needs to create a new entry in the catalog.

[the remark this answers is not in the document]
shlomitk1: @roytman Plotter won't be aware of any of these. As I understand, storage estimate and new/existing dataset are used to determine the existing plotter arguments: where to allocate storage, data destination, etc.

charts/fybrik/files/adminconfig/default_policies.rego
simanadler: OK.  Will require updates to some of the tests in rego_file_evaluator_test.go in a follow up PR.

manager/controllers/app/fybrikapplication_controller.go
simanadler: Changed to single object

charts/fybrik/files/taxonomy/fybrik_application.json
```diff
@@ -14,6 +14,12 @@
                     "items": {
                         "type": "object",
                         "properties": {
+                            "flows": {
+                                "type": "array",
```
shlomitk1, line 18: Where do "flows" come from? It is not an array any longer, right?

manager/controllers/app/fybrikapplication_controller_unit_test.go
```diff
@@ -752,6 +753,7 @@ func TestCopyData(t *testing.T) {
 	application := &app.FybrikApplication{}
 	g.Expect(readObjectFromFile("../../testdata/unittests/ingest.yaml", application)).NotTo(gomega.HaveOccurred())
 	application.Spec.Data[0].DataSetID = assetName
+	application.Spec.Data[0].Flow = taxonomy.CopyFlow
```
shlomitk1, line 756: why not change the flow inside the yaml?

sample-policy.rego
```diff
@@ -0,0 +1,9 @@
+package dataapi.authz
+
+rule[{"action": {"name":"RedactAction", "columns": column_names}, "policy": description}] {
```
shlomitk1, line 3: I don't think this file should be committed.

shlomitk1:
I still don't feel comfortable with the idea of re-using datasetID based on read/write and marking it as a new dataset for write.
How would you create a FybrikApplication for the following use-case:
The data source is A. It needs to be copied to B (periodically). Both A and B are registered in the catalog.
I would suggest to use
- datasetID as the data source (from the catalog) which will be left empty on write
- targetDatasetID as the data destination, empty if not relevant
- check whether targetDatasetID exists or not in the code, not trusting the spec. There may be a situation when the same application is invoked again, periodically or upon policy change, and the spec will be no longer relevant, i.e. the dataset already exists.

sample-policy.rego
simanadler: Removed

manager/controllers/app/fybrikapplication_controller_unit_test.go
simanadler: Was in both.  Removed the redundancy

charts/fybrik/files/taxonomy/fybrik_application.json
simanadler: Fixed

manager/apis/app/v1alpha1/fybrikapplication_types.go
```diff
@@ -5,30 +5,27 @@ package v1alpha1
...
+// FlowRequirements include the requirements specific to the flow
+// Note: Implicit copies done for data plane optimization by Fybrik do not use these parameters
+type FlowRequirements struct {
+	// Catalog indicates that the data asset must be cataloged, and in which catalog to register it
 	// +optional
-	CatalogService string `json:"service,omitempty"`
+	Catalog string `json:"catalog,omitempty"`
 
-	// CatalogID specifies the catalog where the data will be cataloged.
+	// Storage estimate indicates the estimated amount of storage required when writing new data.
```
roytman, line 20: The comment will be propagated to the Web site, should we add examples of the estimated storage,  MB, GB, TB ...

manager/apis/app/v1alpha1/fybrikapplication_types.go
```diff
@@ -37,32 +34,34 @@ type DataRequirements struct {
 	// +required
 	Interface InterfaceDetails `json:"interface"`
 
-	// CopyRequrements include the requirements for copying the data
+	// FlowParams include the requirements for explicit requests to copy the data
```
roytman, line 37: Is there some inconsistency here? The comment says: "FlowParams include the requirements for explicit requests to copy the data" 

- Does is relevant to `copy` only?
- The `FlowRequirements` comments say:

> FlowRequirements include the requirements specific to the flow (AR, any flow?)
> Note: Implicit copies done for data plane optimization by Fybrik do not use these parameters

manager/apis/app/v1alpha1/fybrikapplication_types.go
```diff
@@ -37,32 +34,34 @@ type DataRequirements struct {
...
+	// For data catalogs that support multiple sub-catalogs, it includes the catalog id and the dataset id.
+	// When writing a new dataset it is the name provided by the user or workload generating it.
 	// +required
 	// +kubebuilder:validation:MinLength=1
 	DataSetID string `json:"dataSetID"`
 
-	// CatalogService represents the catalog service for accessing the requested dataset.
-	// If not specified, the enterprise catalog service will be used.
+	// Flows indicates what is being done with the particular dataset - ex: read, write, copy, delete
+	// If more than one flow is indicated, the order is respected.
```
roytman, line 53: the comment about order is not relevant

charts/fybrik/files/adminconfig/default_policies.rego
```diff
@@ -4,35 +4,35 @@ package adminconfig
...
-    input.request.usage.read == true
+    input.request.usage == "read"
     policy := {"ID": "read-default-enabled", "description":"Read capability is requested for read workloads", "version": "0.1"}
     decision := {"policy": policy, "deploy": "True"}
 }
 
 # read capability deployment
 config[{"capability": "read", "decision": decision}] {
-    input.request.usage.read == false
+    input.request.usage != "read"
```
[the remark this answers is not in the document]
shlomitk1: The intent was "Read capability is requested only for read workloads". Anyway, I believe this and the similar policy for not writing can be removed because I can envision a flow (update) that requires both read and write.

charts/fybrik/files/adminconfig/default_policies.rego
```diff
@@ -4,35 +4,35 @@ package adminconfig
...
-    input.request.usage.write == false
+    input.request.usage != "write"
     policy := {"ID": "write-default-disabled", "description":"Write capability is requested for workloads that write data", "version": "0.1"}
     decision := {"policy": policy, "deploy": "False"}
 }
 
 # copy requested by the user
 config[{"capability": "copy", "decision": decision}] {
-    input.request.usage.copy == true
+    input.request.usage == "copy"
```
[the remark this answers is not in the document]
shlomitk1: Not the negations. usage != "delete" does not mean the delete won't happen, maybe for "move" we'll need both copy and delete. As for delete and move, I think that now it is useless to add policies for capabilities that we don't support.

manager/apis/app/v1alpha1/fybrikapplication_types.go
shlomitk1: Catalog is used only for registration of a new asset. There is no need for SourceCatalogID as datasetID includes this information.

manager/apis/app/v1alpha1/fybrikapplication_types.go
roytman: In that case, maybe - `NewCatalogID`? 
`TargetCatalogID` can confuse in the `delete` flow

simanadler: @roytman This field would not be used in the case of delete.  The DataSetID would contain the catalog and datsetid of what needs to be deleted.  I will update the comment appropriately.

manager/apis/app/v1alpha1/fybrikapplication_types.go
simanadler: Thanks.  Will fix

manager/apis/app/v1alpha1/fybrikapplication_types.go
simanadler: Will fix comment

manager/apis/app/v1alpha1/fybrikapplication_types.go
simanadler: Adding comment about storage units

simanadler:
> I still don't feel comfortable with the idea of re-using datasetID based on read/write and marking it as a new dataset for write. How would you create a FybrikApplication for the following use-case: The data source is A. It needs to be copied to B (periodically). Both A and B are registered in the catalog. 

It was decided not to support explicit (user requested) copies in the FybrikApplication Spec, because we shouldn't be asking the user to explicitly define flows.  The user/workload wants to read/write/delete data.  If copies are needed he shouldn't need to deal with that.

Instead there may be contextual information provided by the user in appInfo.  This can be used in config policies, which can determine if/when copies should be made. Fybrik then orchestrates these copies.  The exception to this is ingest, which we do already support.

I would suggest to use
> 
>     * datasetID as the data source (from the catalog) which will be left empty on write

datasetID is used throughout the manager.  Leaving it empty will complicate things unnecessarily in my opinion.  We will have to automatically generate some other key instead
> 
>     * targetDatasetID as the data destination, empty if not relevant
> 
>     * check whether targetDatasetID exists or not in the code, not trusting the spec. There may be a situation when the same application is invoked again, periodically or upon policy change, and the spec will be no longer relevant, i.e. the dataset already exists.

The connector can return a warning or error in such instances.

charts/fybrik/files/adminconfig/default_policies.rego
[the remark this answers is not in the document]
simanadler: Removing

charts/fybrik/files/adminconfig/default_policies.rego
```diff
@@ -4,35 +4,35 @@ package adminconfig
...
-    input.request.usage.write == true
+    input.request.usage == "write"
     policy := {"ID": "write-default-enabled", "description":"Write capability is requested for workloads that write data", "version": "0.1"}
     decision := {"policy": policy, "deploy": "True"}
 }
 
 # write capability deployment
 config[{"capability": "write", "decision": decision}] {
-    input.request.usage.write == false
+    input.request.usage != "write"
```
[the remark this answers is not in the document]
simanadler: Removing

roytman (approved):

shlomitk1 (approved):

PR #1279, opened 2022-02-08
Remove path restrictions from the web update gitHub action

### LINKED ISSUE #1278: Update the website-deploy github action.
In the previous Fybrik versions the web site recommends to install the latest Fybrik version, now it points to a specific version - the latest minor version.  Therefore,  if in the past the site should be updated only when documentation was changed, not we should update it per each release.

roytman:
In the previous Fybrik versions the web site recommended to install the latest Fybrik version, now it points to a specific version - the latest minor version. Therefore, if in the past the site should be updated only when documentation was changed, not we should update it per each release.

/fixes #1278 

### DIFF
--- a/.github/workflows/website-deploy.yml
+++ b/.github/workflows/website-deploy.yml
@@ -5,11 +5,6 @@ on:
       - '^v[0-9]+\.[0-9]+(\.[0-9]+)*'
     branches:
       - master
-    paths:
-      - site/**
-      - .github/workflows/website-*
-      - pkg/connectors/protos/**
-      - charts/fybrik-crd/templates/**
   workflow_dispatch:
 jobs:
   website-deploy:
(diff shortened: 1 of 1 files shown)


.github/workflows/website-deploy.yml
```diff
@@ -30,6 +25,7 @@ jobs:
           pip install mkdocs-material=="8.*"
           pip install mike
           pip install mkdocs-macros-plugin
+          pip install lunr
```
revit13, line 28: why this is needed?

.github/workflows/website-deploy.yml
```diff
@@ -5,11 +5,6 @@ on:
       - '^v[0-9]+\.[0-9]+(\.[0-9]+)*'
     branches:
       - master
-    paths:
```
revit13, line 8: The [documentation ](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions) mention that `Path filters are not evaluated for pushes of tags.`  so the site will be deployed upon a tag regardless if the site was changed... am I missing something?

.github/workflows/website-deploy.yml
roytman: the action's logs contain a warning about missing lunr. In general lunr is a simple search agent.

PR #1282, opened 2022-02-09
Added module restrictions validation when selecting modules

### LINKED ISSUE #964: Task: respect module restrictions while selecting a module
Admin config policies may affect the choice of the selected modules by applying some constraints on them, for example: module type should be `plugin`, or scope should be `workload`. However, the current implementation ignores module restrictions while selecting a module.

YingchaoDu:
Fixes #964.


manager/controllers/app/select_modules.go
```diff
@@ -346,7 +346,30 @@ func allowCapability(item *DataInfo, capability taxonomy.Capability) bool {
...
+		return true
+	}
+	moduleCapability := edge.Module.Spec.Capabilities[edge.CapabilityIndex]
+	capabilityDetails, err := utils.StructToMap(&moduleCapability)
+	if err != nil {
+		return false
+	}
+	for key, values := range restrictions {
+		fields := strings.Split(key, ".")
+		value, found, err := unstructured.NestedString(capabilityDetails, fields...)
```
shlomitk1, line 365: This won't work for a property that belongs to a module capability. For example, "capabilities.scope" does not really exist. For a path capabilities.XXX you should look for XXX inside a selected capability, or change to 
capabilities.<edge.CapabilityIndex>.XXX

YingchaoDu: Thanks for the comment. I'd like to confirm that for the module restriction, it can be any property inside a moduleCapability or it is limited to Scope?

shlomitk1: This can be any property inside a capability, or in the module spec unrelated to the capabilities, e.g. "type"

manager/controllers/app/select_modules.go
```diff
@@ -354,9 +354,16 @@ func allowCapability(item *DataInfo, capability taxonomy.Capability) bool {
...
 }
 
-func validateModuleRestrictions(item *DataInfo, edge *Edge) bool {
-	// TODO(shlomitk1): validate module restrictions
-	return true
+func (p *PlotterGenerator) validateModuleRestrictions(item *DataInfo, edge *Edge) bool {
+	capability := edge.Module.Spec.Capabilities[edge.CapabilityIndex]
+	moduleSpec := edge.Module.Spec
+	restrictions := item.Configuration.ConfigDecisions[capability.Capability].DeploymentRestrictions.Modules
+	for i := 0; i < len(restrictions); i++ {
```
shlomitk1, line 361: for i := range restrictions

manager/controllers/app/select_modules.go
```diff
@@ -430,3 +437,31 @@ func (p *PlotterGenerator) validateRestrictions(restrictions []adminrules.Restri
 	}
 	return true
 }
+
+func NestedFieldNoCopy(obj map[string]interface{}, fields ...string) (interface{}, bool, error) {
```
shlomitk1, line 441: Why did you implement this function on your own?

YingchaoDu: Because the original method can only deal with nested maps, like map[a: map[b: map[c: d]]]. However, in our case, we should process something like map[capabilities: [ map[scope: workload], map[ a: b]]]. That's map>slice>map. So I add line 448-456 to address the slice part.

manager/controllers/app/select_modules.go
YingchaoDu: Changed!

shlomitk1 (approved):

PR #1290, opened 2022-02-14
Generate json schemas for CRDs.

### LINKED ISSUE #1091: Generate json schemas for validation from FybrikApplication and FybrikModule code
Looking at the flow for taxonomy validation of CRDs (e.g., FybrikApplication and [FybrikModule](https://github.com/fybrik/fybrik/pull/804)) I'm a bit worried about keeping consistency between the structure in the CRD and the structure in the validation object. The latter must be a subset of the former with `additionalProperties: true` set in the right places to allow the additional fields that don't require validation. Whenever the structure of the CRD would change we will need to be careful about the structure of the validation object. 

_Originally posted by @roee88 in https://github.com/fybrik/fybrik/discussions/848_

Mohammad-nassar10:
Generate json schemas for `fybrikApplication` and `fybrikModule` automatically using an updated version of the `json-schema-generator`. Also, adding a test `fybrikmodule_webhook_test.go` for checking the `fybrikModule` taxonomy with valid and invalid `fybrikModules` 

Closes #1091 

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,24 +56,8 @@ jobs:
         restore-keys: ${{ runner.os }}-t-${{ hashFiles('hack/make-rules/tools.mk') }}-go
     - name: Install tools
       run: make install-tools
-<<<<<<< HEAD
-<<<<<<< HEAD
-<<<<<<< HEAD
-<<<<<<< HEAD
     - name: install json-schema-generator
       run: ./hack/tools/install_json-schema-generator.sh
-=======
->>>>>>> Update schema.
-=======
-    - name: install json-schema-generator
-      run: ./hack/tools/install_json-schema-generator.sh
->>>>>>> Update build workflow.
-=======
->>>>>>> Update schema.
-=======
-    - name: install json-schema-generator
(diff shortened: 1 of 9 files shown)


shlomitk1:
Why is this PR required? Please connect to the relevant issue.

shlomitk1 (commented):
Please rebase. There are changes that are in master, yet marked as diffs here.

manager/controllers/app/fybrikmodule_controller.go
```diff
@@ -107,7 +107,8 @@ func ValidateFybrikModule(module *fapp.FybrikModule, taxonomyFile string) error
 	var allErrs []*field.Error
 
 	// Convert Fybrik module Go struct to JSON
-	moduleJSON, err := json.Marshal(module)
+	spec := module.Spec.DeepCopy()
+	moduleJSON, err := json.Marshal(spec)
```
shlomitk1, line 111: Why do you need a deep copy here? json.Marshal(&module.Spec) should be enough, you are not using the saved pointer, just a list of bytes...

roytman:
@Mohammad-nassar10 please rebase again

PR #1297, opened 2022-02-16
unit test cases for select_modules.go

hyiyang2022:
Signed-off-by: hyiyang <hyiyang@cn.ibm.com>

This PR added ut case for the following functions:
- CheckDependencies
- SupportsDependencies
- GetDependencies

Right now, modules under manager/testdata have no dependencies, we can add some more cases when got more modules that have dependencies.

### DIFF
--- a/manager/controllers/app/select_modules_unit_test.go
+++ b/manager/controllers/app/select_modules_unit_test.go
@@ -1,0 +1,128 @@
+// Copyright 2020 IBM Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+package app
+
+import (
+	"context"
+	"testing"
+
+	app "fybrik.io/fybrik/manager/apis/app/v1alpha1"
+	"fybrik.io/fybrik/manager/controllers/utils"
+	"github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


manager/controllers/app/select_modules_unit_test.go
```diff
@@ -0,0 +1,128 @@
...
+	// Set the logger to development mode for verbose logs.
+	logf.SetLogger(zap.New(zap.UseDevMode(true)))
+
+	application := &app.FybrikApplication{}
+	g.Expect(readObjectFromFile("../../testdata/unittests/fybrikcopyapp-csv.yaml", application)).To(gomega.BeNil(), "Cannot read fybrikapplication file for test")
+	application.SetGeneration(1)
+	application.SetUID("1")
+	// Objects to track in the fake client.
+	objs := []runtime.Object{
+		application,
```
shlomitk1, line 31: This and other tests create an application object which is not relevant for the tested functions.
Instead of creating modules with `cl.Create`, you could insert them into `objs `here.

hyiyang2022: Thanks for the review. I will address your comments later and create a new PR.

Issue #1198, opened 2022-01-20
Smart blueprint: taxonomy for infrastructure json object


shlomitk1:
- [ ] Add sample taxonomy for infrastructure attributes
- [ ] Validate infrastructure.json after reading the file

YingchaoDu:
Implemented

PR #1302, opened 2022-02-16
Added taxonomy validation for infrastructure json object

YingchaoDu:
Fixes issue #1198. 


pkg/infrastructure/attribute_manager.go
```diff
@@ -46,12 +51,27 @@ func readInfrastructure() (infraattributes.Infrastructure, error) {
 	if err != nil {
 		return attributes, err
 	}
+	if err := validateStructure(content, ValidationPath); err != nil {
```
shlomitk1, line 54: Shouldn't it be `err =`  instead of `err :=` because err is already defined?

roee88: `err` here is scoped to the if statement

samples/taxonomy/infrastructure/attribute.yaml
```diff
@@ -0,0 +1,5 @@
+definitions:
+  Attribute:
+    enum:
+      - storage-cost
+      - bandwidth
```
shlomitk1, line 5: new line at the end of the file

samples/taxonomy/infrastructure/units.yaml
```diff
@@ -0,0 +1,6 @@
+definitions:
+  Units:
+    enum:
+      - US-dollar
+      - GB
+      - GB/sec
```
shlomitk1, line 6: new line at the end of the file

samples/taxonomy/infrastructure/units.yaml
```diff
@@ -0,0 +1,6 @@
+definitions:
+  Units:
+    enum:
+      - US-dollar
```
shlomitk1, line 4: US Dollar (to be consistent with an example)

shlomitk1 (commented):
Add tests for validating a correct infrastructure object and an invalid infrastructure object

samples/taxonomy/infrastructure/units.yaml
shlomitk1: Replace GB/sec with GB/s
Add also Mbp/s, Gbp/s, Tbp/s, MB/s, TB/s, MB, KB, TB

pkg/infrastructure/attribute_manager.go
```diff
@@ -47,12 +52,27 @@ func readInfrastructure() (infraattributes.Infrastructure, error) {
...
+		return attributes, err
+	}
 	if err = json.Unmarshal(content, &attributes); err != nil {
 		return attributes, errors.Wrap(err, "could not parse infrastructure json")
 	}
 	return attributes, nil
 }
 
+func validateStructure(bytes []byte, taxonomySchema string) error {
+	allErrs, err := validate.TaxonomyCheck(bytes, ValidationPath)
```
shlomitk1, line 65: Use taxonomySchema instead of ValidationPath in the function call

shlomitk1 (approved):

samples/taxonomy/infrastructure/units.yaml
```diff
@@ -0,0 +1,14 @@
+definitions:
+  Units:
```
simanadler, line 2: Do we want all the different types of units in one enum?  Or should there be separate ones for cost, size, etc?

pkg/infrastructure/attribute_manager.go
```diff
@@ -47,12 +52,27 @@ func readInfrastructure() (infraattributes.Infrastructure, error) {
...
+	if err = validateStructure(content, ValidationPath); err != nil {
+		return attributes, err
+	}
 	if err = json.Unmarshal(content, &attributes); err != nil {
 		return attributes, errors.Wrap(err, "could not parse infrastructure json")
 	}
 	return attributes, nil
 }
 
+func validateStructure(bytes []byte, taxonomySchema string) error {
```
simanadler, line 64: Should this be a general helper function provided by the taxonomy, or is it truly specific to the attribute manager?

shlomitk1: It's just a wrapper for the function in `validate ` package

samples/taxonomy/infrastructure/units.yaml
shlomitk1: Will be handled in https://github.com/fybrik/fybrik/issues/1331

PR #1305, opened 2022-02-17
Adding createAssetInfo interface to datacatalog with taxonomy definitions 

rohithdv:
This PR adds `createAssetInfo` interface to data catalog and adds the corresponding Open API definition. It also generates the corresponding openapi client code using `openapi-generator-cli ` tool. The purpose of the `createAssetInfo` interface is to write an asset information into a catalog (like WKC, built-in catalog, etc).


pkg/connectors/datacatalog/openapiclient/docs/DefaultApi.md
```diff
@@ -0,0 +1,142 @@
+# \DefaultApi
+
+All URIs are relative to *https://localhost:8080*
+
+Method | HTTP request | Description
+------------- | ------------- | -------------
+[**GetAssetInfo**](DefaultApi.md#GetAssetInfo) | **Post** /getAssetInfo | This REST API gets data asset information from the data catalog configured in fybrik for the data sets indicated in FybrikApplication yaml
+[**WriteAssetInfo**](DefaultApi.md#WriteAssetInfo) | **Post** /createAssetInfo | This REST API write data asset information to the data catalog configured in fybrik
```
shlomitk1, line 8: write -> writes

pkg/connectors/datacatalog/clients/datacatalog.go
```diff
@@ -17,6 +17,9 @@ import (
 type DataCatalog interface {
 	GetAssetInfo(in *datacatalog.GetAssetRequest, creds string) (*datacatalog.GetAssetResponse, error)
 	io.Closer
+
+	CreateAssetInfo(in *datacatalog.CreateAssetRequest, creds string) (*datacatalog.CreateAssetResponse, error)
+	io.Closer
```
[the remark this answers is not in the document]
rohithdv: fixed. Please check. Thanks.

pkg/connectors/datacatalog/openapiclient/docs/DefaultApi.md
rohithdv: fixed. Please check. Thanks.

connectors/api/Makefile
```diff
@@ -0,0 +1,18 @@
+ROOT_DIR := ../..
```
shlomitk1, line 1: Why is this Makefile required? I don't see any usage for "generate-client-datacatalog" target. Besides, the open-api client for datacatalog has already been generated in the master version without using this rule, so what is changed now?

rohithdv: yes I have just added this temporarily just to show how the openapi client was re-generated. The change to the open api client is that there is additional code generated in api_default.go related to createAssetInfo. I will remove it anyways. Thanks.

shlomitk1: I suggest you document this and not in the Makefile, for example in README or perhaps there is a more appropriate place.

rohithdv: Ok let me think about this. Please suggest if there is a good candidate for this. It may help in future also if someone needs to add a new api.

connectors/api/datacatalog.spec.yaml
```diff
@@ -31,5 +31,33 @@ paths:
...
+            required: true
+        requestBody:
+          description: Write Asset Request
+          required: true
+          content:
+            application/json:
+              schema:
+                $ref: "../../charts/fybrik/files/taxonomy/datacatalog.json#/definitions/CreateAssetRequest"
+        responses:
+          '200':
```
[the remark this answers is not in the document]
rohithdv: Fixed. Please check. Thanks.

pkg/connectors/datacatalog/openapiclient/docs/DefaultApi.md
```diff
@@ -0,0 +1,142 @@
+# \DefaultApi
+
+All URIs are relative to *https://localhost:8080*
+
+Method | HTTP request | Description
+------------- | ------------- | -------------
+[**GetAssetInfo**](DefaultApi.md#GetAssetInfo) | **Post** /getAssetInfo | This REST API gets data asset information from the data catalog configured in fybrik for the data sets indicated in FybrikApplication yaml
```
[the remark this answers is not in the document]
rohithdv: We decided to move to POST as otherwise we would need to pass credentials through query parameters.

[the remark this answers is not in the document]
rohithdv: Please suggest if this approach is ok.

manager/controllers/mockup/datacatalog.go
```diff
@@ -45,6 +45,11 @@ func (d *DataCatalogDummy) GetAssetInfo(in *datacatalog.GetAssetRequest, creds s
 	return nil, errors.New(app.InvalidAssetID)
 }
 
+func (d *DataCatalogDummy) CreateAssetInfo(in *datacatalog.CreateAssetRequest, creds string) (*datacatalog.CreateAssetResponse, error) {
+	// will be provide a proper implementation once the implementation of CreateAssetInfo in katalog-connector is completed in a future PR. Till then a dummy implementation is provided.
```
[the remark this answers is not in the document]
rohithdv: fixed. thanks.

connectors/api/Makefile
rohithdv: I think for now I will not include the Makefile details as the creatAssetInfo method is still under development. Once we are able to complete the end to end testing (which may take about 2-3 PRs more) then we can decide a way to document this. Hope this is ok. Thanks.

pkg/connectors/datacatalog/clients/datacatalog_openapi.go
```diff
@@ -44,11 +44,19 @@ func NewOpenAPIDataCatalog(name string, connectionURL string, connectionTimeout
...
 		return nil, errors.Wrap(err, fmt.Sprintf("get asset info from %s failed", m.name))
 	}
-	return &resp, nil
+	return resp, nil
+}
+
+func (m *openAPIDataCatalog) CreateAssetInfo(in *datacatalog.CreateAssetRequest, creds string) (*datacatalog.CreateAssetResponse, error) {
+	resp, _, err := m.client.DefaultApi.WriteAssetInfo(context.Background()).XRequestDatacatalogWriteCred(creds).CreateAssetRequest(*in).Execute()
+	if err != nil {
+		return nil, errors.Wrap(err, fmt.Sprintf("get asset info from %s failed", m.name))
```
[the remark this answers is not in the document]
rohithdv: fixed. Please check. Thanks.

charts/fybrik/files/taxonomy/datacatalog.json
```diff
@@ -1,6 +1,41 @@
...
   "definitions": {
+    "CreateAssetRequest": {
+      "type": "object",
+      "required": [
+        "destinationCatalogID",
+        "resourceMetadata",
+        "details",
+        "credentials"
+      ],
+      "properties": {
```
shlomitk1, line 12: Add an optional property destinationAssetID. Katalog connector will use this field to generate a new asset (instead of test-assetID)

rohithdv: There is a [assetID](https://github.com/rohithdv/fybrik/blob/7df26117650f25076205af6dbb6b4e7968341a15/charts/fybrik/files/taxonomy/datacatalog.json#L31)  field in CreateAssetResponse object. Does it not fulfil this purpose?

shlomitk1: No, it is in the response. The idea is to request a specific asset ID.

rohithdv: Sorry I didnt understand. In `createAssetInfo` why would we be requesting a specific asset ? it is the job of `getAssetInfo` right ? Maybe I am missing something here.

shlomitk1: What assetID would the registration functionality provide? There are two options: either there is no control over the generated asset ID, or some meaningful name (or prefix) can be used.  For katalog-connector for example, you could use the original asset ID as a prefix for the new asset ID, thus after ingesting paysim-csv, we could get a new asset called paysim-csv-1234. @simanadler @roytman - what do you think?

pkg/connectors/datacatalog/openapiclient/docs/DefaultApi.md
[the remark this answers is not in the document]
roytman: I would change it to `GET`, implementation of a get operation over `POST` is weird. You can pass credentials via an HTTP header.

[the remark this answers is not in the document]
rohithdv: Can I take this up in a separate PR as this is related to `getAssetInfo` API method ?

connectors/api/datacatalog.spec.yaml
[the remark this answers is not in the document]
roytman: sorry @rohithdv , I don't see the fix

pkg/connectors/datacatalog/openapiclient/docs/DefaultApi.md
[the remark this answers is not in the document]
roytman: sure, no problem

charts/fybrik/files/taxonomy/datacatalog.json
rohithdv: @shlomitk1 I have added an optional `destinationAssetID` in the interface. Please check. Thanks.

test/services/pilot/create-asset-client/client.go
```diff
@@ -0,0 +1,49 @@
...
+		fmt.Print(err.Error())
+	}
+	req.Header.Add("Accept", "application/json")
+	req.Header.Add("Content-Type", "application/json")
+	resp, err := client.Do(req)
+	if err != nil {
+		fmt.Print(err.Error())
+	}
+	defer resp.Body.Close()
+	bodyBytes, err := ioutil.ReadAll(resp.Body)
```
shlomitk1, line 38: I believe io.ReadAll should be used according to lint (@Raviv-S ?)

test/services/pilot/create-asset-client/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
+# Copyright 2020 IBM Corp.
+# SPDX-License-Identifier: Apache-2.0
+
+resources:
+- create_asset_client.yaml
+namespace: fybrik-system
+
+images:
+- name: create-asset-clt
+  newName: ghcr.io/rohithdv/create-asset-client
```
shlomitk1, line 10: Should not point to a personal account

test/services/pilot/create-asset-client/client.go
Raviv-S: [Yes, ioutil.ReadAll is deprecated](https://pkg.go.dev/io/ioutil#ReadAll`)

shlomitk1 (commented):
I believe test/services/pilot is deprecated and should be removed from the repository.
The right place for the test is in connectors/katalog/pkg/connector/connector_test.go
( See https://github.com/fybrik/fybrik/blob/ed0e96e0d734048a3cdde336e2bc7a8f9a21af02/connectors/katalog/pkg/connector/connector_test.go#L28 for the reference)

rohithdv:
> I believe test/services/pilot is deprecated and should be removed from the repository. The right place for the test is in connectors/katalog/pkg/connector/connector_test.go See for the reference:
> 
> https://github.com/fybrik/fybrik/blob/ed0e96e0d734048a3cdde336e2bc7a8f9a21af02/connectors/katalog/pkg/connector/connector_test.go#L28

Actually the client test code (with all the  relevant kubernetes deployment files) I have added in `test/services/pilot` is not specific to `katalog connector.` It will be used to test any connector which implements the `createAssetInfo` interface (`wkc connector, katalog-connector`, etc). Basically this client code will be deployed as a kubernetes job which will send HTTP requests to the relevant connector which implements the `createAssetInfo` interface. I think it is better to keep such code in a common folder like `test/services`. However as you said any unit test related code regarding the `katalog connector` will go to  `connectors/katalog/pkg/connector/connector_test.go` . Please let me know your views on this. I can modify accordingly. Thanks. @shlomitk1 .

test/services/pilot/create-asset-client/kustomization.yaml
rohithdv: fixed. Added a placeholder for the github username. Please review. Thanks.

test/services/pilot/create-asset-client/client.go
rohithdv: Fixed. Thanks for pointing out. Now using io.readAll instead of ioutil.readAll. Please check.

connectors/api/datacatalog.spec.yaml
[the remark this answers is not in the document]
rohithdv: Have fixed it now. Please check. Thanks.

test/services/pilot/create-asset-client/client.go
```diff
@@ -0,0 +1,103 @@
...
+			Columns: []datacatalog.ResourceColumn{
+				{
+					Name: "c1",
+					Tags: &taxonomy.Tags{Properties: serde.Properties{Items: map[string]interface{}{
+						"PII": true,
+					}}},
+				},
+			},
+		},
+		// ResourceMetadata: datacatalog.ResourceMetadata{
```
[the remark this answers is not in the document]
rohithdv: Fixed. Thanks.

test/services/pilot/create-asset-client/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal
+
+ENV HOME=/tmp
+WORKDIR /tmp
+
+COPY create-asset-client /
```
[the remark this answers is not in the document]
rohithdv: fixed. Please check. Thanks.

shlomitk1 (approved):

shlomitk1:
> > I believe test/services/pilot is deprecated and should be removed from the repository. The right place for the test is in connectors/katalog/pkg/connector/connector_test.go See for the reference:
> > https://github.com/fybrik/fybrik/blob/ed0e96e0d734048a3cdde336e2bc7a8f9a21af02/connectors/katalog/pkg/connector/connector_test.go#L28
> 
> Actually the client test code (with all the relevant kubernetes deployment files) I have added in `test/services/pilot` is not specific to `katalog connector.` It will be used to test any connector which implements the `createAssetInfo` interface (`wkc connector, katalog-connector`, etc). Basically this client code will be deployed as a kubernetes job which will send HTTP requests to the relevant connector which implements the `createAssetInfo` interface. I think it is better to keep such code in a common folder like `test/services`. However as you said any unit test related code regarding the `katalog connector` will go to `connectors/katalog/pkg/connector/connector_test.go` . Please let me know your views on this. I can modify accordingly. Thanks. @shlomitk1 .

I think that each connector should come with its own unit test. An integration test is designed for the entire control plane, thus a mock client instead of manager is not required. This makes sense when you develop a new connector and want to test it vs. some client that sends the request but in this case you will need a different setup including the catalog server which is not done here. I would vote for removing test/services/pilot.

rohithdv:
> test/services/pilot.

Removed the `test/services/pilot` folder as suggested. Please check. Thanks.

rohithdv:
> > > I believe test/services/pilot is deprecated and should be removed from the repository. The right place for the test is in connectors/katalog/pkg/connector/connector_test.go See for the reference:
> > > https://github.com/fybrik/fybrik/blob/ed0e96e0d734048a3cdde336e2bc7a8f9a21af02/connectors/katalog/pkg/connector/connector_test.go#L28
> > 
> > 
> > Actually the client test code (with all the relevant kubernetes deployment files) I have added in `test/services/pilot` is not specific to `katalog connector.` It will be used to test any connector which implements the `createAssetInfo` interface (`wkc connector, katalog-connector`, etc). Basically this client code will be deployed as a kubernetes job which will send HTTP requests to the relevant connector which implements the `createAssetInfo` interface. I think it is better to keep such code in a common folder like `test/services`. However as you said any unit test related code regarding the `katalog connector` will go to `connectors/katalog/pkg/connector/connector_test.go` . Please let me know your views on this. I can modify accordingly. Thanks. @shlomitk1 .
> 
> I think that each connector should come with its own unit test. An integration test is designed for the entire control plane, thus a mock client instead of manager is not required. This makes sense when you develop a new connector and want to test it vs. some client that sends the request but in this case you will need a different setup including the catalog server which is not done here. I would vote for removing test/services/pilot.


Removed the `test/services/pilot` folder as suggested. Please check. Thanks.

roytman (commented):
A small comment about new objects descriptions.

pkg/model/datacatalog/api.go
```diff
@@ -19,3 +19,20 @@ type GetAssetResponse struct {
 	// This value is assumed to be known to the catalog connector.
 	Credentials string `json:"credentials"`
 }
+
+// +fybrik:validation:object
+type CreateAssetRequest struct {
```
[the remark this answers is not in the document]
rohithdv: Thanks @roytman for pointing out this issue. I have added the relevant descriptions. Please check the updated documentation and let me know if it is ok. 

However as I was  checking the current documentation I figured out other 2 minor issues which are described below. 

(1) There is  a minor inconsistency with the Details field as given [here](https://github.com/fybrik/fybrik/blob/052c0b7db24eeff77e387709fd305da2d9fad86e/pkg/model/datacatalog/api.go#L17) . As you can check in the previous link, the name of the field (which is `Details`) not matching with the go struct name  (which is `ResourceDetails`).  However, this inconsistency is not there if you check the `ResourceMetadata` field [here](https://github.com/fybrik/fybrik/blob/052c0b7db24eeff77e387709fd305da2d9fad86e/pkg/model/datacatalog/api.go#L16). This inconsistency has minor implications in the generation of mkdocs documentation. If you check the documentation, the link for the `Details` field is broken. 

So I thought it is better to change the field name to `ResourceDetails` from `Details` as [is done](https://github.com/fybrik/fybrik/blob/052c0b7db24eeff77e387709fd305da2d9fad86e/pkg/model/datacatalog/api.go#L16) for `ResourceMetadata` field. This leads to minor changes in other files in fybrik related to manager. Please review. 

(2) In the CreateAssetRequest go struct, I could not understand the real role of credentials field. It could be used to send the credentials of the destination data through Kubernetes secret. Hence I have made the field optional. Please review. Thanks. 

@shlomitk1 Please review from your side as well. Thanks.

charts/fybrik/files/taxonomy/datacatalog.json
```diff
@@ -1,6 +1,48 @@
...
+        "resourceMetadata",
+        "resourceDetails"
+      ],
+      "properties": {
+        "credentials": {
+          "description": "This optional field has the vault plugin path where the destination data credentials will be stored as kubernetes secrets",
+          "type": "string"
+        },
+        "destinationAssetID": {
+          "description": "This is an optional field provided to give information about the asset id to be used for the created asset.",
```
shlomitk1, line 17: Simplify: Asset ID to be used for the created asset.

charts/fybrik/files/taxonomy/datacatalog.json
```diff
@@ -1,6 +1,48 @@
...
+          "type": "string"
+        },
+        "destinationCatalogID": {
+          "description": "This has the information about the destination catalog id that new asset that will be created with the information provided in ResourceMetadata and Details field will be stored.",
```
shlomitk1, line 21: ID of the catalog where the new asset will be registered.

charts/fybrik/files/taxonomy/datacatalog.json
```diff
@@ -1,6 +1,48 @@
...
+      }
+    },
+    "CreateAssetResponse": {
+      "type": "object",
+      "required": [
+        "assetID"
+      ],
+      "properties": {
+        "assetID": {
+          "description": "This field stores the newly created asset id based on the source asset information given in CreateAssetRequest object.",
```
shlomitk1, line 41: ID of the created asset.

charts/fybrik/files/taxonomy/datacatalog.json
roytman: Indeed, you don't have to explicitly specify that a field is optional, this info is auto-generated. 
Interesting, that there is no documentation to the `resourceDetails ` and `resourceMetadata`.
image

roytman: I agree with @shlomitk1 , you don't have to write "Tis optional field has ... ", just "vault plugin path where the destination data credentials will be stored as kubernetes secrets", see as an example K8s documentation - https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/deployment-v1/

pkg/model/datacatalog/api.go
```diff
@@ -14,8 +14,30 @@ type GetAssetRequest struct {
...
+	// This has the information about the destination catalog id that new asset that will be created with the information provided in ResourceMetadata and Details field will be stored.
+	DestinationCatalogID string `json:"destinationCatalogID"`
+	// +kubebuilder:validation:Optional
+	// This is an optional field provided to give information about the asset id to be used for the created asset.
+	DestinationAssetID string `json:"destinationAssetID,omitempty"`
+	// This field has the information about the source asset metadata like asset name, owner, geography, etc
+	ResourceMetadata ResourceMetadata `json:"resourceMetadata"`
+	// This field has more details about the source asset like connection and dataformat
```
roytman, line 32: dataformat -> data format

charts/fybrik/files/taxonomy/datacatalog.json
rohithdv: Fixed. Please check. Thanks.

pkg/model/datacatalog/api.go
rohithdv: Fixed. Please check. Thanks.

charts/fybrik/files/taxonomy/datacatalog.json
rohithdv: Fixed. Please check. Thanks.

charts/fybrik/files/taxonomy/datacatalog.json
rohithdv: Fixed. Please check. Thanks.

rohithdv: Regarding documentation for `resourceDetails` and `resourceMetadata` objects, I think this issue is due to the openapigenerator tool not parsing the `description` field for non-primitive objects (which are defined by using `$ref ` field in the json schema). However, there is a description shown once we click the hyperlinks corresponding to `resourceDetails` and `resourceMetadata` in the above table. 
To resolve this issue fully, we may need deeper investigation on this and it may require more effort in order to understand the tool better. So,  I am not resolving this issue for now. Thanks.

roytman (approved):

shlomitk1 (approved):

Issue #1280, opened 2022-02-08
Customized Fybrik deployment


roytman:
If fybrik user wants to use his services as a replacement to Fybrik components, we have to allow him customization of Fybrik deployment.
 The candidates to be an optional components are:

-  Katalog
- opa connector
- opa server
- vault server

revit13:
@roytman I think chart deployment of fybrik has the option to choose other components if desired, for example, https://github.com/fybrik/fybrik/blob/master/charts/fybrik/values.yaml#L66

roytman:
The mentioned above mechanism allows to deploy or not katalog, OPA connector and OPA server. It was validated by @rohithdv .

If a user doesn't use Vault and his modules do not use it either, we might allow to deployment Fybrik without Vault.

PR #1309, opened 2022-02-20
Add the enabled flag for vault

roytman:
Some deployments do not use Vault, so this PR allows skipping the vault configuration setting.

Part of #1280 

### DIFF
--- a/charts/fybrik/values.yaml
+++ b/charts/fybrik/values.yaml
@@ -79,7 +79,9 @@ coordinator:
 
   # Configure the vault instance to be used by the coordinator manager
   vault:
-    # if you don't plan to use Vault, set it to false. If it is false, you can remove all other entries below
+    # WARNING: it's an advance feature, set it to "false", if you all your modules and connectors do not require getting
+    # credentials to from Fybrik.
+    # If it is "false", you can remove all other entries below.
     enabled: true
     # Set to the Vault address.
     address: "http://vault.{{ .Release.Namespace }}:8200"
--- a/charts/fybrik/values.yaml
+++ b/charts/fybrik/values.yaml
@@ -79,7 +79,9 @@ coordinator:
 
   # Configure the vault instance to be used by the coordinator manager
   vault:
-    # Set to the Vault address. 
+    # if you don't plan to use Vault, set it to false. If it is false, you can remove all other entries below
+    enabled: true
+    # Set to the Vault address.
     address: "http://vault.{{ .Release.Namespace }}:8200"
     # Login method to Vault
     login:
--- a/charts/fybrik/values.yaml
+++ b/charts/fybrik/values.yaml
@@ -79,8 +79,8 @@ coordinator:
 
   # Configure the vault instance to be used by the coordinator manager
   vault:
-    # WARNING: it's an advance feature, set it to "false", if you all your modules and connectors do not require getting
-    # credentials to from Fybrik.
+    # WARNING: it's an advanced feature, set it to "false" if all your modules and connectors do not require getting
+    # credentials from Fybrik.
     # If it is "false", you can remove all other entries below.
     enabled: true
     # Set to the Vault address.
--- a/manager/controllers/app/policy_compiler_interface.go
+++ b/manager/controllers/app/policy_compiler_interface.go
@@ -77,7 +77,11 @@ func LookupPolicyDecisions(datasetID string, policyManager connectors.PolicyMana
 
 	var creds string
 	if appContext.Application.Spec.SecretRef != "" {
-		creds = utils.GetVaultAddress() + vault.PathForReadingKubeSecret(appContext.Application.Namespace, appContext.Application.Spec.SecretRef)
+		if !utils.IsVaultEnabled() {
+			appContext.Log.Error().Str("SecretRef", appContext.Application.Spec.SecretRef).Msg("SecretRef defined [%s], but vault is disabled")
+		} else {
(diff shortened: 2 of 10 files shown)


charts/fybrik/values.yaml
```diff
@@ -79,7 +79,9 @@ coordinator:
 
   # Configure the vault instance to be used by the coordinator manager
   vault:
-    # Set to the Vault address. 
+    # if you don't plan to use Vault, set it to false. If it is false, you can remove all other entries below
+    enabled: false
```
shlomitk1, line 83: The default should be true. All open-source modules require credentials.

roytman: of course, my mistake, fixed it.

revit13 (approved):

manager/controllers/utils/configuration.go
```diff
@@ -16,6 +16,7 @@ import (
 // Attributes that are defined in a config map or the runtime environment
 const (
 	CatalogConnectorServiceAddressKey string = "CATALOG_CONNECTOR_URL"
+	VaultEnabled                      string = "VAULT_ENABLED"
```
revit13, line 19: All other const have `key` suffix...

roytman: what do you mean?

revit13: `CatalogConnectorServiceAddressKey`, `VaultAddressKey` ...

roytman:
Any ideas on how to add unittests ? Do we have any modules without vault?

manager/controllers/app/policy_compiler_interface.go
```diff
@@ -75,7 +75,11 @@ func LookupPolicyDecisions(datasetID string, policyManager connectors.PolicyMana
 
 	var creds string
 	if appContext.Application.Spec.SecretRef != "" {
-		creds = utils.GetVaultAddress() + vault.PathForReadingKubeSecret(appContext.Application.Namespace, appContext.Application.Spec.SecretRef)
+		if !utils.IsVaultEnabled() {
+			appContext.Log.Error().Str("SecretRef", appContext.Application.Spec.SecretRef).Msg("SecretRef defined [%s], but vault is disabled")
+		} else {
+			creds = utils.GetVaultAddress() + vault.PathForReadingKubeSecret(appContext.Application.Namespace, appContext.Application.Spec.SecretRef)
+		}
```
revit13, line 82: There is the assumption that if vault is disabled then SecretRef in fybrikapp should not be used. it makes sense to me but not sure if that would always be the case. for example, if the datacatalog could retrieve the creds from the secret without vault.

shlomitk1: Right. Instead of an error, you can pass namespace/name of the secret.

shlomitk1:
> Any ideas on how to add unittests ? Do we have any modules without vault?

All our modules require credentials for accessing data. The mechanism that was implemented for accessing a secret with credentials (possibly from a remote cluster) uses vault plugins.
What is the use-case? Data that can be accessed without credentials? Data module that retrieves credentials directly from the catalog? I don't see any tests applicable here.

charts/fybrik/values.yaml
```diff
@@ -79,7 +79,9 @@ coordinator:
 
   # Configure the vault instance to be used by the coordinator manager
   vault:
-    # Set to the Vault address. 
+    # if you don't plan to use Vault, set it to false. If it is false, you can remove all other entries below
```
shlomitk1, line 82: I would replace this comment with a strong warning explaining that this could be done only if all deployed modules do not require getting credentials to the data from Fybrik.

roytman: done

manager/controllers/app/policy_compiler_interface.go
roytman: I don't think it is a good idea to return `namespace/name `, at least now. 
The `creds` string is used as an URL to send grpc or OpenAPI requests. For K8s secret base certificates, we need to create an additional pass. Let's see if we need it.
Thoughts ?

shlomitk1: Ok, let's stay with an error.

charts/fybrik/values.yaml
```diff
@@ -79,7 +79,11 @@ coordinator:
 
   # Configure the vault instance to be used by the coordinator manager
   vault:
-    # Set to the Vault address. 
+    # WARNING: it's an advance feature, set it to "false", if you all your modules and connectors do not require getting
```
shlomitk1, line 82: advance->advanced
remove you + no comma before if ( set it to "false" only if all your modules...)

roytman: you're right, fixed, thanks

charts/fybrik/values.yaml
```diff
@@ -79,8 +79,8 @@ coordinator:
 
   # Configure the vault instance to be used by the coordinator manager
   vault:
-    # WARNING: it's an advance feature, set it to "false", if you all your modules and connectors do not require getting
-    # credentials to from Fybrik.
+    # WARNING: it's an advanced feature, set it to "false" if all your modules and connectors do not require getting
```
shlomitk1, line 82: Actually, I think the proper grammar is "if none of your modules and connectors requires..."

shlomitk1 (approved):
See a comment about the warning

charts/fybrik/values.yaml
roytman: as I know, both are correct

PR #1311, opened 2022-02-21
Hardening pod security

### LINKED ISSUE #1310: Hardening pods security
In order to harden pods security, the `SecurityContext` and `PodSecurityContext` should be defined.
Currently, they are empty.

### LINKED ISSUE #1482: Misconfigured Security Context for a Pod or Container
@roytman

## Description
A security context defines privilege and access control settings for a Pod or Container. Security context settings includes:

**- Privileged** - To run privileged or unprivileged containers. Defaults to false.
**- RunasUser** - The uid to run the entry point of container process. If unspecified, defaults to user specified in the image metadata. This should be set to value greater than 10000.
**- RunAsNonRoot** - To ensure your container should not run as root. If unspecified or false, defaults to false, kubelet will not validate the image at runtime. This should be set to True.
**- readOnlyRootFilesystem** - To ensure containers root filesystem is read-only. If unspecified, defaults to false. This should be set to True.
**- AllowPrivilegeEscalation** - To ensure a process can’t gain more privileges that its parent process. If unspecified, defaults to true. This should be set to False
**- Linux Capabilities** - To ensure process get least privileges not all privileges of the root user. Defaults to the default set of capabilities granted by the container runtime. Drop All capabilities first and add the required ones.
**- Seccomp Profile** - To filter processes system calls. If unspecified, no seccomp profile will be applied by default. A seccomp profile should be enabled by specifying it.
**- Apparmor/SeLinux** - To enable mandatory access controls. If unspecified, the container runtime will allocate a random SeLinux context for each container.

**Please see below link for the comprehensive list:**

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#securitycontext-v1-core

During pentest, below security context fields misconfiguration were observed :

1. readOnlyRootFilesystem - It was set to nil/false for all the pods

2. Missing Drop-ALL option under securityContext - **fybrik-operator-manager**

5. RunasUser - all containers are running with UID value below 10000 

6. Apparmor/SeLinux - Not set for all pods

## Please refer below screenshot:

截屏2022-05-13 15 49 09

## Affected Container/Pod(s)
```
fybrik-manager 
fybrik-operator-manager 
```

## Risk Assessment
Setting Security Context has the greatest impact on reducing the attack surface.

## Fix Recommendation
Setallsecuritycontext fieldsforthecontainer/podtoreducetheattacksurface. 
References:
https://kubernetes.io/docs/tasks/configure-pod-container/security-context/

### LINKED ISSUE #1489: Missing Seccomp Profile
@roytman

## Description
Seccomp (secure computing mode) is used to restrict the set of system calls applications can make, allowing cluster administrators greater control over the security of workloads.
It was observed that seccomp profile was not set inside containers or disabled.

## Reference:
https://blog.aquasec.com/new-docker-security-features-and-what-they-mean-seccomp-profiles

## Below is the related screenshot:

截屏2022-05-13 17 17 11

## Affected Container/Pod(s)
```
fybrik-manager 
fybrik-operator-manager 
```

## Risk Assessment
Ignoring seccomp would allow an attacker to make system calls that might compromise the security of the containers.
   
## Fix Recommendation
Seccomp support is achieved via two annotations in the pod configuration:

- seccomp.security.alpha.kubernetes.io/pod: profile applies to all containers in the pod that do not override
- container.seccomp.security.alpha.kubernetes.io/<container_name>: container -specific profile override

**Reference:**
https://kubernetes.io/docs/tutorials/security/seccomp/ 
https://itnext.io/hardening-docker-and-kubernetes-with-seccomp-a88b1b4e2111

roytman:
Add `SecurityContext` and `PodSecurityContext` 

closes #1310 
closes #1482
closes #1489

Signed-off-by: Alexey Roytman [roytman@il.ibm.com](mailto:roytman@il.ibm.com)

Co-authored-by: Revital Sur [eres@il.ibm.com](mailto:eres@il.ibm.com)
Co-authored-by:  Alexey Roytman [roytman@il.ibm.com](mailto:roytman@il.ibm.com)


rohithdv (commented):
Just a query. Is there a way to provide default security context for any pod started from values.yaml? We can override permissions for specific pods if needed under the specific pod configuration. This may avoid duplication.

laurieaw63:
Is there a reason why we want this in the values.yaml as opposed to adding helper method and using the defined helpers on pods?  That way we know what we are getting and not leaving it up the user.

roytman:
> Just a query. Is there a way to provide default security context for any pod started from values.yaml? We can override permissions for specific pods if needed under the specific pod configuration. This may avoid duplication.

> Is there a reason why we want this in the values.yaml as opposed to adding helper method and using the defined helpers on pods? That way we know what we are getting and not leaving it up the user.

Hi @rohithdv , @laurieaw63 , thanks for your comments. I'll change the PR to draft it is not finished yet. 
In general, I wanted to check how we can provide better security for our deployments. It is required by different attestation processes. We don't want to hide the security content setting from users, because these settings are for protecting the users. A user with relevant permissions always can change deployment settings. My current problem here is that OpenShift uses different user/group ranges. Probbaly we will have to add the `vault.global.openshift=true`  flag similar to vault deployment. 
I will check the option of the default security content, thanks.

charts/fybrik/templates/manager-deployment.yaml
```diff
@@ -136,6 +138,8 @@ spec:
             defaultMode: 420
             secretName: webhook-server-cert
         {{- end }}
+        - name: tmp
+          emptyDir: {}
```
shlomitk1, line 142: How is it going to work? An empty volume is mapped to /tmp while another non-empty volume is mapped to /tmp/admiconfig? Won't there be a collision?

shlomitk1: https://github.com/fybrik/fybrik/blob/304ac036a2b9f8ac3beea2b2a1dbd48d742f09db/manager/Dockerfile#L4-L8
According to this, /tmp is the root directory of the manager pod. On the other hand, we want to be able to create files in /tmp. I'd say it's going to be a problem to mark /tmp as read-only.

roytman: No, the settings above mean that `/tmp` is the home directory for the manager user and it is the running directory.
We don't want to mark `/tmp` as read-only, opposite we want to set all filesystem except `/tmp` as read-only.

shlomitk1: OK. Still, to avoid any collisions now and in the future, let's replace /tmp with /data (including policies and taxonomy), and allocate /tmp as a new empty volume.

revit13: currently, the path to policies and taxonomy is hard-coded to `/tmp/adminconfig` and `/tmp/taxonomy`.  Does it make sense to insert the path to helm values with the default prefix in `data`?

roytman: yes, please do it, and defined a Pod environment variable with the value. So we will be able to define them in a single place

shlomitk1: Why would we want to add an unnecessary configuration? There is a chance that a "wrong" value will cause problems in the deployment, which will be very hard to debug.

roytman: @shlomitk1 I don't agree with you about hardcoded configurations. Why do you think that env variables create problems for deployments?

shlomitk1: > @shlomitk1 I don't agree with you about hardcoded configurations. Why do you think that env variables create problems for deployments?

Env variables don't. Extra helm values can present a problem. We are trying to avoid /tmp being used for "data" files, so why let anyone to change the default prefix to something we don't want? 
It's fine to let users configure in which namespace they want to run modules, which connectors to deploy, etc., but the file structure of the pod is internal and I don't see a reason to let anyone change it.

shlomitk1 (approved):

charts/fybrik/values.yaml
```diff
@@ -160,16 +180,16 @@ manager:
...
-    #   drop:
-    #   - ALL
-    # readOnlyRootFilesystem: true
-    # runAsNonRoot: true
-    # runAsUser: 1000
+  # Container Security Context. If set, the fields of containerSecurityContext override
+  # the equivalent fields of .Values.global.containerSecurityContext.
+  # ref: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#security-context-1
+  containerSecurityContext:
+    readOnlyRootFilesystem: false
```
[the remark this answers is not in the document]
roytman: @laurieaw63, currently manager writes into `/tmp` directory, e.g. during helm installation. 
We are planning to add a separate volume for `/tmp` and define `readOnlyRootFileSystem: true` in a separate PR

charts/fybrik/values.yaml
```diff
@@ -50,6 +50,26 @@ global:
   # zerolog verbosity level 
   # ref: https://github.com/rs/zerolog#leveled-logging
   loggingVerbosity: -1
+  # True if the target deployment is OpenShift. Otherwise false. Default value is false.
+  openshift: false
```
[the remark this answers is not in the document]
roytman: interesting, we were not aware of this option. will update accordingly

charts/fybrik/values.yaml
```diff
@@ -50,6 +50,26 @@ global:
...
+  # Pod Security Context. This is the default setting for all pods, and can be
+  # overwritten by a specific podSecurityContext settings.
+  # ref: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#security-context
+  podSecurityContext:
```
laurieaw63, line 58: Is there a specific use case you are trying to support by allowing the pod security context and container security context to be configurable? Typically these are not configurable as they define the workload’s security posture.

charts/fybrik/values.yaml
```diff
@@ -160,16 +180,16 @@ manager:
...
-    # readOnlyRootFilesystem: true
-    # runAsNonRoot: true
-    # runAsUser: 1000
+  # Container Security Context. If set, the fields of containerSecurityContext override
+  # the equivalent fields of .Values.global.containerSecurityContext.
+  # ref: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#security-context-1
+  containerSecurityContext:
+    readOnlyRootFilesystem: false
 
   resources: {}
```
laurieaw63, line 194: Typically default request and limits are defined for the workload. Setting default resource limits help to minimize the impact of memory leaks and other ways that pods and containers can impact on other components.

charts/fybrik/values.yaml
```diff
@@ -50,6 +50,26 @@ global:
...
   # ref: https://github.com/rs/zerolog#leveled-logging
   loggingVerbosity: -1
+  # True if the target deployment is OpenShift. Otherwise false. Default value is false.
+  openshift: false
+  # Pod Security Context. This is the default setting for all pods, and can be
+  # overwritten by a specific podSecurityContext settings.
+  # ref: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#security-context
+  podSecurityContext:
+    runAsNonRoot: true
+    runAsUser: 10001
```
laurieaw63, line 60: On openshift you should support running as an arbitrary user so runAsUser should not be specified.

roytman: we don't set `runAsUser` for OpenShift

charts/fybrik/values.yaml
roytman: This is an open source project and we leave fine security tuning to our users.

charts/fybrik/values.yaml
laurieaw63: But you need to set all the other security context on openshift. Just not the runAsUser.  If running on openshift the security context will not be set. Or am I missing something?
```
      {{- if not .Values.global.openshift }}
      securityContext:
        {{- mergeOverwrite (deepCopy .Values.global.podSecurityContext) .Values.manager.podSecurityContext | toYaml | nindent 8 }}
      {{- end }}
```

PR #1315, opened 2022-02-21
Plotter reports ready when a blueprint can't be created

shlomitk1:

Fixes https://github.com/fybrik/fybrik/issues/1210

When a new blueprint is created, plotter controller sets the readiness state to false. This was done after a successful creation only, which caused the error. This PR sets the readiness state to false before checking whether the creation was successful or not.

### DIFF
--- a/manager/controllers/app/plotter_controller.go
+++ b/manager/controllers/app/plotter_controller.go
@@ -419,6 +419,7 @@ func (r *PlotterReconciler) reconcile(plotter *app.Plotter) (ctrl.Result, []erro
 				blueprint.Labels[key] = val
 			}
 			err := r.ClusterManager.CreateBlueprint(cluster, blueprint)
+			isReady = false
 			if err != nil {
 				errorCollection = append(errorCollection, err)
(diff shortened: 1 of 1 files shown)


ityuhui:
I see there is a similar piece of code for `UpdateBlueprint` in the same file. For your information. It might help.

https://github.com/fybrik/fybrik/blob/6b7f89398e284284d850b2a9055e74340ac31093/manager/controllers/app/plotter_controller.go#L363-L375

shlomitk1:
> I see there is a similar piece of code for `UpdateBlueprint` in the same file. For your information. It might help.
> 
Thanks. For UpdateBlueprint, though, isReady is set to false in all relevant branches. The only problem was for CreateBlueprint where isReady was left true at the end of the loop.

revit13 (approved):

revit13:
Would it make sense to create a test for this scenario?

shlomitk1:
> Would it make sense to create a test for this scenario?

The test will require imitating an error on a blueprint creation, which we don't have in the test environment. I don't think it is worth it.

PR #1319, opened 2022-02-22
Fixed lints in some of the packages in pkg/

Raviv-S:

### DIFF
--- a/pkg/multicluster/razee/razee_cluster_manager.go
+++ b/pkg/multicluster/razee/razee_cluster_manager.go
@@ -82,7 +82,6 @@ func (r *razeeClusterManager) GetClusters() ([]multicluster.Cluster, error) {
 			r.log.Error().Err(err).Str(clusterKey, c.Name).Msg("Resource content returned is nil! Skipping cluster")
 			continue
 		}
-		scheme = runtime.NewScheme()
 		cmcm := corev1.ConfigMap{}
 		err = multicluster.Decode(resourceContent.Content, scheme, &cmcm)
 		if err != nil {
--- a/pkg/storage/storage_provisioning.go
+++ b/pkg/storage/storage_provisioning.go
@@ -25,8 +25,14 @@ import (
 )
 
 const (
-	specLiteral  = "spec"
-	localLiteral = "local"
+	specLiteral         = "spec"
+	localLiteral        = "local"
+	statusLiteral       = "status"
+	bucketKey           = "bucket"
+	secretNameKey       = "secret-name"
+	scecretNamespaceKey = "secret-namespace"
+	endpointKey         = "endpoint"
+	provisionKey        = "provision"
 )
 
 var (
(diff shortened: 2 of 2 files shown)


pkg/multicluster/local/local_cluster_manager.go
```diff
@@ -106,9 +106,9 @@ func (cm *localClusterManager) DeleteBlueprint(cluster string, namespace string,
 }
 
 // NewClusterManager creates an instance of ClusterManager for a local cluster configuration
-func NewClusterManager(client client.Client, namespace string) (multicluster.ClusterManager, error) {
+func NewClusterManager(clnt client.Client, namespace string) (multicluster.ClusterManager, error) {
```
shlomitk1, line 109: Use cl for client as we do in some other files

pkg/storage/storage_provisioning.go
```diff
@@ -97,16 +102,16 @@ func getValue(obj map[string]interface{}, path ...string) string {
 
 func equal(required *ProvisionedBucket, existing *unstructured.Unstructured) bool {
 	obj := existing.UnstructuredContent()
-	if required.Name != getValue(obj, "spec", "local", "bucket") {
+	if required.Name != getValue(obj, specLiteral, localLiteral, "bucket") {
```
shlomitk1, line 105: "spec" and "local" are replaced with constants but no complaints for "bucket"?

Raviv-S: The complaints are only if there are more than 4 instances of the string.
"bucket" has only 2 instances.

pkg/multicluster/local/local_cluster_manager.go
Raviv-S: I'll change it

pkg/storage/storage_provisioning.go
roytman: @Raviv-S even if the linter doesn't complain about `bucket ` and it has only 2 instances, I'd define a constant for it too.

pkg/multicluster/razee/razee_cluster_manager.go
```diff
@@ -70,16 +73,16 @@ func (r *razeeClusterManager) GetClusters() ([]multicluster.Cluster, error) {
...
 		}
 		// If no content for the resource was found the cluster is not part of Fybrik or not installed
 		// correctly. Fybrik should ignore those clusters and continue.
 		if resourceContent == nil {
-			r.log.Error().Err(err).Str("cluster", c.Name).Msg("Resource content returned is nil! Skipping cluster")
+			r.log.Error().Err(err).Str(clusterKey, c.Name).Msg("Resource content returned is nil! Skipping cluster")
 			continue
 		}
-		scheme := runtime.NewScheme()
+		scheme = runtime.NewScheme()
```
Raviv-S, line 85: @roytman 
I'm not sure if the re-assignment of `scheme` would  create a problem here

PR #1328, opened 2022-02-23
Issue 1281 flatzinc read write

### LINKED ISSUE #1281: Implement FlatZinc reader and writer

zivnevo:
An implementation of a FlatZincModel class, capable of writing the model as FlatZinc and reading CSP-solver results.
Fixes #1281


pkg/optimizer/bad.fzn_solution
```diff
@@ -0,0 +1 @@
+ladder_num = array1d(1..14, [14, 6, 5, 13, 7, 8, 11, 9, 10, 1, 2, 3, 4, 12);
```
shlomitk1, line 1: I would suggest to move all test files (*.fzn_solution) to a subfolder `testdata`

pkg/optimizer/flatzinc_model.go
```diff
@@ -0,0 +1,217 @@
...
+		rightBracketPos := strings.Index(value, "]")
+		if leftBracketPos == -1 || rightBracketPos == -1 || leftBracketPos > rightBracketPos {
+			return "", nil, err
+		}
+		values = strings.Split(value[leftBracketPos+1:rightBracketPos], ",")
+	}
+	return varName, values, nil
+}
+
+func (fzw *FlatZincModel) ReadSolutions(fileName string) (map[string][]string, error) {
```
shlomitk1, line 178: Seems as only the last solution is returned and not all of them.

pkg/optimizer/flatzinc_model.go
```diff
@@ -0,0 +1,217 @@
...
+	for lineNum, line := range lines {
+		line = strings.Join(strings.Fields(line), "") // remove all whitespaces
+		switch {
+		case len(line) == 0:
+			continue // empty line
+		case strings.HasPrefix(line, "%%%"):
+			continue // stat lines are ignored
+		case line == strings.Repeat("=", 10):
+			return res, nil // a solution was found and the whole search space was covered
+		case strings.HasPrefix(line, "===="):
```
shlomitk1, line 196: Does this mean UNSAT?  Wouldn't it be clearer to check for `UNSATISFIABLE` substring?

pkg/optimizer/flatzinc_model.go
```diff
@@ -0,0 +1,217 @@
...
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"path"
+	"strings"
+)
+
+type FlatZincParam struct {
```
simanadler, line 14: General comment - please provide documentation for all structures and functions, and please add a link to FlatZinc

pkg/optimizer/flatzinc_model.go
```diff
@@ -0,0 +1,217 @@
...
+func (cnstr FlatZincConstraint) ConstraintStatement() string {
+	exprs := strings.Join(cnstr.Expressions, ", ")
+	return fmt.Sprintf("constraint %s(%s)%s;\n", cnstr.Identifier, exprs, cnstr.AnnotationString())
+}
+
+type SolveGoal int64
+
+const (
+	Satisfy SolveGoal = iota
+	Minimize
```
simanadler, line 68: I believe that the config policy code has constants for minimize and maximize.  Should be put somewhere where it can be used by both the config policies and the optimizer.

pkg/optimizer/flatzinc_model.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+func (fzw *FlatZincModel) SetSolveTarget(goal SolveGoal, expr string, annotations ...string) {
+	fzw.SolveTarget = FlatZincSolveItem{goal, expr, annotations}
+}
+
+func (fzw *FlatZincModel) Dump(fileName string) error {
+	file, err := os.Create(path.Clean(fileName))
+	if err != nil {
+		return fmt.Errorf("failed opening file %s for writing", fileName)
```
simanadler, line 128: Please use standardized fybrik logging approach.  https://fybrik.io/dev/contribute/logging/.  There are helper functions in pkg/logging

pkg/optimizer/flatzinc_model.go
```diff
@@ -0,0 +1,217 @@
...
+func (fzw *FlatZincModel) AddConstraint(identifier string, exprs []string, annotations ...string) {
+	fzw.Constraints = append(fzw.Constraints, FlatZincConstraint{identifier, exprs, annotations})
+}
+
+func (fzw *FlatZincModel) SetSolveTarget(goal SolveGoal, expr string, annotations ...string) {
+	fzw.SolveTarget = FlatZincSolveItem{goal, expr, annotations}
+}
+
+func (fzw *FlatZincModel) Dump(fileName string) error {
+	file, err := os.Create(path.Clean(fileName))
```
simanadler, line 126: Long term are we going to use a file system based approach, or will we be able to invoke the solver via API?

pkg/optimizer/flatzinc_model.go
```diff
@@ -0,0 +1,217 @@
...
+		if leftBracketPos == -1 || rightBracketPos == -1 || leftBracketPos > rightBracketPos {
+			return "", nil, err
+		}
+		values = strings.Split(value[leftBracketPos+1:rightBracketPos], ",")
+	}
+	return varName, values, nil
+}
+
+func (fzw *FlatZincModel) ReadSolutions(fileName string) (map[string][]string, error) {
+	data, err := os.ReadFile(path.Clean(fileName))
```
simanadler, line 179: Same question about file based approach

simanadler (commented):
Congratulations on your first fybrik PR!

pkg/optimizer/flatzinc_model.go
shlomitk1: @simanadler This is the returned error, the caller function can log it.

pkg/optimizer/flatzinc_model.go
shlomitk1: These are integers. Config policies use their textual representation: lines 77, 79.

pkg/optimizer/flatzinc_model.go
```diff
@@ -0,0 +1,217 @@
...
+	switch s {
+	case Satisfy:
+		return "satisfy"
+	case Minimize:
+		return "minimize"
+	case Maximize:
+		return "maximize"
+	}
+	return "unknown"
+}
```
shlomitk1, line 82: I believe we'll need the reverse function as well - given the directives received from the policies, determine the SolveGoal value.

pkg/optimizer/flatzinc_model.go
```diff
@@ -0,0 +1,217 @@
...
+	if len(annotsStr) > 0 {
+		annotsStr = " :: " + annotsStr
+	}
+	return annotsStr
+}
+
+type FlatZincVariable struct {
+	NumInstances uint // NumInstances > 1 means an array of vars
+	Name         string
+	Type         string
```
shlomitk1, line 41: This can not be a free string. It's better to define a type over the string and specify the valid values as constants including "int", "float", ...

pkg/optimizer/flatzinc_model.go
roee88: Consider using errors.Wrapf to preserve the original error details

pkg/optimizer/flatzinc_model.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+func (fzw *FlatZincModel) Dump(fileName string) error {
+	file, err := os.Create(path.Clean(fileName))
+	if err != nil {
+		return fmt.Errorf("failed opening file %s for writing", fileName)
+	}
+	defer func() {
+		if err := file.Close(); err != nil {
+			log.Printf("Error closing file: %s\n", err)
```
roytman, line 132: maybe to add the fileName to the error message. The error object might not include it.

pkg/optimizer/bad.fzn_solution
zivnevo: Done

pkg/optimizer/flatzinc_model.go
```diff
@@ -11,14 +11,23 @@ import (
 	"strings"
 )
 
+/*
+This file implements FlatZincModel: a class to hold a constraint problem, based on the FlatZinc format.
+The class can dump the a constraint program to a file, using the FlatZinc specification.
```
shlomitk1, line 16: typo: the a

pkg/optimizer/flatzinc_model.go
zivnevo: We now have two methods:
* `ReadSolutions` will return all solutions as a slice of `CPSolution`s
* `ReadBestSolution` will return only the last solution, which should be the best one

pkg/optimizer/flatzinc_model_test.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+}
+
+func TestReadingResults(t *testing.T) {
+	myReader := NewFlatZincModel()
+	res, err := myReader.ReadSolutions("test1.fzn_solution")
+	if err != nil {
+		t.Errorf("%s", err)
+	}
+	fmt.Println(res)
```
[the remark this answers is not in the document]
zivnevo: done

pkg/optimizer/flatzinc_model.go
zivnevo: There are several possible solver answers that have this `====` prefix, UNSAT being just one of them.
However, you are correct that we need to differentiate UNSAT from e.g., UNKNOWN, because an UNSAT may mean we need to retry with more modules.
I changed the code so that an UNSAT will return a slice with a single empty solution.
This way, legitimate solver outputs should always have one solution, and it's up to the caller to check that the solution is not empty.
If the result is not UNSAT and if no solution is found, we return with an error.

pkg/optimizer/flatzinc_model.go
zivnevo: done

pkg/optimizer/flatzinc_model.go
zivnevo: I believe we are going to use a file-system based approach.
There is no standard FlatZinc API; every solver defines its own. Hence, using solver API is not portable.
MiniZinc does define [a standard command-line interface](https://www.minizinc.org/doc-2.5.5/en/fzn-spec.html#command-line-interface-and-standard-options) so maybe we can have a single adapter for all solvers.

pkg/optimizer/flatzinc_model.go
zivnevo: Solvers should output their result to `stdout`, so we can read it directly from there without going through files

pkg/optimizer/flatzinc_model.go
zivnevo: Will the directives received from the policies use the same keywords?

pkg/optimizer/flatzinc_model.go
zivnevo: It's not that easy, because type can be something like `6..10` or `0.7..5.3` or `{9, 15, 22}`, which are all types we may have to use

pkg/optimizer/flatzinc_model.go
zivnevo: fixed

pkg/optimizer/flatzinc_model.go
zivnevo: done

pkg/optimizer/flatzinc_model.go
shlomitk1: This is the suggested definition (can be changed if needed):
```
type OptimizationDirective string

// List of directives
const (
	Minimize OptimizationDirective = "min"
	Maximize OptimizationDirective = "max"
)
```

pkg/optimizer/flatzinc_model.go
shlomitk1: Ok

pkg/optimizer/flatzinc_model.go
shlomitk1: Will this work as expected in case of multiple reconciles simultaneously, i.e. applying the solver in multiple processes with different inputs/outputs?

zivnevo: For now, the plan is to block-wait until the CSP solver returns an answer, hoping that it doesn't take too long.
We'll just use `exec.Command("CSPsolver").Output()`. Hence, no issue of mixing answers from different solver runs.

shlomitk1 (approved):

roytman (approved):

PR #1332, opened 2022-02-24
code to create fybrik-python-logging library

cdoron:
code to create fybrik-python-logging library
plus github workflow to publish package to PyPI

We are still waiting for a FybrikUser account.
For now, we will be publishing on the **Test** PyPI repository under the cdoron account.

Once the FybrikUser account user is created, we will:
1. create an account on PyPI for FybrikUser . Create token for this new user.
2. create a new repository secret PYPI_API_TOKEN , which will contain the token FybrikUser
3. remove `repository_url: https://test.pypi.org/legacy/` from `workflows/python-publish.yaml` (which means that we publish to regular PyPI and not Test PyPI)
4. replace password: `${{ secrets.TEST_PYPI_API_TOKEN }}` with password: `${{ secrets.PYPI_API_TOKEN }}`
5. change the trigger for the workflow, so it runs with every code merge


.github/workflows/python-publish.yaml
```diff
@@ -0,0 +1,30 @@
+name: python-publish
+on:
+  push:
+    tags:
+      - '^v[0-9]+\.[0-9]+(\.[0-9]+)*'
+    branches:
+      - master
+    paths:
+      - pkg/logging/python_logging/**
```
roee88, line 9: Are you sure that you want to publish on push to master? Or is it just for the testing phase and will be removed later?

.github/workflows/python-publish.yaml
```diff
@@ -0,0 +1,30 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install build
+    - name: Build package
+      run: RELEASE=${{ steps.version.outputs.version }} python -m build ./pkg/logging/python_logging --outdir ./dist
```
roee88, line 24: I don't see a version step

.github/workflows/python-publish.yaml
```diff
@@ -0,0 +1,30 @@
...
+    - name: Publish package
+      uses: pypa/gh-action-pypi-publish@27b31702a0e7fc50959f5ad993c78deac1bdfc29
```
roee88, line 26: Their README says `pypa/gh-action-pypi-publish@release/v1`

.github/workflows/python-publish.yaml
cdoron: will be removed later. I will update the comment accordingly.

roytman:
Why this Python package should be part of the `fybrik` repository written in goLang? 
The next steps will put here Java and Scala (`mover` is written in scala)  packages? 
And what will we do if somebody would like to implement his modules in C, C++, Ruby ...

pkg/logging/python_logging/setup.py
```diff
@@ -0,0 +1,15 @@
+from setuptools import setup
+import os
+
+setup(name='fybrik_python_logging',
+      version=os.environ.get('RELEASE', '0.1'),
+      description='Python Logging Package for Fybrik Components',
+      license='Apache License, Version 2.0',
+      author='Doron Chen',
```
revit13, line 8: Until FybrikUser account is ready consider using a fake user and email

cdoron:
> Why this Python package should be part of the `fybrik` repository written in goLang? The next steps will put here Java and Scala (`mover` is written in scala) packages? And what will we do if somebody would like to implement his modules in C, C++, Ruby ...
> 
> I think that all SDK, and this package is a starting point of Python SDK creation should be in a separate repositories.

If I understood correctly, Sima wanted this python library to be part of the fybrik repository.
Let's discuss. Thanks!

roytman:
> > Why this Python package should be part of the `fybrik` repository written in goLang? The next steps will put here Java and Scala (`mover` is written in scala) packages? And what will we do if somebody would like to implement his modules in C, C++, Ruby ...
> > I think that all SDK, and this package is a starting point of Python SDK creation should be in separate repositories.
> 
> If I understood correctly, Sima wanted this python library to be part of the fybrik repository. Let's discuss. Thanks!

On the second thought, before the creation of SDKs, let's do it in the following way. Please create a directory `python` in the root directory, and put your code into it, e.g. `python/fybrik_logging`. What do you think ? will it work for you ? Do you want to discuss ?

cdoron:
> Why this Python package should be part of the `fybrik` repository written in goLang? The next steps will put here Java and Scala (`mover` is written in scala) packages? And what will we do if somebody would like to implement his modules in C, C++, Ruby ...
> 
> I think that all SDK, and this package is a starting point of Python SDK creation, should be in separate repositories.

Following our discussion today, I placed the code in a new python/ library at the root of the repostory

.github/workflows/python-publish.yaml
```diff
@@ -0,0 +1,37 @@
...
+        pip install build
+    - name: Build package
+      run: |
+        version="${GITHUB_REF#refs/tags/v}"
+        if  [[ $version == refs/* ]] ;
+        then
+          RELEASE="0.0.0" python -m build ./pkg/logging/python_logging --outdir ./dist
+        else
+          RELEASE=$version python -m build ./pkg/logging/python_logging --outdir ./dist
+        fi
```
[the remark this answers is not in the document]
cdoron: We decided to maintain a separate versioning system from that of Fybrik

revit13:
minor comment: would it make sense to create a `logging` directory with sub-directory `python` instead of `python` dir with `logging`  as sub-directory. This way all the other languages could be placed in directories under `logging` dir.

roytman:
> minor comment: would it make sense to create a `logging` directory with sub-directory `python` instead of `python` dir with `logging` as sub-directory. This way all the other languages could be placed in directories under `logging` dir.

@revit13 , the idea is to prepare client SDKs, the directories (`python`, `java` ) will probably contain other components in addition to logging, e.g. authentication/validation .

.github/workflows/python-publish.yaml
```diff
@@ -0,0 +1,28 @@
+name: python-publish
+on:
+  workflow_dispatch:
+    inputs:
+      library-version:
+        required: true
```
roytman, line 6: as optional parameters, you can add destination repository, username, and password.
It can help to test too, what do you think?

cdoron: I added the `pypi-repository-url` and `pypi-user` as optional parameters.
It seems that secrets cannot be referenced from the `inputs` section.

roytman (approved):

PR #1352, opened 2022-03-06
Handle the case of non empty dataformat field in application interface.

Mohammad-nassar10:
Handling the case when the `spec.data.dataSetID.requirement.interface.dataformat` field in `fybrikApplication` yaml file is not empty. If the value is the same as in the `asset` then ignore it, else return an error.


shlomitk1:
What is the problem you are trying to solve? Let's say the workload is sending data in csv format to be written in an object store as a parquet file. Either there is a write module that accepts csv in api and parquet as sink, or it is not. If there is, all is good - the module did all the proper conversions. If not, user gets an error. Why isn't it enough?

manager/controllers/app/select_modules.go
```diff
@@ -85,9 +85,20 @@ func (p *PlotterGenerator) FindPaths(item *DataInfo, appContext *app.FybrikAppli
...
+	// for write flow the dataformat in the application is not needed
+	WriteNodeFromAppRequirements := Node{
+		Connection: &taxonomy.Interface{
+			Protocol: item.Context.Requirements.Interface.Protocol,
+		},
+	}
 
 	if item.Context.Flow == taxonomy.WriteFlow {
-		source = NodeFromAppRequirements
+		if item.Context.Requirements.Interface.DataFormat != "" && item.Context.Requirements.Interface.DataFormat != item.DataDetails.Details.DataFormat {
```
simanadler, line 96: Not sure this should result in an error.  There may be modules available that can transform between the formats.  If so, they should be included in the data path.  If there are no such modules available then an error should be returned

simanadler (commented):
See comment

shlomitk1:
I believe this PR is no longer required.

PR #1355, opened 2022-03-07
Fixed more lints on /manager and /pkg

Raviv-S:


pkg/connectors/policymanager/clients/policymanager_grpc.go
```diff
@@ -24,6 +24,19 @@ import (
 
 var _ PolicyManager = (*grpcPolicyManager)(nil)
 
+const (
```
shlomitk1, line 27: This file is going to be deleted in the nearest future, so you can just add some general nolint comment instead of fixing lint issues.

roytman (approved):

Issue #1331, opened 2022-02-24
Validation of infrastructure attributes


shlomitk1:
- [ ] Define in the taxonomy which units are relevant for a specific attribute, e.g. MB/s for bandwidth, US Dollar for storage-cost.
- [ ] Add tests for valid cases and invalid (invalid attribute, invalid unit, invalid match, invalid structure)

PR #1360, opened 2022-03-09
Add test for validation of infrastructure attributes

YingchaoDu:
Fixes issue #1331 


shlomitk1 (approved):

pkg/infrastructure/attribute_manager_test.go
```diff
@@ -0,0 +1,80 @@
...
+func TestInvalidInfrastructureAttributeBW(t *testing.T) {
+	t.Parallel()
+
+	data := `{
+		"infrastructure":[{
+			"attribute": "bandwidth",
+			"description": "neverland object store",
+			"value": "100",
+			"type": "numeric",
+			"units": "GB",
```
roytman, line 74: @shlomitk1 , maybe this is unrelated to the PR, but do we have or plan to implement units transformation/comparison. I mean if we have 2 infrastructure entries one with 1GB and another 1000 (1024) MB, how they will be compared?

shlomitk1: We have but for the future. The first version of optimizer will assume one type of units per attribute.

shlomitk1 (approved):

PR #1362, opened 2022-03-13
Add strategy to run tests with the "old" and "new" K8s versions

roytman:


.github/workflows/build.yml
```diff
@@ -9,9 +9,12 @@ on:
   pull_request:
     branches:
       - master
-
+  workflow_dispatch:
+
```
revit13, line 13: I think we need to be careful with manually triggering the workflow as I am not sure how it will effect push images stage.

.github/workflows/build.yml
```diff
@@ -86,6 +92,8 @@ jobs:
     - name: Install tools
       run: make install-tools
     - name: Integration tests
+      env:
+        K8S_VERSION: ${{ matrix.k8s_version }}
```
revit13, line 96: Cool!

.github/workflows/build.yml
```diff
@@ -121,7 +139,15 @@ jobs:
         restore-keys: ${{ runner.os }}-t-${{ hashFiles('hack/make-rules/tools.mk') }}-go
     - name: Install tools
       run: make install-tools
-    - name: Notebook tests
+    - name: Notebook tests old
```
revit13, line 142: maybe rename: `Notebook tests old k8s version`

.github/workflows/build.yml
```diff
@@ -121,7 +139,15 @@ jobs:
...
+      if: ${{ matrix.k8s_version == 'old' }}
+      env:
+        K8S_VERSION: ${{ env.OLD_K8S_VERSION }}
+      run: make run-notebook-tests
+    - name: Notebook tests new
```
revit13, line 147: maybe rename: Notebook tests k8s version 1.22

revit13:
@roytman LGTM. There is a setting of K8S_VERSION=1.19.2 in  install_kubebuilder.sh script. not sure if that should be changed too

.github/workflows/build.yml
roytman: OK, removed it

.github/workflows/build.yml
roytman: It is only a step name, doesn't the job name, and I don't want to put specific K8S versions in the action body, only in its header.
Plus, we cannot use env variables in the strategy values. ;-(
Does it make sense ?

.github/workflows/build.yml
roytman: see above

roytman:
> @roytman LGTM. There is a setting of K8S_VERSION=1.19.2 in install_kubebuilder.sh script. not sure if that should be changed too

Thanks @revit13 , let's put it in a separate commit

revit13 (approved):

PR #1363, opened 2022-03-13
fixes to write flow

shlomitk1:
Refers to https://github.com/fybrik/fybrik/issues/1227

This PR revisits the way data path is constructed regarding a write flow.
Some changes:
- support cascading
- expose the correct endpoint in fybrik app status
- governance actions applied for writing an existing asset (update)
- store credentials for various flows in the plotter


manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -494,10 +493,10 @@ func (r *FybrikApplicationReconciler) constructDataInfo(req *DataInfo, appContex
 	configEvaluatorInput.Workload.Cluster = workloadCluster
 	configEvaluatorInput.Request = CreateDataRequest(input, req.Context, &req.DataDetails.ResourceMetadata)
 
-	// Read policies for data that is processed in the workload geography
+	// Policies for data that is processed in the workload geography
 	if configEvaluatorInput.Request.Usage == taxonomy.ReadFlow {
```
Mohammad-nassar10, line 497: Is this still need to be `taxonomy.ReadFlow`?

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -303,28 +303,27 @@ func (r *FybrikApplicationReconciler) deleteExternalResources(applicationContext
...
+				endpointMap[flow.AssetID] = lastStep.Parameters.API.Connection
 			}
 		}
 	}
 	// populate endpoints in application status
 	for _, asset := range application.Spec.Data {
 		state := application.Status.AssetStates[asset.DataSetID]
-		state.Endpoint = readEndpointMap[asset.DataSetID]
+		state.Endpoint = endpointMap[asset.DataSetID]
 		application.Status.AssetStates[asset.DataSetID] = state
```
revit13, line 327: is it enough to do `application.Status.AssetStates[asset.DataSetID].Endpoint = endpointMap[asset.DataSetID]` instead of the three lines above?

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -303,28 +303,27 @@ func (r *FybrikApplicationReconciler) deleteExternalResources(applicationContext
...
+		// sanity check
+		if len(flow.SubFlows) == 0 {
+			continue
+		}
+		subflow := flow.SubFlows[len(flow.SubFlows)-1]
+		for _, sequentialSteps := range subflow.Steps {
+			// Check the last step in the sequential flow (this will expose the api)
+			lastStep := sequentialSteps[len(sequentialSteps)-1]
+			if lastStep.Parameters.API != nil {
+				endpointMap[flow.AssetID] = lastStep.Parameters.API.Connection
```
revit13, line 319: in what case the api of the last step is not nil? is it worth to emit a log message in this case? (or an error if we always expect it to be not empty)

manager/controllers/app/fybrikapplication_controller.go
shlomitk1: It's a Go restriction that prevents you from updating a field of a map.

manager/controllers/app/fybrikapplication_controller.go
shlomitk1: Not all flows expose an api, e.g. copy or delete.

manager/controllers/app/select_modules.go
```diff
@@ -116,51 +115,81 @@ func (p *PlotterGenerator) validSolutions(item *DataInfo, solutions []Solution,
...
+				item.Configuration.ConfigDecisions[moduleCapability.Capability].DeploymentRestrictions.StorageAccounts,
+				&p.StorageAccounts[accountInd].Spec,
+				p.StorageAccounts[accountInd].Name) {
+				p.Log.Debug().Str(logging.DATASETID, item.Context.DataSetID).Msgf("storage account %s does not match the requirements",
+					p.StorageAccounts[accountInd].Name)
+				continue
+			}
+			// query the policy manager whether WRITE operation is allowed
+			if !item.Context.Requirements.FlowParams.IsNewDataSet {
 				operation.Destination = string(p.StorageAccounts[accountInd].Spec.Region)
```
simanadler, line 152: I'm wondering if we should add a function in the data governance engine connector that does some of this work.  I.e. receives a dataset id, flow, and a list of locations (storage locations) and returns which are allowed.  Wouldn't that be cleaner that doing it here?

manager/testdata/unittests/write_asset.yaml
```diff
@@ -0,0 +1,25 @@
...
+        app: notebook
+  appInfo:
+    intent: Fraud Detection
+    role: Security
+  data:
+  - dataSetID: s3/redact-dataset
+    flow: write
+    requirements:
+      interface:
+        protocol: fybrik-arrow-flight
```
simanadler, line 25: protocol should be fybrik-flight (to differentiate from product flight which has different API), dataFormat should be arrow

manager/testdata/unittests/module-transform.yaml
```diff
@@ -19,9 +19,23 @@ spec:
...
+      supportedInterfaces: 
       - source:
           protocol: fybrik-arrow-flight
       actions:
         - name: RedactAction
         - name: RemoveAction
+    - capability: transform
+      api:
+        connection:
+          name: fybrik-arrow-flight
```
simanadler, line 31: fybrik-flight

manager/testdata/unittests/module-transform.yaml
```diff
@@ -19,9 +19,23 @@ spec:
...
+          fybrik-arrow-flight:
+            hostname: arrow-flight-transform
+            port: 80
+            scheme: grpc
+      supportedInterfaces: 
+      - sink:
+          protocol: fybrik-arrow-flight
```
simanadler, line 38: fybrik-flight

simanadler: Add dataFormat = arrow

manager/testdata/unittests/module-read-write.yaml
```diff
@@ -13,6 +13,9 @@ spec:
   capabilities:
     - capability: read
       scope: workload
+      actions:
+        - name: RedactAction
+        - name: RemoveAction
       api:
         connection:
           name: fybrik-arrow-flight
```
simanadler, line 21: fybrik-flight

simanadler: add dataFormat = arrow

shlomitk1: I believe "arrow" format has been deprecated as well as "table" for db2.

manager/testdata/unittests/write_asset.yaml
shlomitk1: Do you mean that the module we use in the quick-start scenario should have a different definition?
What about e2e and notebook tests? 
@simanadler Please open a separate issue to handle the formats, taxonomy changes (otherwise all the tests will fail), module definitions. I don't think this should be done in this PR.

manager/controllers/app/select_modules.go
shlomitk1: Yes, it will. The output should include a list of actions per each location, since storing in some may require to reduce a column, storing in another can be done as-is, ...

manager/controllers/app/fybrikapplication_controller.go
shlomitk1: I'm rewriting this function. However, there are two main issues here:
- no 1:1 matching from flow to action type, i.e. there is no such thing as checking if copy is allowed (only WRITE action is supported by the connectors, no COPY).
- For Copy and Write flows we can't send a request to the policy manager because we don't know the final destination, it is checked later when a storage account is determined.
At this stage we can check only Read and Delete permissions that do not depend on the data path construction.

manager/controllers/app/select_modules.go
```diff
@@ -116,51 +115,81 @@ func (p *PlotterGenerator) validSolutions(item *DataInfo, solutions []Solution,
...
+		// query the policy manager whether WRITE operation is allowed
+		operation.Destination = item.DataDetails.ResourceMetadata.Geography
+		operation.ProcessingLocation = taxonomy.ProcessingLocation(item.WorkloadCluster.Metadata.Region)
+		actions, err = LookupPolicyDecisions(item.Context.DataSetID, p.PolicyManager, appContext, operation)
+		if err != nil {
+			p.Log.Error().Err(err).Msg("could not validate governance action for writing data")
+			return false
+		}
+	} else {
+		for accountInd := range p.StorageAccounts {
```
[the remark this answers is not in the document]
shlomitk1: Added a comment. 
It is not related to format/protocol, but rather to governance and admin config policies.

manager/testdata/unittests/write_asset.yaml
simanadler: @roee88 I thought you said that this was changed in the taxonomy.  Would appreciate it if you could weigh in here.

manager/controllers/app/select_modules.go
```diff
@@ -116,51 +115,81 @@ func (p *PlotterGenerator) validSolutions(item *DataInfo, solutions []Solution,
...
-				}
-				// query the policy manager whether WRITE operation is allowed
-				operation := new(policymanager.RequestAction)
-				operation.ActionType = taxonomy.WriteFlow
+	var err error
+	var actions []taxonomy.Action
+	operation := &policymanager.RequestAction{}
+	operation.ActionType = taxonomy.WriteFlow
+
+	if item.Context.Flow == taxonomy.WriteFlow && !item.Context.Requirements.FlowParams.IsNewDataSet {
```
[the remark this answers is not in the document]
shlomitk1: No. This function is common to all cases where data should be written.
The cases are:
- Read with implicit copy
- Copy (ingest)
- Write where the destination is known (update)
- Write where the destination is unknown.
Except for the third case (update), Fybrik should select a storage account out of the existing and allocate a storage. 
The condition in line 128 checks the update flow (write to an existing dataset)

manager/controllers/app/fybrikapplication_controller_unit_test.go
```diff
@@ -1463,3 +1463,76 @@ func TestWriteRegisteredAsset(t *testing.T) {
 	g.Expect(plotter.Spec.Assets).To(gomega.HaveLen(2))
 	g.Expect(plotter.Spec.Templates).To(gomega.HaveLen(2)) // expect two templates: one for read and one for write
 }
+
+func TestWriteAndTransform(t *testing.T) {
```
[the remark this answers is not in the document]
shlomitk1: Writing a new asset is not supported yet. When we construct a data path, we look for a module / pipeline of modules that connects asset interface (e.g. s3, parquet) to the workload interface (e.g. rest, json). The application defines the letter. We are missing the former for a new asset.

manager/controllers/app/select_modules.go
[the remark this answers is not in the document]
shlomitk1: Rewritten this function. There is no point to check for governance actions on each constructed path for an update flow since the information is known beforehand.

manager/controllers/app/fybrikapplication_controller_unit_test.go
```diff
@@ -1463,3 +1464,133 @@ func TestWriteRegisteredAsset(t *testing.T) {
...
+	}
+
+	_, err := r.Reconcile(context.Background(), req)
+	g.Expect(err).To(gomega.BeNil())
+
+	err = cl.Get(context.TODO(), req.NamespacedName, application)
+	g.Expect(err).To(gomega.BeNil(), "Cannot fetch fybrikapplication")
+	// Expect Deny condition
+	cond := application.Status.AssetStates["s3/deny-dataset"].Conditions[DenyConditionIndex]
+	g.Expect(cond.Status).To(gomega.BeIdenticalTo(corev1.ConditionTrue), "Deny condition is not set")
```
revit13, line 1593: not set -> set

Mohammad-nassar10 (approved):

PR #1385, opened 2022-03-27
Resource updates

shlomitk1:
Fixes https://github.com/fybrik/fybrik/issues/1376

This PR prevents kubernetes errors on conflict when updating resources.


manager/controllers/utils/utils.go
```diff
@@ -145,3 +151,45 @@ func GetFybrikApplicationUUIDfromAnnotations(annotations map[string]string) stri
...
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		// Retrieve the latest version of the object before attempting update
+		// RetryOnConflict uses exponential backoff to avoid exhausting the apiserver
+		if err := cl.Get(ctx, client.ObjectKeyFromObject(obj), obj); err != nil {
+			return err
+		}
+		obj.SetFinalizers(finalizers)
+		return cl.Update(ctx, obj)
+	})
+}
```
roytman, line 167: you have the object, why not to use it first 
```
func UpdateFinalizers(ctx context.Context, cl client.Client, obj client.Object) error {
	finalizers := obj.GetFinalizers()
	if err := cl.Update(ctx, obj); err != nil {
		if ! errors.IsConflict(err) {
			return err
		}
		return retry.RetryOnConflict(retry.DefaultRetry, func() error {
			// Retrieve the latest version of the object before attempting update
			// RetryOnConflict uses exponential backoff to avoid exhausting the apiserver
			if err := cl.Get(ctx, client.ObjectKeyFromObject(obj), obj); err != nil {
				return err
			}
			obj.SetFinalizers(finalizers)
			return cl.Update(ctx, obj)
		})
	}
	return nil
}
```

roytman (commented):
BTW, I created concurrent updates, it works.

manager/controllers/utils/utils.go
shlomitk1: Fixed

roytman (approved):

PR #1386, opened 2022-03-27
Fix Katalog RBAC, add comments to values.yaml, remove fybrik-user-cr.…

roytman:
- Fix Katalog RBAC, now when Katalog supports `createAssets` it needs write permissions, `katalog-editor-role/cr` doesn't need access to secretes.
-  add comments to values.yaml, explaining `modulesNamespace` and `applicationNamespace` settings
-  remove charts/fybrik/templates/fybrik-user-cr.yaml, fybrik-user-cr is not used.


charts/fybrik/values.yaml
```diff
@@ -13,9 +13,10 @@ fullnameOverride: ""
 
 clusterScoped: true
 
+# If we work with "clusterScope:false", the namespace must already exist
 modulesNamespace: "fybrik-blueprints"
 
-# Set if need to only watch for FybrikApplication from a specific namespace
+# Set if needs to watch for FybrikApplication from a specific namespace only, or when we work with "clusterScope:false"
```
shlomitk1, line 19: No need to change the comment, clusterScope:false is used for watching FybrikApplications from a single namespace

charts/fybrik/values.yaml
```diff
@@ -13,9 +13,10 @@ fullnameOverride: ""
 
 clusterScoped: true
 
+# If we work with "clusterScope:false", the namespace must already exist
```
shlomitk1, line 16: I am not sure about this requirement

charts/fybrik/templates/katalog-connector/katalog-viewer-role.yaml
```diff
@@ -5,11 +5,11 @@ apiVersion: rbac.authorization.k8s.io/v1
 {{- if .Values.clusterScoped }}
```
shlomitk1, line 5: Where is this used? Does a (cluster)rolebinding exist for this role?

charts/fybrik/values.yaml
roytman: no, it works for clusterScope:true too, if it is defined, only a single namespace is watching.

charts/fybrik/values.yaml
roytman: it is required. We have started the discussion today from it.

charts/fybrik/templates/katalog-connector/katalog-viewer-role.yaml
roytman: it was, I changed it.  will remove the file at all.

charts/fybrik/templates/katalog-connector/katalog-connector-rbac.yaml
```diff
@@ -5,18 +5,19 @@ apiVersion: rbac.authorization.k8s.io/v1
 {{- if .Values.clusterScoped }}
 kind: ClusterRoleBinding
 metadata:
-  name:  katalog-connector-viewer
+  name:  fybrik-katalog-connector-crb
```
revit13, line 8: minor comment: In other places I see `{{ template "fybrik.fullname" . }}` is used, for example: https://github.com/fybrik/fybrik/blob/master/charts/fybrik/templates/fybrik-blueprints-rb.yaml#L5

revit13 (approved):

charts/fybrik/templates/katalog-connector/katalog-connector-rbac.yaml
roytman: correct, Kritika from the BU works to replace all names to be with prefix "fybrik-" only,  (operator doesn't support templates. But meantime, no problem I'll add this here too.

charts/fybrik/templates/_helpers.tpl
```diff
@@ -116,10 +116,6 @@ cert-manager.io/v1
 Get modules namespace
 */}}
 {{- define "fybrik.getModulesNamespace" -}}
-{{- if .Values.modulesNamespace -}}
-{{- .Values.modulesNamespace -}}
-{{- else -}}
-{{- "fybrik-blueprints" -}}
-{{- end -}}
+{{- .Values.modulesNamespace.name | default "fybrik-blueprints" -}}
```
shlomitk1, line 119: .Values.modulesNamespace (without .name)

charts/fybrik/templates/fybrik-blueprints-role.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
```
shlomitk1, line 3: Why was the corresponding rolebinding deleted?

roytman: Mistake, fix it, thanks

shlomitk1 (approved):

PR #1388, opened 2022-03-28
Rename createAssetInfo to createAsset

rohithdv:
This PR renames createAssetInfo method to createAsset. 


connectors/api/Makefile
```diff
@@ -0,0 +1,18 @@
+ROOT_DIR := ../..
+include $(ROOT_DIR)/Makefile.env
```
roytman, line 2: Why do you need this Makefile, there are one in the `connectors` dir, can you use it ?

rohithdv: As suggested earlier by @shlomitk1, we dont need the Makefile. I just checked in my mistake. The information in this Makefile we can put later in some readme under a section to add a new method in Fybrik. I have removed this Makefil for now. Thanks.

roytman (approved):

Issue #1403, opened 2022-04-01
Lack of information about usage fybrik logging package 


roytman:
Fybrik has a logger wrapper and there is a [page ](https://fybrik.io/v0.6/contribute/logging/)describeds it, or at least motivation to do it. 
However, there is no information on how to use logger API, for example when and how to use `LogStructure`.

Another issue with `LogStructure` is that it prints log messages with `caller` pointed to itself L:133, and hides a real caller.  

Closes #1403

roytman:
closes https://github.com/fybrik/fybrik/issues/1403

PR #1404, opened 2022-04-03
Addresses logger bug #1403

simanadler:
LogStructure now indicates the correct caller (the function that called LogStructure) instead of itself.  Also added documentation about the LogStructure function in the website page about logging

Example log entry generated by LogStructure:
DBG plotter_generator.go:85 > ProvisionedStorage element: {
	"Storage": {
		"Name": "notebook-default8f06da7fbb",
		"Endpoint": "http://s3.eu.cloud-object-storage.appdomain.cloud",
		"SecretRef": {
			"Namespace": "fybrik-system",
			"Name": "credentials-theshire"
		}
	}


pkg/logging/logging.go
```diff
@@ -130,6 +130,7 @@ func LogStructure(argName string, argStruct interface{}, log *zerolog.Logger, fo
 	if err != nil {
 		log.Warn().Msg("Failed converting " + argName + " to json")
 	} else {
-		log.Debug().Bool(FORUSER, forUser).Bool(AUDIT, audit).Msg(argName + ": " + string(jsonStruct))
+		// Log the info making sure that the calling function is listed as the caller
+		log.Debug().CallerSkipFrame(3).Bool(FORUSER, forUser).Bool(AUDIT, audit).Msg(argName + ": " + string(jsonStruct))
```
roytman: - Please fix the golint error.
- why do you skip 3 frames, it provides the wrong output, you should skip a single frame (the `LogStructure` function itself"

site/docs/contribute/logging.md
```diff
@@ -63,3 +63,5 @@ The fields in this section are typically generated by the logging libraries.
 - LOGGING_VERBOSITY - should be set to one of the levels described in the previous section.  
 - PRETTY_LOGGING - If true log entries are in human readable format.  If false, they are in json. Should only be true during individual development, since json is preferred to enable easy parsing by aggregator tools.
 
+## Logging of Structures
+Fybrik provides a helper function called `LogStructure` in pkg/logging/logging.go for writing Go structures in json format to the log.
```
roytman, line 67: My point about documentation was that we need to provide some guidelines when to use `LogStructure` and when `log.Info` , `log.Trace` ....

roytman (commented):
@simanadler I did not understand why you assigned the issue to me, assume you meant to ask for review  ;-).

pkg/logging/logging.go
roytman: P.S. as I remember golint doesn't require to create a constant for "1"

simanadler: @roytman Golint does require a constant.  Also, skipping only one or two frames didn't solve the problem.  Only once I put 3 was it resolved.

site/docs/contribute/logging.md
simanadler: @roytman See https://fybrik.io/dev/contribute/logging/#log-entry-verbosity for when to use the different levels of logging.  LogStructure is when you want to print out a golang structure to be included in the log.  It handles the conversion to json.

pkg/logging/logging.go
roytman: @simanadler , i can show you that the skip should be a single frame, see rego_file_evaluator.go:54, do you see a log statement there ? 
And golint is configured to skip "1", see https://github.com/fybrik/fybrik/blob/451c4192920c132350d83b1678edbff3f34c90a7/.golangci.yml#L64

site/docs/contribute/logging.md
roytman: sorry, but I don't think that the explanation is clear.

roytman:
The implemented `LogStructure` function has 2 issues: (a) it always marshals supplied object independently on the configured verbosy level; (b) it always prints messages in the DEBUG level, maybe we would like to print something in other levels too. 
I'd suggest changing the function to 
```
func LogStructure(argName string, argStruct interface{}, log *zerolog.Logger, level zerolog.Level, forUser, audit bool) {
	if log.GetLevel() > level {
		return
	}
	var jsonStruct []byte
	var err error
	if PrettyLogging() {
		jsonStruct, err = json.MarshalIndent(argStruct, "", "\t")
	} else {
		jsonStruct, err = json.Marshal(argStruct)
	}

	if err != nil {
		log.Warn().Msg("Failed converting " + argName + " to json")
	} else {
		// Log the info making sure that the calling function is listed as the caller
		log.WithLevel(level).CallerSkipFrame(3).Bool(FORUSER, forUser).Bool(AUDIT, audit).Msg(argName + ": " + string(jsonStruct))
	}
}
```
What do you think ?

pkg/logging/logging.go
simanadler: @roytman fixed

site/docs/contribute/logging.md
simanadler: @roytman I rewrote it. Hopefully it's clearer now.

simanadler:
> The implemented `LogStructure` function has 2 issues: (a) it always marshals supplied object independently on the configured verbosy level; (b) it always prints messages in the DEBUG level, maybe we would like to print something in other levels too. I'd suggest changing the function to
> 
> ```
> func LogStructure(argName string, argStruct interface{}, log *zerolog.Logger, level zerolog.Level, forUser, audit bool) {
> 	if log.GetLevel() > level {
> 		return
> 	}
> 	var jsonStruct []byte
> 	var err error
> 	if PrettyLogging() {
> 		jsonStruct, err = json.MarshalIndent(argStruct, "", "\t")
> 	} else {
> 		jsonStruct, err = json.Marshal(argStruct)
> 	}
> 
> 	if err != nil {
> 		log.Warn().Msg("Failed converting " + argName + " to json")
> 	} else {
> 		// Log the info making sure that the calling function is listed as the caller
> 		log.WithLevel(level).CallerSkipFrame(3).Bool(FORUSER, forUser).Bool(AUDIT, audit).Msg(argName + ": " + string(jsonStruct))
> 	}
> }
> ```
> 
> What do you think ?

Done

pkg/logging/logging.go
```diff
@@ -130,6 +133,7 @@ func LogStructure(argName string, argStruct interface{}, log *zerolog.Logger, fo
 	if err != nil {
 		log.Warn().Msg("Failed converting " + argName + " to json")
```
roytman, line 134: in order to prevent missing messages, maybe to add 
log.WithLevel(verbosity).CallerSkipFrame(1).Bool(FORUSER, forUser).Bool(AUDIT, audit).Msg(argName + ": " + fmt.Sprintf("%v", argStruct))

roytman:
@simanadler , thanks for all fixes.
a nit comment, if you're changing the `LogStructure` could you please fix its comment - "... prints out **a the** provided ..."

pkg/logging/logging.go
simanadler: done

simanadler:
> @simanadler , thanks for all fixes. a nit comment, if you're changing the `LogStructure` could you please fix its comment - "... prints out **a the** provided ..."

done

roytman (approved):

roytman:
the golint error can be resolved by:
```
if err != nil {
   log.Warn().Msg("Failed converting " + argName + " to json")
   jsonStruct = []byte(fmt.Sprintf("%v", argStruct))
}
// Log the info making sure that the calling function is listed as the caller
log.WithLevel(verbosity).CallerSkipFrame(1).Bool(FORUSER, forUser).Bool(AUDIT, audit).Msg(argName + ": " + string(jsonStruct))
```

PR #1405, opened 2022-04-03
Handle the case of IsNewDataset is true in write flow.

Mohammad-nassar10:
Create a new asset to handle the case of `IsNewDataSet = true` in write flow. Also, adding a unit test to check this scenario.

TODO in next PR: register the asset in the catalog. 

Signed-off-by: mohammad-nassar10 <mohammad.nassar@ibm.com>
Co-authored-by: Revital Sur [eres@il.ibm.com](mailto:eres@il.ibm.com)
Co-authored-by: Shlomit Koyfman [shlomitk@il.ibm.com](mailto:shlomitk@il.ibm.com)

### DIFF
--- a/manager/controllers/app/fybrikapplication_controller.go
+++ b/manager/controllers/app/fybrikapplication_controller.go
@@ -492,7 +492,7 @@ func (r *FybrikApplicationReconciler) constructDataInfo(req *DataInfo, appContex
 	input := appContext.Application
 	log := appContext.Log.With().Str(logging.DATASETID, req.Context.DataSetID).Logger()
 	var err error
-	if req.Context.Requirements.FlowParams.IsNewDataSet == false {
+	if !req.Context.Requirements.FlowParams.IsNewDataSet {
 		var credentialPath string
 		if input.Spec.SecretRef != "" {
 			if !utils.IsVaultEnabled() {
@@ -784,7 +784,7 @@ func (r *FybrikApplicationReconciler) buildSolution(applicationContext Applicati
 			continue
 		}
 		// If the flag IsNewDataSet is true then a new asset must be allocated and registered to a catalog
-		if requirements[ind].Context.Requirements.FlowParams.IsNewDataSet == true {
+		if requirements[ind].Context.Requirements.FlowParams.IsNewDataSet {
 			err = plotterGen.HandleNewAsset(&requirements[ind], applicationContext.Application, &path, plotterSpec)
 			if err != nil {
 				setErrorCondition(applicationContext, requirements[ind].Context.DataSetID, err.Error())
--- a/manager/controllers/app/fybrikapplication_controller.go
+++ b/manager/controllers/app/fybrikapplication_controller.go
@@ -765,7 +765,6 @@ func (r *FybrikApplicationReconciler) buildSolution(applicationContext Applicati
 		Owner:              client.ObjectKeyFromObject(applicationContext.Application),
 		Provision:          r.Provision,
 		ProvisionedStorage: make(map[string]NewAssetInfo),
-		DataCatalog:        r.DataCatalog,
 	}
 
 	plotterSpec := &api.PlotterSpec{
--- a/manager/controllers/app/fybrikapplication_controller.go
+++ b/manager/controllers/app/fybrikapplication_controller.go
@@ -491,32 +491,34 @@ func (r *FybrikApplicationReconciler) constructDataInfo(req *DataInfo, appContex
 	// Call the DataCatalog service to get info about the dataset
 	input := appContext.Application
 	log := appContext.Log.With().Str(logging.DATASETID, req.Context.DataSetID).Logger()
-	var credentialPath string
-	if input.Spec.SecretRef != "" {
-		if !utils.IsVaultEnabled() {
-			log.Error().Str("SecretRef", input.Spec.SecretRef).Msg("SecretRef defined [%s], but vault is disabled")
-		} else {
-			credentialPath = utils.GetVaultAddress() + vault.PathForReadingKubeSecret(input.Namespace, input.Spec.SecretRef)
-		}
-	}
-	var response *datacatalog.GetAssetResponse
 	var err error
-	request := datacatalog.GetAssetRequest{
-		AssetID:       taxonomy.AssetID(req.Context.DataSetID),
-		OperationType: datacatalog.READ}
+	if req.Context.Requirements.FlowParams.IsNewDataSet == false {
+		var credentialPath string
+		if input.Spec.SecretRef != "" {
+			if !utils.IsVaultEnabled() {
+				log.Error().Str("SecretRef", input.Spec.SecretRef).Msg("SecretRef defined [%s], but vault is disabled")
+			} else {
+				credentialPath = utils.GetVaultAddress() + vault.PathForReadingKubeSecret(input.Namespace, input.Spec.SecretRef)
+			}
+		}
+		var response *datacatalog.GetAssetResponse
+		request := datacatalog.GetAssetRequest{
+			AssetID:       taxonomy.AssetID(req.Context.DataSetID),
+			OperationType: datacatalog.READ}
 
-	if response, err = r.DataCatalog.GetAssetInfo(&request, credentialPath); err != nil {
-		log.Error().Err(err).Msg("failed to receive the catalog connector response")
-		return err
-	}
+		if response, err = r.DataCatalog.GetAssetInfo(&request, credentialPath); err != nil {
+			log.Error().Err(err).Msg("failed to receive the catalog connector response")
+			return err
+		}
 
-	err = r.ValidateAssetResponse(response, DataCatalogTaxonomy, req.Context.DataSetID)
-	if err != nil {
-		log.Error().Err(err).Msg("failed to validate the catalog connector response")
-		return err
+		err = r.ValidateAssetResponse(response, DataCatalogTaxonomy, req.Context.DataSetID)
+		if err != nil {
+			log.Error().Err(err).Msg("failed to validate the catalog connector response")
+			return err
+		}
+		logging.LogStructure("Catalog connector response", response, &log, zerolog.DebugLevel, false, false)
+		response.DeepCopyInto(req.DataDetails)
 	}
-	logging.LogStructure("Catalog connector response", response, &log, zerolog.DebugLevel, false, false)
-	response.DeepCopyInto(req.DataDetails)
 	configEvaluatorInput := &adminconfig.EvaluatorInput{}
 	configEvaluatorInput.Workload.UUID = utils.GetFybrikApplicationUUID(input)
 	input.Spec.AppInfo.DeepCopyInto(&configEvaluatorInput.Workload.Properties)
@@ -763,6 +765,7 @@ func (r *FybrikApplicationReconciler) buildSolution(applicationContext Applicati
 		Owner:              client.ObjectKeyFromObject(applicationContext.Application),
 		Provision:          r.Provision,
 		ProvisionedStorage: make(map[string]NewAssetInfo),
+		DataCatalog:        r.DataCatalog,
 	}
 
 	plotterSpec := &api.PlotterSpec{
@@ -779,6 +782,14 @@ func (r *FybrikApplicationReconciler) buildSolution(applicationContext Applicati
 		if err != nil {
 			setErrorCondition(applicationContext, requirements[ind].Context.DataSetID, err.Error())
 			continue
+		}
+		// If the flag IsNewDataSet is true then a new asset must be allocated and registered to a catalog
+		if requirements[ind].Context.Requirements.FlowParams.IsNewDataSet == true {
+			err = plotterGen.HandleNewAsset(&requirements[ind], applicationContext.Application, &path, plotterSpec)
+			if err != nil {
+				setErrorCondition(applicationContext, requirements[ind].Context.DataSetID, err.Error())
+				return plotterGen.ProvisionedStorage, plotterSpec, err
+			}
 		}
 		err = plotterGen.AddFlowInfoForAsset(&requirements[ind], applicationContext.Application, &path, plotterSpec)
 		if err != nil {
--- a/manager/controllers/app/fybrikapplication_controller.go
+++ b/manager/controllers/app/fybrikapplication_controller.go
@@ -784,7 +784,7 @@ func (r *FybrikApplicationReconciler) buildSolution(applicationContext Applicati
 		}
 		// If the flag IsNewDataSet is true then a new asset must be allocated
 		if requirements[ind].Context.Requirements.FlowParams.IsNewDataSet {
-			err = plotterGen.HandleNewAsset(&requirements[ind], applicationContext.Application, &path, plotterSpec)
+			err = plotterGen.handleNewAsset(&requirements[ind], &path)
 			if err != nil {
 				setErrorCondition(applicationContext, requirements[ind].Context.DataSetID, err.Error())
 				return plotterGen.ProvisionedStorage, plotterSpec, err
--- a/manager/controllers/app/fybrikapplication_controller.go
+++ b/manager/controllers/app/fybrikapplication_controller.go
@@ -782,7 +782,7 @@ func (r *FybrikApplicationReconciler) buildSolution(applicationContext Applicati
 			setErrorCondition(applicationContext, requirements[ind].Context.DataSetID, err.Error())
 			continue
 		}
-		// If the flag IsNewDataSet is true then a new asset must be allocated and registered to a catalog
+		// If the flag IsNewDataSet is true then a new asset must be allocated
 		if requirements[ind].Context.Requirements.FlowParams.IsNewDataSet {
 			err = plotterGen.HandleNewAsset(&requirements[ind], applicationContext.Application, &path, plotterSpec)
 			if err != nil {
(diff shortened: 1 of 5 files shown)


manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -491,32 +491,34 @@ func (r *FybrikApplicationReconciler) constructDataInfo(req *DataInfo, appContex
...
-		} else {
-			credentialPath = utils.GetVaultAddress() + vault.PathForReadingKubeSecret(input.Namespace, input.Spec.SecretRef)
-		}
-	}
-	var response *datacatalog.GetAssetResponse
 	var err error
-	request := datacatalog.GetAssetRequest{
-		AssetID:       taxonomy.AssetID(req.Context.DataSetID),
-		OperationType: datacatalog.READ}
+	if req.Context.Requirements.FlowParams.IsNewDataSet == false {
```
shlomitk1, line 495: simplify: if !req.Context.xxx

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -780,6 +783,14 @@ func (r *FybrikApplicationReconciler) buildSolution(applicationContext Applicati
 			setErrorCondition(applicationContext, requirements[ind].Context.DataSetID, err.Error())
 			continue
 		}
+		// If the flag IsNewDataSet is true then a new asset must be allocated and registered to a catalog
+		if requirements[ind].Context.Requirements.FlowParams.IsNewDataSet == true {
```
shlomitk1, line 787: omit "== true"

manager/controllers/app/fybrikapplication_controller_unit_test.go
```diff
@@ -1389,6 +1389,84 @@ func TestReadAndTransform(t *testing.T) {
...
+	s := utils.NewScheme(g)
+
+	// Create a fake client to mock API calls.
+	cl := fake.NewFakeClientWithScheme(s, objs...)
+
+	// Read module
+	readWriteModule := &app.FybrikModule{}
+	g.Expect(readObjectFromFile("../../testdata/unittests/module-read-write.yaml", readWriteModule)).NotTo(gomega.HaveOccurred())
+	readWriteModule.Namespace = utils.GetControllerNamespace()
+	g.Expect(cl.Create(context.TODO(), readWriteModule)).NotTo(gomega.HaveOccurred(), "the read module could not be created")
```
shlomitk1, line 1421: write instead of read

manager/controllers/app/fybrikapplication_controller_unit_test.go
```diff
@@ -1389,6 +1389,84 @@ func TestReadAndTransform(t *testing.T) {
...
+	_, err := r.Reconcile(context.Background(), req)
+	g.Expect(err).To(gomega.BeNil())
+
+	err = cl.Get(context.TODO(), req.NamespacedName, application)
+	g.Expect(err).To(gomega.BeNil(), "Cannot fetch fybrikapplication")
+	g.Expect(getErrorMessages(application)).To(gomega.BeEmpty())
+	// check plotter creation
+	g.Expect(application.Status.AssetStates).To(gomega.HaveLen(1))
+	g.Expect(application.Status.Generated).ToNot(gomega.BeNil())
+	readOriginalDatalEndpoint := application.Status.AssetStates[application.Spec.Data[0].DataSetID].Endpoint
```
shlomitk1, line 1450: writeEndpoint

manager/controllers/app/fybrikapplication_controller_unit_test.go
```diff
@@ -1389,6 +1389,84 @@ func TestReadAndTransform(t *testing.T) {
...
+
+	plotterObjectKey := types.NamespacedName{
+		Namespace: application.Status.Generated.Namespace,
+		Name:      application.Status.Generated.Name,
+	}
+	plotter := &app.Plotter{}
+	err = cl.Get(context.Background(), plotterObjectKey, plotter)
+	g.Expect(err).NotTo(gomega.HaveOccurred())
+	g.Expect(plotter.Spec.Assets).To(gomega.HaveLen(1))
+	g.Expect(plotter.Spec.Templates).To(gomega.HaveLen(2))
```
shlomitk1, line 1467: Why two templates?

Mohammad-nassar10: It should be one.

shlomitk1 (changes requested):
There is already a place holder for functionality to register a new asset. It is done in checkReadiness function when fybrikApp gets a Ready status. We don't need to handle asset registration in several places. Besides, a bucket should be marked as permanent what is already done there.

manager/controllers/app/plotter_generator.go
```diff
@@ -201,6 +204,55 @@ func (p *PlotterGenerator) addStep(element *ResolvedEdge, datasetID string, api
...
+// Create a new asset and register it to the catalog. Used when the IsNewDataSet flag is true
+func (p *PlotterGenerator) HandleNewAsset(item *DataInfo, application *app.FybrikApplication, selection *Solution,
+	plotterSpec *app.PlotterSpec) error {
+	var err error
+	if item.DataDetails.Details.DataFormat != "" {
+		return nil
+	}
+	var sinkDataStore *app.DataStore
+	selectionLen := len(selection.DataPath)
+	element := selection.DataPath[selectionLen-1]
```
shlomitk1, line 216: It would be cleaner to go over the DataPath and look for an element with a non-empty storage account instead of relying on it to be the last.

manager/controllers/app/plotter_generator.go
```diff
@@ -201,6 +204,55 @@ func (p *PlotterGenerator) addStep(element *ResolvedEdge, datasetID string, api
...
+	// allocate storge
+	if sinkDataStore, err = p.AllocateStorage(item, element.Sink.Connection, &element.StorageAccount); err != nil {
+		p.Log.Error().Err(err).Str(logging.DATASETID, item.Context.DataSetID).Msg("Storage allocation failed")
+		return err
+	}
+
+	item.DataDetails.Details.DataFormat = sinkDataStore.Format
```
shlomitk1, line 223: Not only data format should be populated

manager/controllers/app/select_modules.go
```diff
@@ -50,9 +50,16 @@ func (p *PathBuilder) solve() (Solution, error) {
 // Then, transformations are added to the found paths, and clusters are matched to satisfy restrictions from admin config policies.
 // Optimization is done by the shortest path (the paths are sorted by the length). To be changed in future versions.
 func (p *PathBuilder) FindPaths() []Solution {
+	var protocol taxonomy.ConnectionType
+	// If the connection name is empty, the default protocol is s3.
+	if p.Asset.DataDetails.Details.Connection.Name == "" {
```
shlomitk1, line 55: Wouldn't the entire structure be nil?

manager/controllers/app/select_modules.go
```diff
@@ -50,9 +50,16 @@ func (p *PathBuilder) solve() (Solution, error) {
...
+		protocol = "s3"
```
shlomitk1, line 56: There is somewhere a constant for s3

manager/controllers/app/select_modules.go
```diff
@@ -50,9 +50,16 @@ func (p *PathBuilder) solve() (Solution, error) {
...
+	} else {
+		protocol = p.Asset.DataDetails.Details.Connection.Name
+	}
 	NodeFromAssetMetadata := Node{
 		Connection: &taxonomy.Interface{
-			Protocol:   p.Asset.DataDetails.Details.Connection.Name,
+			Protocol:   protocol,
 			DataFormat: p.Asset.DataDetails.Details.DataFormat,
```
shlomitk1, line 63: Where is this coming from? Changes to format are done after the path has been found.

manager/controllers/app/plotter_generator.go
revit13: @shlomitk1 we fixed that. Please take a look. Thanks

manager/controllers/app/select_modules.go
revit13: It seems to be allocated anyway: https://github.com/fybrik/fybrik/blob/master/manager/controllers/app/fybrikapplication_controller.go#L367

manager/controllers/app/select_modules.go
revit13: @shlomitk1 please explain, in this line we changed protocol not dataformat.

shlomitk1: p.Asset.DataDetails.Details.DataFormat is not set. If you want to use an empty string, do it explicitly.

revit13: @shlomitk1 this line already exists in the current code: https://github.com/fybrik/fybrik/blob/master/manager/controllers/app/select_modules.go#L56
it does not refer only to the case where dataformat is empty.

shlomitk1: What I mean is that you change protocol, but not data format. Both are not set. It's better to set values explicitly and not rely on an empty field value.

revit13: Fixed. Thanks

manager/controllers/app/plotter_generator.go
revit13: Done. Thanks

manager/controllers/app/fybrikapplication_controller_unit_test.go
```diff
@@ -1389,6 +1389,85 @@ func TestReadAndTransform(t *testing.T) {
 	g.Expect(plotter.Spec.Flows[0].SubFlows[0].Steps[0]).To(gomega.HaveLen(2))
 }
 
+func TestWriteNotRegisteredAsset(t *testing.T) {
```
shlomitk1, line 1392: Looks like copy-paste from another test. You want to test here that 1. a plotter is generated successfully and 2. the asset connection in the plotter is s3 and has non-empty format.

manager/controllers/app/plotter_generator.go
```diff
@@ -201,6 +201,61 @@ func (p *PlotterGenerator) addStep(element *ResolvedEdge, datasetID string, api
...
+		p.Log.Error().Err(err).Str(logging.DATASETID, item.Context.DataSetID).Msg("Storage allocation failed")
+		return err
+	}
+
+	// Reset StorageAccount to prevent re-allocation
+	element.StorageAccount.Region = ""
+
+	resourceMetadata := datacatalog.ResourceMetadata{
+		Name:      item.Context.DataSetID,
+		Geography: string(element.StorageAccount.Region),
```
shlomitk1, line 238: You have emptied this value in line 234.

manager/controllers/app/select_modules.go
```diff
@@ -230,9 +240,13 @@ func (p *PathBuilder) findPathsWithinLimit(source, sink *Node, n int) []Solution
 			if !supportsSinkInterface(&edge, sink) {
 				continue
 			}
+
 			edge.Sink = sink
 			// if a module supports both source and sink interfaces, it's an end of the recursion
 			if supportsSourceInterface(&edge, source) {
+				if sink.Connection.DataFormat == "" {
```
shlomitk1, line 247: Why here? It should be done at the end when the path has been validated.

manager/controllers/app/fybrikapplication_controller_unit_test.go
revit13: Fixed. Thanks

manager/controllers/app/plotter_generator.go
revit13: Fixed. Thanks

manager/controllers/app/select_modules.go
revit13: Fixed. Thanks

manager/controllers/app/fybrikapplication_controller_unit_test.go
```diff
@@ -1389,6 +1389,89 @@ func TestReadAndTransform(t *testing.T) {
...
+	}
+
+	_, err := r.Reconcile(context.Background(), req)
+	g.Expect(err).To(gomega.BeNil())
+
+	err = cl.Get(context.TODO(), req.NamespacedName, application)
+	g.Expect(err).To(gomega.BeNil(), "Cannot fetch fybrikapplication")
+	g.Expect(getErrorMessages(application)).To(gomega.BeEmpty())
+	// check plotter creation
+	g.Expect(application.Status.AssetStates).To(gomega.HaveLen(1))
```
shlomitk1, line 1449: Not needed. AssetStates is of the same length as number of assets.

manager/controllers/app/fybrikapplication_controller_unit_test.go
```diff
@@ -1389,6 +1389,89 @@ func TestReadAndTransform(t *testing.T) {
...
+	g.Expect(application.Status.Generated).ToNot(gomega.BeNil())
+	writeDatalEndpoint := application.Status.AssetStates[application.Spec.Data[0].DataSetID].Endpoint
```
shlomitk1, line 1451: writeEndpoint

manager/controllers/app/fybrikapplication_controller_unit_test.go
```diff
@@ -1389,6 +1389,89 @@ func TestReadAndTransform(t *testing.T) {
...
+
+	plotterObjectKey := types.NamespacedName{
+		Namespace: application.Status.Generated.Namespace,
+		Name:      application.Status.Generated.Name,
+	}
+	plotter := &app.Plotter{}
+	err = cl.Get(context.Background(), plotterObjectKey, plotter)
+	g.Expect(err).NotTo(gomega.HaveOccurred())
+	g.Expect(plotter.Spec.Assets).To(gomega.HaveLen(1))
+	g.Expect(plotter.Spec.Assets["s3-not-exists/new-dataset"].DataStore.Connection.Name == app.S3).
```
shlomitk1, line 1468: Rewrite Expect statements to use gomega.Equal, gomega.BeEmpty, ...

manager/controllers/app/plotter_generator.go
```diff
@@ -201,6 +201,62 @@ func (p *PlotterGenerator) addStep(element *ResolvedEdge, datasetID string, api
 	return steps
 }
 
+// Handle a new asset: allocate storage and update its metadata. Used when the
+// IsNewDataSet flag is true.
+func (p *PlotterGenerator) HandleNewAsset(item *DataInfo, application *app.FybrikApplication, selection *Solution,
```
shlomitk1, line 206: handleNewAsset

manager/controllers/app/select_modules.go
```diff
@@ -78,9 +88,19 @@ func (p *PathBuilder) FindPaths() []Solution {
 	}
 	// get valid solutions by extending data paths with transformations and selecting an appropriate cluster for each capability
 	solutions = p.validSolutions(solutions)
+	// set empty data format in sink nodes from the chosen modules data format
+	p.setSinkDataFormatFromModule(solutions)
+
 	return solutions
 }
 
+// set empty data formats in sink nodes with the modules data format
```
shlomitk1, line 97: This is relevant for new assets only. Why not deal with format as well in handleNewAsset()?

revit13: Fixed. Thanks

revit13:
@shlomitk1 please review again. As discussed we will address the catalog registration in a separate pr. Thanks

manager/controllers/app/plotter_generator.go
```diff
@@ -201,6 +201,80 @@ func (p *PlotterGenerator) addStep(element *ResolvedEdge, datasetID string, api
 	return steps
 }
 
+// setEmptySinkDataFormat sets empty data format in sink nodes to be the first
+// dataformat supported by the modules sink interface.
+func (p *PlotterGenerator) setEmptySinkDataFormat(solution *Solution) {
```
shlomitk1, line 206: There are several issues with function. First, it goes over all supported sinks of the module while checking in each iteration whether DataFormat is empty. Also, it goes over all nodes while in the calling function we have a specific node that should be fixed.
I would suggest to replace this function with getSupportedFormat(capability) that will return a supported dataFormat of the selected capability, and call it prior to AllocateStorage, i.e. when we have an element that needs storage allocation.

manager/controllers/app/select_modules.go
```diff
@@ -391,10 +394,22 @@ func supportsSinkInterface(edge *Edge, sinkNode *Node) bool {
...
-	return &Node{Connection: &taxonomy.Interface{
-		Protocol:   p.Asset.DataDetails.Details.Connection.Name,
-		DataFormat: p.Asset.DataDetails.Details.DataFormat,
-	}}
+	var protocol taxonomy.ConnectionType
+	var dataFormat taxonomy.DataFormat
+	// If the connection name is empty, the default protocol is s3.
+	if p.Asset.DataDetails == nil || p.Asset.DataDetails.Details.Connection.Name == "" {
+		protocol = app.S3
+		dataFormat = ""
```
shlomitk1, line 402: I think var declaration will set dataFormat to an empty string.

shlomitk1 (changes requested):
Please see my comment about setting a dataFormat field when empty.

manager/controllers/app/plotter_generator.go
revit13: We changed the implementation. Thanks

manager/controllers/app/plotter_generator.go
```diff
@@ -201,6 +201,77 @@ func (p *PlotterGenerator) addStep(element *ResolvedEdge, datasetID string, api
 	return steps
 }
 
+// getSupportedFormat returns the first dataformat supported by the module's capability sink interface
+func (p *PlotterGenerator) getSupportedFormat(capability *app.ModuleCapability) taxonomy.DataFormat {
+	for _, inter := range capability.SupportedInterfaces {
+		if inter.Sink == nil {
```
shlomitk1, line 207: rewrite:
for ... {
    if inter.Sink != nil {
        return inter.Sink.DataFormat
    }
}    
...

shlomitk1 (approved):

PR #1407, opened 2022-04-05
Change http method in getAssetInfo and getPoliciesDecisions from POST to GET

rohithdv:
Based on suggestions by @roytman, this PR has the following improvements to the taxonomy:
- [x] Change getAssetInfo POST to GET
- [x] Change getPoliciesDecisions POST to GET
- [x] Add logs in opa connector
- [x] Updated client code generation for policy manager with respect to openapi-generator 5.4.0 (Current master version was compiled for version 5.3.0 which is quite dated)

PS: Once this PR is accepted, We would need updates in wkc connector also (TODO: This PR needs to be raised in WKC Connector repo).


connectors/opa/connector.go
```diff
@@ -83,6 +90,8 @@ func (r *ConnectorController) GetPoliciesDecisions(c *gin.Context) {
 		c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
 		return
 	}
+	r.log.Info().Msg(
```
roytman, line 93: All other log messages are printed with `DEBUG` verbosity level, but this one with `INFO` , why?

roytman (dismissed):
A small logging-related comment. 
Set it to draft to prevent merge, as mentioned in the PR, we have to update the WKC connector too

roytman:
looks like [install_openapi-generator-cli.sh](https://github.com/fybrik/fybrik/blob/master/hack/tools/install_openapi-generator-cli.sh)  tries to install the latest version. In my case it was 5.4
If you want to update the generated files you have to update install_openapi-generator-cli.sh too.
and https://github.com/fybrik/fybrik/blob/7a658a8d72c22d001d84e7433ca9289568fdaddd/site/Makefile#L6

connectors/opa/connector.go
rohithdv: Fixed. Changed log level to Debug in opa connector and katalog-connector. Please check. Thanks.

PR #1408, opened 2022-04-05
An optional interface in dataset requirements

shlomitk1:
Fixes https://github.com/fybrik/fybrik/issues/1381


manager/controllers/app/select_modules.go
```diff
@@ -324,7 +322,14 @@ func match(source, sink *taxonomy.Interface) bool {
 	if source == nil || sink == nil {
 		return false
 	}
-	return source.DataFormat == sink.DataFormat && source.Protocol == sink.Protocol
+	if source.Protocol != sink.Protocol {
+		return false
+	}
+	// an empty DataFormat value is not checked
+	if source.DataFormat != "" && sink.DataFormat != "" && source.DataFormat != sink.DataFormat {
```
roytman, line 329: in the following two cases the return will be `true` 
source.DataFormat  == "" but sink.DataFormat != ""
source.DataFormat  != "" but sink.DataFormat == ""

is it correct ?

manager/controllers/app/select_modules.go
```diff
@@ -338,18 +343,21 @@ func supportsSourceInterface(edge *Edge, source *Node) bool {
 		}
 		hasSources = true
 		// connection via Source
-		if match(inter.Source, source.Connection) {
+		if source != nil && match(inter.Source, source.Connection) {
```
roytman, line 346: I'd consider to change the name, otherwise, it looks weird when you are comparing `Source` and `source.Connections` 
maybe instead of `source` use `sourceNode` ?

manager/controllers/app/select_modules.go
```diff
@@ -363,21 +371,38 @@ func supportsSinkInterface(edge *Edge, sink *Node) bool {
 			continue
 		}
 		hasSinks = true
-		if match(inter.Sink, sink.Connection) {
+		if sink != nil && match(inter.Sink, sink.Connection) {
```
roytman, line 374: the same comment about names

manager/controllers/app/select_modules.go
shlomitk1: Changed.

revit13 (approved):

PR #1416, opened 2022-04-07
Resolve update conflicts

shlomitk1:
https://github.com/fybrik/fybrik/pull/1385 for releases/0.6.5

### DIFF
--- a/manager/controllers/app/fybrikapplication_controller.go
+++ b/manager/controllers/app/fybrikapplication_controller.go
@@ -237,7 +237,7 @@ func (r *FybrikApplicationReconciler) reconcileFinalizers(ctx context.Context, a
 	hasFinalizer := ctrlutil.ContainsFinalizer(applicationContext.Application, finalizerName)
 
 	// If the object has a scheduled deletion time, delete it and all resources it has created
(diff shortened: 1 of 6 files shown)


manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -137,10 +136,8 @@ func (r *FybrikApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		if result, err := r.reconcile(applicationContext); err != nil {
 			// another attempt will be done
 			// users should be informed in case of errors
-			if !equality.Semantic.DeepEqual(&application.Status, observedStatus) {
-				// ignore an update error, a new reconcile will be made in any case
-				_ = r.Client.Status().Update(ctx, application)
-			}
+			// ignore an update error, a new reconcile will be made in any case
```
roytman, line 139: if we do reconcile/update at any case, why do we need this Update call at all ?

shlomitk1: If the error does not go away, it will never be propagated to the status.

roytman (approved):

PR #1396, opened 2022-03-30
Design DeleteAsset interface in Fybrik

rohithdv:
This PR add `DeleteAsset` method in the Fybrik Data catalog interface. The corresponding REST method is called `deleteAsset`. 

This PR partially fixes #1242 .

Signed-off-by: Rohith D Vallam [rovallam@in.ibm.com](mailto:rovallam@in.ibm.com)

PR #1426, opened 2022-04-11
DeleteAsset implementation in katalog connector

### LINKED ISSUE #1242: Catalog Connector Delete Dataset
Add interface to delete a dataset from a data catalog , and implement an example implementation in katalog

rohithdv:
This PR is a followup to #1396 . This PR partially fixes #1242 .
This PR adds `deleteAsset` implementation to `katalog connector`. This PR also has minor changes to the deleteAsset openapi definition.  


connectors/api/datacatalog.spec.yaml
```diff
@@ -90,4 +90,8 @@ paths:
                 schema:
                   $ref: "../../charts/fybrik/files/taxonomy/datacatalog.json#/definitions/DeleteAssetResponse"
           '400':
-            description: Bad request - server cannot process the request due to client error
\ No newline at end of file
+            description: Bad request - server cannot process the request due to client error
+          '404':
```
shlomitk1, line 94: BTW, why don't we have the same for getAsset in case the id does not exist?

pkg/connectors/datacatalog/openapiclient/.openapi-generator/FILES
```diff
@@ -0,0 +1,13 @@
+.gitignore
```
shlomitk1, line 1: Does this file need to be committed?

roytman (commented):
@rohithdv we have to discuss the catalog REST API

pkg/connectors/datacatalog/openapiclient/.openapi-generator/FILES
rohithdv: This file is now removed from the commit. Please check. Thanks.

.gitignore
```diff
@@ -63,3 +63,4 @@ pipeline/vault_cre*
 pipeline/custom-*
 
 dist/
+pkg/connectors/datacatalog/openapiclient/.openapi-generator/FILES
```
roytman, line 66: do we need this entry ?

rohithdv: It may be needed in the scenario where we update the openapi spec and generate the client code again. In this case the file FILES will be recreated and we may mistakenly checkin the file. So it may be better to add this file in the gitignore.

roytman (approved):
Probably we will get back to the REST API implementation.

PR #1435, opened 2022-04-12
Run integration tests with clusterScoped false

### LINKED ISSUE #1434: Run integration tests with clusterScoped false
This issue is about adding another run of integration tests with  fybrik deployed using clusterScoped false.

revit13:
Closes #1434 

This PR includes:
- a new target in the Makefile to run integration tests with clusterScoped false.
- a new step in build workflow to run the above target


.github/workflows/build.yml
```diff
@@ -92,6 +92,11 @@ jobs:
         restore-keys: ${{ runner.os }}-t-${{ hashFiles('hack/make-rules/tools.mk') }}-go
     - name: Install tools
       run: make install-tools
+    - name: Integration tests no cluster scope new
+      if: ${{ matrix.k8s_version == 'new' }}
+      env:
+        K8S_VERSION: ${{ env.NEW_K8S_VERSION }}
+      run: make run-integration-tests-no-cluster-scope
```
shlomitk1, line 99: You can call it run-namespace-scoped-tests

Makefile
```diff
@@ -99,6 +99,25 @@ run-notebook-tests:
...
+run-integration-tests-no-cluster-scope:
+	$(MAKE) kind
+	$(MAKE) cluster-prepare
+	$(MAKE) docker-build docker-push
+	$(MAKE) -C test/services docker-build docker-push
+	$(MAKE) cluster-prepare-wait
+	$(MAKE) -C charts test
+	$(MAKE) deploy
+	$(MAKE) configure-vault
+	$(MAKE) -C modules helm
```
shlomitk1, line 115: Helm tests are independent of a scope and can be removed.

Makefile
```diff
@@ -99,6 +99,25 @@ run-notebook-tests:
...
+	$(MAKE) -C modules helm-uninstall # Uninstalls the deployed tests from previous command
+	$(MAKE) -C pkg/helm test
+	$(MAKE) -C manager run-integration-tests
+	$(MAKE) -C modules test
```
shlomitk1, line 119: Not needed.

Makefile
```diff
@@ -99,6 +99,25 @@ run-notebook-tests:
...
+	$(MAKE) cluster-prepare
+	$(MAKE) docker-build docker-push
+	$(MAKE) -C test/services docker-build docker-push
+	$(MAKE) cluster-prepare-wait
+	$(MAKE) -C charts test
+	$(MAKE) deploy
+	$(MAKE) configure-vault
+	$(MAKE) -C modules helm
+	$(MAKE) -C modules helm-uninstall # Uninstalls the deployed tests from previous command
+	$(MAKE) -C pkg/helm test
```
shlomitk1, line 117: Not needed.

.github/workflows/build.yml
roytman: I'd call it "Namescope integration tests"

Makefile
roytman: I see that the `test` target in the modules/Makefile is empty and has the following comment 
"# Tests if implicit copy definitions can be installed"

Do we need it? When do we use it?

charts/fybrik/integration-tests.no-cluster-scope.values.yaml
```diff
@@ -0,0 +1,65 @@
+clusterScoped: false
```
roytman, line 1: Do we need this entire file? can we just use the flag `--set` for the `helm install` command to overwrite the values ?

Makefile
revit13: I think it's leftover from the time we had the implicit copy modules in fybrik repo. As far as I know we can remove it.

charts/fybrik/integration-tests.no-cluster-scope.values.yaml
revit13: I think its cleaner to have the entire file, otherwise, we need to add --set to the helm install in https://github.com/fybrik/fybrik/blob/master/Makefile#L44 (and we do not always add new settings)

Makefile
roytman: so, please remove it

revit13: Done. Thanks

shlomitk1 (approved):

roytman (approved):

PR #1219, opened 2022-01-27
Set helm version to 3.8

revit13:
Blocked by https://github.com/fybrik/fybrik/pull/1215 as Helm 3.8 is compatible with K8s v1.20-v1.23 according to https://helm.sh/docs/topics/version_skew/

Closes #1217

PR #1441, opened 2022-04-24
move to the upstream helm

### LINKED ISSUE #1443: Move from a proprietary helm fork to the upstream helm repository
Originally, some of the helm repository methods were under experimental/internal distribution. Therefore, Fybrik used a proprietary helm fork. Recently, helm was reorganized, and it allows using the upstream repository directly.

Elimination of the private helm fork, will significantly simplify the Fybrik release process, help easily include helm updates, and eventually increase Fybrik robustness.

roytman:
Move to the upstream helm repo.

Bump to upstream helm 3.8.2 plus update a list of other third-party dependencies.

/closes #1443
/closes #1219


hack/tools/install_helm.sh
```diff
@@ -5,7 +5,7 @@
 
 source ./common.sh
 
-DESIRED_VERSION=v3.7.2
+DESIRED_VERSION=v3.8.1
```
shlomitk1, line 8: It's 3.8.2 in go.mod - does it matter?

roytman: Thanks, fix it

shlomitk1 (dismissed):

hack/make-rules/helm.mk
```diff
@@ -22,7 +22,6 @@ HELM_TAG ?= 0.0.0
 
 CHART_REGISTRY_PATH := oci://${DOCKER_HOSTNAME}/${DOCKER_NAMESPACE}
 
-export HELM_EXPERIMENTAL_OCI=1
 export GODEBUG=x509ignoreCN=0
```
revit13, line 25: for versions lower than 3.8 HELM_EXPERIMENTAL_OCI is needed and as `DESIRED_VERSION` can  changed I suggest keeping HELM_EXPERIMENTAL_OCI

roytman: Why will somebody use the old version?

revit13 (approved):

PR #1461, opened 2022-05-02
Add the ability to specify tags for new written data

### LINKED ISSUE #1460: Add the ability to specify tags for new written data
Currently, there is no way to specify tags for a new written asset. This issue deals with adding a way to do so.

As suggested by @shlomitk1 fybrikapplication can be extended to include resourceMetadata in the flow requirements that the user will fill, for example:
```
    - dataSetID: 'new-data-parquet'
      flow: write
      requirements:
        flowParams:
          isNewDataSet: true
          catalog: fybrik-notebook-sample
          metadata:
            tags:
              finance: true
            columns:
            - name: nameOrig
              tags:
                PII: true
            - name: oldbalanceOrg
              tags:
                sensitive: true
```

blocked by issue #1458

revit13:
Closes #1460 

Currently, there is no way to specify tags for a new written asset. This issue deals with adding a way to do so.

As suggested by @shlomitk1 fybrikapplication can be extended to include resourceMetadata in the flow requirements that the user will fill, for example:

```
 - dataSetID: 'new-data-parquet'
      flow: write
      requirements:
        flowParams:
          isNewDataSet: true
          catalog: fybrik-notebook-sample
          metadata:
            tags:
              finance: true
            columns:
            - name: nameOrig
              tags:
                PII: true
            - name: oldbalanceOrg
              tags:
                sensitive: true
```

Co-authored-by: mohammad-nassar10 <mohammad.nassar@ibm.com>


manager/controllers/app/data_catalog_interface.go
```diff
@@ -30,18 +32,23 @@ func (r *FybrikApplicationReconciler) RegisterAsset(assetID string, catalogID st
 	}
 
 	var resourceMetadata datacatalog.ResourceMetadata
-	if info.ResourceMetadata != nil {
-		resourceMetadata = *info.ResourceMetadata.DeepCopy()
+	if resourceMetadataFromRequirements != nil {
+		resourceMetadata = *resourceMetadataFromRequirements.DeepCopy()
```
shlomitk1, line 36: DeepCopy is applied twice for the same object

shlomitk1:
I have a problem with using metadata that is not in catalog yet for governance decisions. 
Suppose I want to update someone else's asset which I am not allowed to do since I am not the owner. But I could provide my name as an owner in a new metadata, and suddenly it's ok, I have overwritten the asset marking myself as an owner.

roytman:
> I have a problem with using metadata that is not in catalog yet for governance decisions. Suppose I want to update someone else's asset which I am not allowed to do since I am not the owner. But I could provide my name as an owner in a new metadata, and suddenly it's ok, I have overwritten the asset marking myself as an owner.

Can we check that an asset doesn't exist before creating a new catalog entry for it?

manager/controllers/app/data_catalog_interface.go
```diff
@@ -16,11 +16,13 @@ import (
...
 // - assetID: DataSetID as it appears in fybrik-application
 // - catalogID: the destination catalog identifier
+// - resourceMetadataFromRequirements: resource metadata from the flow requirements
 // - info: connection and credential details
 // Returns:
 // - an error if happened
 // - the new asset identifier
-func (r *FybrikApplicationReconciler) RegisterAsset(assetID string, catalogID string, info *app.DatasetDetails,
+func (r *FybrikApplicationReconciler) RegisterAsset(assetID string, catalogID string,
+	resourceMetadataFromRequirements *datacatalog.ResourceMetadata, info *app.DatasetDetails,
```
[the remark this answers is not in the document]
revit13: I find that merging the resource metadata is overkill as in `provisionedBucketRef.ResourceMetadata` holds only information about the geography. I will try to check conflicts only in that field.

revit13:
@roytman the resource metadata from the application flow requirements can be used only when 'isNewDataSet` is true so we will not update an existing asset. In the current implementation if `isNewDataSet` is true then a new asset is registered in the catalog - it is not updating an existing asset

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -542,23 +546,26 @@ func (r *FybrikApplicationReconciler) checkGovernanceActions(configEvaluatorInpu
...
 		}
 	case taxonomy.ReadFlow, taxonomy.DeleteFlow:
 		reqAction := policymanager.RequestAction{
 			ActionType:         configEvaluatorInput.Request.Usage,
 			Destination:        configEvaluatorInput.Workload.Cluster.Metadata.Region,
 			ProcessingLocation: taxonomy.ProcessingLocation(configEvaluatorInput.Workload.Cluster.Metadata.Region),
 		}
-		req.Actions, err = LookupPolicyDecisions(req.Context.DataSetID, r.PolicyManager, appContext, &reqAction)
+		req.Actions, err = LookupPolicyDecisions(req.Context.DataSetID, req.Context.Requirements.FlowParams.ResourceMetadata,
+			r.PolicyManager, appContext, &reqAction)
```
shlomitk1, line 568: What is the meaning of taking metadata not from the catalog for read/delete scenarios?

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -542,23 +546,26 @@ func (r *FybrikApplicationReconciler) checkGovernanceActions(configEvaluatorInpu
 	var err error
 	switch configEvaluatorInput.Request.Usage {
 	case taxonomy.WriteFlow:
-		if !req.Context.Requirements.FlowParams.IsNewDataSet {
+		if !req.Context.Requirements.FlowParams.IsNewDataSet ||
+			req.Context.Requirements.FlowParams.ResourceMetadata != nil {
```
shlomitk1, line 550: This code is reached by either writing a new asset or updating an existing asset. In both cases it is problematic to use the supplied metadata.

revit13: we will revisit this part. ResourceMetadata from flow requirements should be taken into account only when `isNewDataSet` is true

manager/controllers/app/fybrikapplication_controller.go
revit13: we will revisit this part. ResourceMetadata from flow requirements should be taken into account only when isNewDataSet is true

manager/controllers/app/data_catalog_interface.go
revit13: will fix that thanks

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -542,23 +546,26 @@ func (r *FybrikApplicationReconciler) checkGovernanceActions(configEvaluatorInpu
...
 			// update an existing dataset
 			// query the policy manager whether the operation is allowed
 			reqAction := policymanager.RequestAction{
 				ActionType:         configEvaluatorInput.Request.Usage,
 				Destination:        req.DataDetails.ResourceMetadata.Geography,
```
shlomitk1, line 555: req.DataDetails does not have Geography for a new asset.

manager/controllers/app/fybrikapplication_controller.go
shlomitk1: Even not then. You don't know the destination at this point, so what will the WRITE query be about?

manager/controllers/app/fybrikapplication_controller.go
revit13: ok, it set to empty string in this case...

manager/controllers/app/fybrikapplication_controller.go
revit13: not sure if that answers the question but we have the tags so we can get policies related to it.

pkg/model/datacatalog/model.go
```diff
@@ -43,3 +43,19 @@ type ResourceDetails struct {
 	// Data format
 	DataFormat taxonomy.DataFormat `json:"dataFormat,omitempty"`
 }
+
+// Return true if resourceMetadata is not empty. Otherwise return false.
+func ResourceMetadataNotEmpty(resourceMetadata *ResourceMetadata) bool {
```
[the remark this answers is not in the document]
revit13: Fixed thanks

revit13:
@shlomitk1 please take a look at the updated version in particular the code related to the flow which calls policy compiler in `GetPoliciesDecisions`. please note that the policy compiler is been called also when `isNewDataSet` is true as we need to check for permission to write new data. it is still under testing but I appreciate your review. Thanks

charts/fybrik/files/taxonomy/fybrik_application.json
```diff
@@ -41,11 +41,25 @@
...
+          "$ref": "#/definitions/FlowRequirements",
+          "description": "FlowParams include the requirements for particular data flows"
+        },
         "interface": {
           "$ref": "taxonomy.json#/definitions/Interface",
           "description": "Interface indicates the protocol and format expected by the data user"
         }
       }
+    },
+    "FlowRequirements": {
```
shlomitk1, line 54: How was it generated? FlowRequirements has more fields than just metadata

revit13: hmmm I ran `make all`... strange

shlomitk1:
Looking now at the information sent to the policy manager, it seems that no changes have been made since updating the API. The purpose of Resource field in the request is to avoid an extra call to catalog connector, and separate the connectors completely. Why does opa-connector need to invoke GetAssetInfo if it receives the information about the resource in the request? 
Besides, it is not the responsibility of policy manager to decide how to merge two inputs of metadata.
Manager should provide to the policy manager connector the complete information about the asset in Resource field.

revit13:
@shlomitk1 the call to GetAssetInfo from opa was before the resource metadata change. This pr contains the code to avoid it if resource metadata exists.

shlomitk1:
> @shlomitk1 the call to GetAssetInfo from opa was before the resource metadata change. This pr contains the code to avoid it if resource metadata exists.

And if resource metadata does not exist? How would the call change anything? I'm saying that it is unnecessary, the calls to the policy manager should be updated accordingly.

manager/controllers/app/fybrikapplication_controller.go
revit13: @shlomitk1 as you suggested the metadata is now passed to the policy compiler and the call to the data catalog from policy compiler is removed.

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -217,9 +218,22 @@ func (r *FybrikApplicationReconciler) checkReadiness(applicationContext Applicat
 				setErrorCondition(applicationContext, assetID, err.Error())
 				continue
 			}
+			reqResource := dataCtx.Requirements.FlowParams.ResourceMetadata
+			if reqResource != nil {
+				// we assume to have only the geography field set at this point
+				// in the provisionedBucket ResourceMetadata
+				geo := provisionedBucketRef.ResourceMetadata.Geography
+				if reqResource.Geography != "" && geo != reqResource.Geography {
+					// log an error and make a new attempt to register the asset
```
shlomitk1, line 227: Who will make a new attempt? Print warning that the geography field is going to be ignored, and continue with registration.

revit13: Fixed. Thanks

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -514,6 +528,9 @@ func (r *FybrikApplicationReconciler) constructDataInfo(req *DataInfo, appContex
 		}
 		logging.LogStructure("Catalog connector response", response, &log, zerolog.DebugLevel, false, false)
 		response.DeepCopyInto(req.DataDetails)
+	} else if req.Context.Requirements.FlowParams.ResourceMetadata != nil {
+		// Fill req.DataDetails with the metadata from the fybrikapplication
+		req.DataDetails.ResourceMetadata = *req.Context.Requirements.FlowParams.ResourceMetadata.DeepCopy()
```
shlomitk1, line 533: Why is DeepCopy needed?

revit13: @shlomitk1 to avoid changing the values when it further processed.

shlomitk1: `req.DataDetails.ResourceMetadata = *req.Context.Requirements.FlowParams.ResourceMetadata` is enough, the structure is copied.

revit13: Fixed Thanks

shlomitk1 (approved):

PR #1462, opened 2022-05-02
Monitor changes of admin config policies and infrastructure attributes

shlomitk1:
Fixes https://github.com/fybrik/fybrik/issues/1010
- Upon changes in /tmp/adminconfig, manager recompiles the policies/ re-reads the infrastructure attributes if necessary. This is done in a separate thread, outside of the reconcile, thus mutex is used to prevent data corruption when reading and writing the same data.


roytman:
Before diving into the actual code review, a couple of questions:

- why do you create a proprietary polling mechanism, when there are goLang libraries for FS monitoring, which are based on OS events. e.g. [fsnotify](https://github.com/fsnotify/fsnotify) or [notify](https://github.com/rjeczalik/notify)
-  just to better understand, when adminconfig policies or infrastructure attributes are changed, `RegoPolicyEvaluator.Query` or `AttributeManager.Infrastructure` are updated, but we don't initiate the reconcile loop and have to wait for the next one, correct?
- what happen if as result of policies or infrastructure updates a new dataPath should be created, how user workload be updated?
- Where is an issue/design of the future?    


TODO: (unrelated to the PR) add a JSON admin interface to modify managers adminconfig policies, and infrastructure attributes.

roytman:
Before diving into the actual code review, a couple of questions:

- why do you create a proprietary polling mechanism, when there are goLang libraries for FS monitoring, which are based on OS events. e.g. [fsnotify](https://github.com/fsnotify/fsnotify) or [notify](https://github.com/rjeczalik/notify)
-  just to better understand, when adminconfig policies or infrastructure attributes are changed, `RegoPolicyEvaluator.Query` or `AttributeManager.Infrastructure` are updated, but we don't initiate the reconcile loop and have to wait for the next one, correct?
- what happen if as result of policies or infrastructure updates a new dataPath should be created, how user workload be updated?
- Where is an issue/design of the future?    


TODO: (unrelated to the PR) add a JSON admin interface to modify managers adminconfig policies, and infrastructure attributes.

shlomitk1:
> Before diving into the actual code review, a couple of questions:
> 
> * why do you create a proprietary polling mechanism, when there are goLang libraries for FS monitoring, which are based on OS events. e.g. [fsnotify](https://github.com/fsnotify/fsnotify) or [notify](https://github.com/rjeczalik/notify)
> * just to better understand, when adminconfig policies or infrastructure attributes are changed, `RegoPolicyEvaluator.Query` or `AttributeManager.Infrastructure` are updated, but we don't initiate the reconcile loop and have to wait for the next one, correct?
> * what happen if as result of policies or infrastructure updates a new dataPath should be created, how user workload be updated?

- Thanks, I'll look at the notify packages.
- The changes in policies do not affect the existing data plane and will be applied only to new data planes. This is currently by design.

roytman:
> The changes in policies do not affect the existing data plane and will be applied only to new data planes.

That makes sense, but it might lead to a situation when based on given policies and attributes manager created a data plane, after that, the polices/attributes were changed and now the manager creates completely different data plans. 
If we cannot link a data plane to the inputs that were used for its creation, troubleshooting the system can be challengeable.

shlomitk1:
> > The changes in policies do not affect the existing data plane and will be applied only to new data planes.
> 
> That makes sense, but it might lead to a situation when based on given policies and attributes manager created a data plane, after that, the polices/attributes were changed and now the manager creates completely different data plans. If we cannot link a data plane to the inputs that were used for its creation, troubleshooting the system can be challengeable.

Policy decisions are printed to the log. I see no problem in troubleshooting the system given the complete log.

pkg/monitor/filemonitor.go
```diff
@@ -0,0 +1,141 @@
...
+			changeDetected = true
+		}
+		if changeDetected {
+			m.Log.Info().Msg("File change detected, notifying...")
+			s.Entity.OnNotify()
+		}
+	}
+}
+
+func (m *FileMonitor) visit(s *Subscription) (int, time.Time, error) {
```
simanadler, line 101: Please add some documentation about what this function does

pkg/infrastructure/attribute_manager.go
```diff
@@ -32,14 +34,37 @@ const ValidationPath string = "/tmp/taxonomy/infraattributes.json#/definitions/I
...
+		Infrastructure: content,
+		Mux:            &sync.RWMutex{},
+	}, nil
+}
+
+func (m *AttributeManager) OnError(err error) {
+	m.Log.Error().Err(err).Msg("Error reading infrastructure attributes")
+}
+
+func (m *AttributeManager) GetOptions() monitor.FileMonitorOptions {
```
simanadler, line 56: Please add some documentation about what this does.  To what options does it refer?

pkg/infrastructure/attribute_manager.go
```diff
@@ -32,14 +34,37 @@ const ValidationPath string = "/tmp/taxonomy/infraattributes.json#/definitions/I
...
+	return monitor.FileMonitorOptions{Path: RegoPolicyDirectory, Extension: ".json"}
+}
+
+func (m *AttributeManager) OnNotify() {
```
simanadler, line 60: Is this called when the monitoring mechanism identifies a change?  Please add description

pkg/infrastructure/attribute_manager.go
```diff
@@ -32,14 +34,37 @@ const ValidationPath string = "/tmp/taxonomy/infraattributes.json#/definitions/I
...
+	content, err := readInfrastructure()
+	if err != nil {
+		m.OnError(err)
+	}
+	m.Mux.Lock()
+	m.Infrastructure = content
+	m.Mux.Unlock()
 }
 
 func readInfrastructure() (infraattributes.Infrastructure, error) {
```
simanadler, line 70: If I understand correctly, you are reading the infrastructure because the monitor identified a change ... and you are reading it in order to validate that the data structure is (still) valid.  Perhaps rename the function to validateInfrastructure?  and please add documentation about what it function does.

pkg/infrastructure/attribute_manager.go
```diff
@@ -76,8 +101,8 @@ func validateStructure(bytes []byte, taxonomySchema string) error {
 
 // GetAttribute returns an infrastructure attribute based on the attribute and instance names
 func (m *AttributeManager) GetAttribute(name taxonomy.Attribute, instance string) *taxonomy.InfrastructureElement {
-	for i, element := range m.Infrastructure.Items {
-		if element.Attribute == name && element.Instance == instance {
+	for i := range m.Infrastructure.Items {
```
simanadler, line 104: Why not assign to element in the for loop as you did originally?

manager/controllers/app/fybrikapplication_controller_unit_test.go
```diff
@@ -297,60 +296,6 @@ func TestDenyOnRead(t *testing.T) {
 	g.Expect(res).To(gomega.BeEquivalentTo(ctrl.Result{}), "Requests another reconcile")
 }
 
-// Tests selection of read-path module
```
simanadler, line 300: Why is this test being removed?  How is it related to this PR?

manager/controllers/app/fybrikapplication_controller_unit_test.go
```diff
@@ -420,67 +365,6 @@ func TestWrongCopyModule(t *testing.T) {
 	g.Expect(getErrorMessages(application)).NotTo(gomega.BeEmpty())
 }
 
-// Tests finding a module for copy supporting actions
```
simanadler, line 423: Same as previous question

simanadler (commented):
Perhaps add an end-to-end test showing that existing data planes are not updated when the infrastructure / admin policies change, but that new ones are?

pkg/infrastructure/attribute_manager.go
shlomitk1: go-lint complains...

pkg/monitor/filemonitor.go
shlomitk1: Done

pkg/infrastructure/attribute_manager.go
shlomitk1: Done

pkg/infrastructure/attribute_manager.go
shlomitk1: Done

pkg/infrastructure/attribute_manager.go
shlomitk1: Both reading the new information and storing in-memory and validation.
Added documentation.

manager/controllers/app/fybrikapplication_controller_unit_test.go
shlomitk1: Not related, it's just a redundant test... I'll take it out of this PR.

pkg/infrastructure/attribute_manager.go
simanadler: @roytman Can we fix golint to not complain?  This is making the code "uglier" and less efficient.

simanadler (approved):

pkg/infrastructure/attribute_manager.go
roytman: will check it

roytman:
> Thanks, I'll look at the notify packages.

I see that you don't use the notify packages, why?

pkg/infrastructure/attribute_manager.go
shlomitk1: I did some refactoring to make it look better

shlomitk1:
> > Thanks, I'll look at the notify packages.
> 
> I see that you don't use the notify packages, why? what's wrong with them?

because I used fsnotify instead :-) No special preference, seemed simpler to use and more mature.

pkg/infrastructure/attribute_manager.go
roytman: Of course, we can silence the linter here, but ....

roytman:
> > > Thanks, I'll look at the notify packages.
> > 
> > 
> > I see that you don't use the notify packages, why? what's wrong with them?
> 
> because I used fsnotify instead :-) No special preference, seemed simpler to use and more mature.

`fsnotify` is absolutely fine, I thought you use the original monitor ;-)
;כדמםאןכט;
כדמםאןכט

pkg/monitor/filemonitor.go
```diff
@@ -0,0 +1,144 @@
...
+					continue
+				}
+				m.Log.Err(err).Msg("error watching file changes")
+			}
+		}
+	}()
+}
+
+// Check if changes in files require any action
+func (m *FileMonitor) Monitor() {
```
roytman, line 75: why do we scan all files and validate their timestamps?  we can get the same info from the `watcher.Event.Name` and `watcher.Event.Op` . 
```go
// Event represents a single file system notification.
type Event struct {
	Name string // Relative path to the file or directory.
	Op   Op     // File operation that triggered the event.
}

// Op describes a set of file operations.
type Op uint32

// These are the generalized file operations that can trigger a notification.
const (
	Create Op = 1 << iota
	Write
	Remove
	Rename
	Chmod
)
```

shlomitk1: It won't work since the directory is not watched recursively.
This is the content of the watched directory:
<timestamp>
..data -> <timestamp>
*. json/rego -> ../data/*.json/rego

We are getting events about creating <new_timestamp>, deleting <old_timestamp> and modifying ..data folders. That's all. The actual files are inside <timestamp> directory and we won't know what has changed.

roytman: you're right that fsnotify doesn't support recursive watcher (it has been in the roadmap for a long time). 
However:

- if you get an event about directory creation (we can verify that a new file is a directory) you can add this directory to the root watcher, or create a new one.
- all these directories timestamps modification is done by the manager, co it can be changed or it can provide input to the monitor

You have implemented the monitor, but my question is - given fsnotify, can we simplify the entire process and prevent scanning all files?

roytman: Another comment, like you said, for each "update" we get several events, like create/delete <timestamp> and write in data folders and for each of them the `Monitor` method is run. 
Suggestion, use a predefined filename prefix for "timestamps dir name", and run `Monitor` only on 'REMOVE` events of files with this prefix.

shlomitk1: I am using fsnotify to avoid polling for changes. However, since all files are in the same directory, and we don't need to compile rego when json fles change, I still need to understand where the change was. 
I don't control the timestamps dir, it is created automatically when a configmap change is detected and the pod gets an update. 
The sequence of the events when running as a pod is different from running locally (for tests or debug) and depends on the internal implementation of the fsnotify. The more general approach is to react to all events. Also, I don't think that reading timestamps of a few files is that much of an overhead.

shlomitk1: Ignoring CHMOD event (others are required)
Added an integration test that demonstrates change in the policies (a new policy file is added) and how it affects the data plane.

roytman (approved):

PR #1469, opened 2022-05-09
Fix inconsistency in credential var that holds vault secret path

### LINKED ISSUE #1468: Inconsistency in `Credential` var that holds vault secret path
Credentials field  that holds vault secret path like in [api.go](https://github.com/fybrik/fybrik/blob/master/pkg/model/datacatalog/api.go) is expected to be of the following format: `/v1/kubernetes-secrets/bucket-creds?namespace=fybrik-system`
In `GetAssetResponse` it is assumed to be of the format as above while in other places like `CreateAssetRequest` and `GetAssetInfo` function call it is assumed to also contain the vault address. This should be fixed to have the credentials holding **only** the vault path without the vault address which is also consistent with the current format of `SecretPath` field in [vault_types.go](https://github.com/fybrik/fybrik/blob/master/manager/apis/app/v1alpha1/vault_types.go#L13)

revit13:
Closes #1468

Credentials field that holds vault secret path like in [api.go](https://github.com/fybrik/fybrik/blob/master/pkg/model/datacatalog/api.go) is expected to be of the following format: `/v1/kubernetes-secrets/bucket-creds?namespace=fybrik-system`.
In `GetAssetResponse` it is assumed to be of the format as above while in other places like `CreateAssetRequest` and `GetAssetInfo` function call it is also contain the vault address. This should be fixed to have the credentials holding only the vault path without the vault address which is also consistent with the current format of SecretPath field in [vault_types.go](https://github.com/fybrik/fybrik/blob/master/manager/apis/app/v1alpha1/vault_types.go#L13)


shlomitk1:
Are you assuming that all connectors should be aware of the vault address?

revit13:
@shlomitk1 yes that is the assumption. I will verify that the documentation includes that. Thanks

roytman (approved):

PR #1474, opened 2022-05-11
Vault readme

Mohammad-nassar10:


python/vault/README.md
```diff
@@ -0,0 +1,3 @@
+ ## Python Vault Package for Fybrik Components
+This versions of this library do not correspond to the
```
cdoron, line 2: "This versions" should be "The versions".
Please change both here and in the logging library. Thanks

cdoron (approved):

PR #1494, opened 2022-05-16
Using CSP solver to optimize data path

zivnevo:
This includes:
- Translating governance and admin constraints + optimization goals into a FlatZinc model, for a given data-path length
- Dumping the FlatZinc model into a file in /tmp
- Running a solver (its path is given by the env var `CSP_PATH`) on the dumped file
- If the solver returns UNSAT, retrying with a longer data path

The `pkg/optimizer` package includes:
- `optimizer.go` is the main entry point. The `Optimizer` object has a single method `Solve()` that returns a valid solution. This is where the model is built and given to the solver.
- `solution.go` defines the `Solution` object, which contains all data for building a Plotter
- `problem_data.go` defines the `DataInfo` and the `Environment` object which hold all data required for setting the optimization problem
- `datapath_csp.go` defines the `DataPathCSP` object, which does the actual translation from the problem data to a FlatZinc model, setting the required CSP variables and the constraints (for a given path length)
- `flatzinc_model.go` defines the `FlatZincModel` object which is used to incrementally build a FlatZinc model and then dump it to a file. It also supports reading a solver solution.


Makefile
```diff
@@ -65,6 +65,7 @@ test: export MODULES_NAMESPACE?=fybrik-blueprints
 test: export CONTROLLER_NAMESPACE?=fybrik-system
 test: pre-test
 	go test -v ./...
+	CSP_PATH=$(ABSTOOLBIN)/fzn-or-tools go test -v ./manager/controllers/app -count 1
```
shlomitk1, line 68: fzn-or-tools target should be stated as a dependency of pre-test

manager/controllers/utils/configuration.go
```diff
@@ -26,7 +26,7 @@ const (
 	PrettyLoggingKey                  string = "PRETTY_LOGGING"
 	CatalogProviderNameKey            string = "CATALOG_PROVIDER_NAME"
 	DatapathLimitKey                  string = "DATAPATH_LIMIT"
-	UseCSPKey                         string = "USE_CSP"
+	CSPPathKey                        string = "CSP_PATH"
```
shlomitk1, line 29: I don't think the path should be configurable in the deployment.

pkg/optimizer/optimizer_test.go
```diff
@@ -0,0 +1,28 @@
...
+	env := getTestEnv()
+	opt := NewOptimizer(env, getDataInfo(env), os.Getenv("ABSTOOLBIN")+"/fzn-or-tools")
+	solution, err := opt.Solve()
+	if err != nil {
+		t.Fatalf("Failed solving constraint problem: %s", err)
+	}
+
+	solutionLen := len(solution.DataPath)
+	if solutionLen < 1 {
+		t.Error("Solution is too short")
```
shlomitk1, line 21: Doesn't it mean that there is no solution at all?

manager/controllers/app/solve_test.go
```diff
@@ -206,15 +206,15 @@ func TestReadWithTransforms(t *testing.T) {
 	account := &v1alpha1.FybrikStorageAccount{}
 	g.Expect(readObjectFromFile("../../testdata/unittests/account-theshire.yaml", account)).NotTo(gomega.HaveOccurred())
 	addCluster(env, multicluster.Cluster{Metadata: multicluster.ClusterMetadata{Region: string(account.Spec.Region)}})
-	p := PathBuilder{Log: &testLog, Env: env, Asset: createReadRequest()}
-	p.Asset.Actions = []taxonomy.Action{{Name: "RedactAction"}}
-	_, err := p.solve()
+	asset := createReadRequest()
+	asset.Actions = []taxonomy.Action{{Name: "RedactAction"}}
+	_, err := solve(env, asset, &testLog)
```
shlomitk1, line 211: I wonder if instead of running tests twice in `make test` we could run different solvers in each test of solve_test.go and compare the Solution objects in addition to the existing checks. We don't really want to re-test the behavior of reconcilers, just the solver.

PR #1496, opened 2022-05-17
Set deny status condition when no storage is available for writing new data

shlomitk1:
Fixes https://github.com/fybrik/fybrik/issues/1493

When no storage is available for writing, and an explicit copy is required (copy or write a new asset), Deny is reported in the status.


revit13 (approved):

revit13:
@shlomitk1 Thanks for the fix.

Issue #1478, opened 2022-05-12
Optimizer: consider cluster-to-cluster optimization features


zivnevo:
e.g., Bandwidth

Decided to sum their values from last stored data-point to workload.

shlomitk1:
@zivnevo This issue has been implemented and can be closed, correct?

zivnevo:
Done

Issue #1479, opened 2022-05-12
Optimizer: consider several data-path lengths


zivnevo:
It might be that longer-than-minimal data-paths are having a better score (considering optimization goals)

PR #1497, opened 2022-05-17
csp-based data-path generation

### LINKED ISSUE #1133: Optimizer Step 1 - Find legal data path
Defining a data path from dataset to workload, consisting of configured modules. The path should satisfy hard constraints:
• Each required action is handled by a module on the list
• Interfaces match both workload and dataset specs as well as between neighboring modules 
• The list of modules satisfies capability restrictions(not including cluster allocation)

Leverage a CSP approach as described on slides 7 and 8 in Options and Proposed Approach - https://drive.google.com/file/d/1wmU07hwZpTY3FbJaziZhYZEw21si0QHw/view?usp=sharing

### LINKED ISSUE #1134: Optimizer Step 2 - Allocate Modules to Clusters
Find a legal and optimal allocation of modules to clusters for data paths from Optimizer Step 1

Approach is described on slide #10 of Options and Proposed Approach - https://drive.google.com/file/d/1wmU07hwZpTY3FbJaziZhYZEw21si0QHw/view?usp=sharing

zivnevo:
Resolves #1133
Resolves #1134 (except for optimization goals over 2-parties attributes, will be handled in issue #1478)

This includes:
- Translating governance and admin constraints + optimization goals into a FlatZinc model, for a given data-path length
- Dumping the FlatZinc model into a file in /tmp
- Running a solver (its path is given by the env var `CSP_PATH`) on the dumped file
- If the solver returns UNSAT, retrying with a longer data path

The `pkg/optimizer` package includes:
- `optimizer.go` is the main entry point. The `Optimizer` object has a single method `Solve()` that returns a valid solution. This is where the model is built and given to the solver.
- `solution.go` defines the `Solution` object, which contains all data for building a Plotter
- `problem_data.go` defines the `DataInfo` and the `Environment` object which hold all data required for setting the optimization problem
- `datapath_csp.go` defines the `DataPathCSP` object, which does the actual translation from the problem data to a FlatZinc model, setting the required CSP variables and the constraints (for a given path length)
- `flatzinc_model.go` defines the `FlatZincModel` object which is used to incrementally build a FlatZinc model and then dump it to a file. It also supports reading a solver solution.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,7 @@ test: export MODULES_NAMESPACE?=fybrik-blueprints
 test: export CONTROLLER_NAMESPACE?=fybrik-system
 test: pre-test
 	go test -v ./...
+	CSP_PATH=$(ABSTOOLBIN)/fzn-or-tools go test -v ./manager/controllers/app -count 1
 
 .PHONY: run-integration-tests
 run-integration-tests: export DOCKER_HOSTNAME?=localhost:5000
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ deploy: $(TOOLBIN)/kubectl $(TOOLBIN)/helm
                --namespace $(KUBE_NAMESPACE) --wait --timeout 120s
 
 .PHONY: pre-test
-pre-test: generate manifests $(TOOLBIN)/etcd $(TOOLBIN)/kube-apiserver
+pre-test: generate manifests $(TOOLBIN)/etcd $(TOOLBIN)/kube-apiserver $(TOOLBIN)/fzn-or-tools
 	mkdir -p /tmp/taxonomy
 	mkdir -p /tmp/adminconfig
 	cp charts/fybrik/files/taxonomy/*.json /tmp/taxonomy/
--- a/Makefile
+++ b/Makefile
@@ -63,9 +63,10 @@ pre-test: generate manifests $(TOOLBIN)/etcd $(TOOLBIN)/kube-apiserver $(TOOLBIN
 .PHONY: test
 test: export MODULES_NAMESPACE?=fybrik-blueprints
 test: export CONTROLLER_NAMESPACE?=fybrik-system
+test: export CSP_PATH=$(ABSTOOLBIN)/fzn-or-tools
 test: pre-test
 	go test -v ./...
-	USE_CSP=true CSP_PATH=$(ABSTOOLBIN)/fzn-or-tools go test -v ./manager/controllers/app -count 1
+	USE_CSP=true go test -v ./manager/controllers/app -count 1
 
 .PHONY: run-integration-tests
 run-integration-tests: export DOCKER_HOSTNAME?=localhost:5000
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ test: export MODULES_NAMESPACE?=fybrik-blueprints
 test: export CONTROLLER_NAMESPACE?=fybrik-system
 test: pre-test
 	go test -v ./...
-	CSP_PATH=$(ABSTOOLBIN)/fzn-or-tools go test -v ./manager/controllers/app -count 1
+	USE_CSP=true CSP_PATH=$(ABSTOOLBIN)/fzn-or-tools go test -v ./manager/controllers/app -count 1
 
 .PHONY: run-integration-tests
 run-integration-tests: export DOCKER_HOSTNAME?=localhost:5000
--- a/manager/controllers/app/select_modules.go
+++ b/manager/controllers/app/select_modules.go
@@ -65,11 +65,7 @@ func (p *PathBuilder) FindPaths() []optimizer.Solution {
 		solutions = p.findPathsWithinLimit(nodeFromAppRequirements, nodeFromAssetMetadata, bound)
 		// reverse each solution to start with the application requirements, e.g. workload
 		for ind := range solutions {
-			for elementInd := 0; elementInd < len(solutions[ind].DataPath)/2; elementInd++ {
-				reversedInd := len(solutions[ind].DataPath) - elementInd - 1
-				solutions[ind].DataPath[elementInd], solutions[ind].DataPath[reversedInd] =
-					solutions[ind].DataPath[reversedInd], solutions[ind].DataPath[elementInd]
-			}
+			solutions[ind].Reverse()
 		}
 	}
 	// get valid solutions by extending data paths with transformations and selecting an appropriate cluster for each capability
--- a/manager/controllers/app/select_modules.go
+++ b/manager/controllers/app/select_modules.go
@@ -4,7 +4,6 @@
 package app
 
 import (
-	"reflect"
 	"strconv"
 	"strings"
 
@@ -17,6 +16,7 @@ import (
 	"fybrik.io/fybrik/pkg/logging"
 	"fybrik.io/fybrik/pkg/model/taxonomy"
 	"fybrik.io/fybrik/pkg/multicluster"
+	"fybrik.io/fybrik/pkg/optimizer"
 )
 
 // Temporary hard-coded capability representing Actions of read/copy capability.
@@ -26,12 +26,12 @@ const Transform = "transform"
 // component responsible for data path construction
 type PathBuilder struct {
 	Log   *zerolog.Logger
-	Env   *Environment
-	Asset *DataInfo
+	Env   *optimizer.Environment
+	Asset *optimizer.DataInfo
 }
 
 // find a solution for data plane orchestration
-func (p *PathBuilder) solve() (Solution, error) {
+func (p *PathBuilder) solve() (optimizer.Solution, error) {
 	p.Log.Trace().Str(logging.DATASETID, p.Asset.Context.DataSetID).Msg("Choose modules for dataset")
 	solutions := p.FindPaths()
 	// No data path found for the asset
@@ -40,7 +40,7 @@ func (p *PathBuilder) solve() (Solution, error) {
 		p.Log.Error().Str(logging.DATASETID, p.Asset.Context.DataSetID).Msg(msg)
 		logging.LogStructure("Data Item Context", p.Asset, p.Log, zerolog.TraceLevel, true, true)
 		logging.LogStructure("Module Map", p.Env.Modules, p.Log, zerolog.TraceLevel, true, true)
-		return Solution{}, errors.New(msg + " for " + p.Asset.Context.DataSetID)
+		return optimizer.Solution{}, errors.New(msg + " for " + p.Asset.Context.DataSetID)
 	}
 	return solutions[0], nil
 }
@@ -49,7 +49,7 @@ func (p *PathBuilder) solve() (Solution, error) {
 // First, data paths are constructed using interface connections, starting from data source.
 // Then, transformations are added to the found paths, and clusters are matched to satisfy restrictions from admin config policies.
 // Optimization is done by the shortest path (the paths are sorted by the length). To be changed in future versions.
-func (p *PathBuilder) FindPaths() []Solution {
+func (p *PathBuilder) FindPaths() []optimizer.Solution {
 	nodeFromAssetMetadata := p.getAssetConnectionNode()
 	nodeFromAppRequirements := p.getRequiredConnectionNode()
 
@@ -58,7 +58,7 @@ func (p *PathBuilder) FindPaths() []Solution {
 	if err != nil {
 		p.Log.Warn().Str(logging.DATASETID, p.Asset.Context.DataSetID).Msg("a default value for DATAPATH_LIMIT will be used")
 	}
-	var solutions []Solution
+	var solutions []optimizer.Solution
 	if p.Asset.Context.Flow != taxonomy.WriteFlow {
 		solutions = p.findPathsWithinLimit(nodeFromAssetMetadata, nodeFromAppRequirements, bound)
 	} else {
@@ -79,8 +79,8 @@ func (p *PathBuilder) FindPaths() []Solution {
 }
 
 // extend the received data paths with transformations and select an appropriate cluster for each capability in a data path
-func (p *PathBuilder) validSolutions(solutions []Solution) []Solution {
-	validPaths := []Solution{}
+func (p *PathBuilder) validSolutions(solutions []optimizer.Solution) []optimizer.Solution {
+	validPaths := []optimizer.Solution{}
 	for ind := range solutions {
 		if p.validate(solutions[ind]) {
 			validPaths = append(validPaths, solutions[ind])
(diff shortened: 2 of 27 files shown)


Makefile
```diff
@@ -65,6 +65,7 @@ test: export MODULES_NAMESPACE?=fybrik-blueprints
 test: export CONTROLLER_NAMESPACE?=fybrik-system
 test: pre-test
 	go test -v ./...
+	CSP_PATH=$(ABSTOOLBIN)/fzn-or-tools go test -v ./manager/controllers/app -count 1
```
shlomitk1, line 68: Add the solver installation target as a dependency to the test target. If I run `make test` without installing the solver, I want it to be installed automatically.

manager/controllers/app/solve.go
```diff
@@ -4,80 +4,20 @@
...
-}
-
 // find a solution for a data path
 // satisfying governance and admin policies
 // with respect to the optimization strategy
-func solve(env *Environment, datasetInfo *DataInfo, log *zerolog.Logger) (Solution, error) {
-	if utils.UseCSP() {
-		return Solution{}, errors.New("CSP solution is not yet implemented")
+func solve(env *optimizer.Environment, datasetInfo *optimizer.DataInfo, log *zerolog.Logger) (optimizer.Solution, error) {
+	cspPath := utils.GetCSPPath()
```
shlomitk1, line 17: Perhaps it would be better to have two variables: one for path, another - a boolean useCsp. In the deployment we'd probably want to propagate the boolean one to the configmap while making the path configured automatically to the location we decide within the deployment. In this way, the user/admin will have only to switch the boolean field without touching the path field. 
The check here would then compare both path length to a positive number and the boolean flag to true before running the solver.

pkg/optimizer/optimizer.go
```diff
@@ -0,0 +1,67 @@
...
+package optimizer
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+)
+
+const (
+	MaxDataPathDepth = 4
```
shlomitk1, line 13: We have an env variable configuring the max path length.

pkg/optimizer/optimizer.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return "", err
+	}
+	return string(solverSolution), nil
+}
+
+// The main method to call for finding a legal and optimal data path
+func (opt *Optimizer) Solve() (Solution, error) {
+	for pathLen := 1; pathLen <= MaxDataPathDepth; pathLen++ {
+		solverSolution, err := opt.getSolution(pathLen)
```
shlomitk1, line 53: Will it stop at the minimal length of the data path? What if copy+read is better than just read but both are legal?

pkg/optimizer/optimizer_test.go
```diff
@@ -0,0 +1,28 @@
...
+	env := getTestEnv()
+	opt := NewOptimizer(env, getDataInfo(env), os.Getenv("ABSTOOLBIN")+"/fzn-or-tools")
+	solution, err := opt.Solve()
+	if err != nil {
+		t.Fatalf("Failed solving constraint problem: %s", err)
+	}
+
+	solutionLen := len(solution.DataPath)
+	if solutionLen < 1 {
+		t.Error("Solution is too short")
```
shlomitk1, line 21: Does this mean that the solution does not exist?

pkg/optimizer/optimizer.go
zivnevo: This will be handled later in issue #1479

manager/controllers/app/solve.go
simanadler: @shlomitk1 I agree.  I was about to make the same comment.

simanadler:
Please add the issue # that this PR addresses.  If it only addresses part of the issue please call out what is and is not included in the PR.

pkg/optimizer/datapath_csp.go
```diff
@@ -0,0 +1,700 @@
...
+			if !modcapSupportsIntfcSink {
+				dpc.preventAssignments([]string{modCapVarname, sinkIntfcVarname}, []int{modCapIdx + 1, intfcIdx}, pathLength)
+			}
+		}
+	}
+}
+
+// If there are optimization goals set, defines appropriate variables and sets the CSP-solver optimization goal
+// Otherwise, just sets the CSP-solver goal as "satisfy"
+func (dpc *DataPathCSP) addOptimizationGoals(pathLength int) error {
```
simanadler, line 460: What in the current code prevents us from optimizing for more than one optimization goal?

pkg/optimizer/datapath_csp.go
```diff
@@ -0,0 +1,700 @@
...
+			Edge:           edge,
+			Actions:        dpc.getSolutionActionsAtPos(solverSolution, pathPos),
+			Cluster:        dpc.env.Clusters[clusterIdx-1].Name,
+			StorageAccount: sa,
+		}
+		solution.DataPath = append(solution.DataPath, &resolvedEdge)
+		srcNode = sinkNode
+	}
+
+	if dpc.problemData.Context.Flow == taxonomy.WriteFlow { // reverse solution
```
simanadler, line 650: I believe this same functionality is implemented in the non-csp path builder.  Perhaps create a common function?

pkg/optimizer/optimizer.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+		solution, err := opt.dpc.decodeSolverSolution(solverSolution, pathLen)
+		if err != nil {
+			return Solution{}, err
+		}
+		if len(solution.DataPath) > 0 {
+			return solution, nil
+		}
+	}
+	msg := "Data path cannot be constructed given the deployed modules and the active restrictions"
```
simanadler, line 65: We will need to write to the log file the context (modules, restrictions, etc) in addition to this message.

Makefile
simanadler: Do we want to add separate deploy with and without csp?

simanadler (commented):
Great progress Ziv.  Very exciting to see this PR!

pkg/optimizer/problem_data.go
```diff
@@ -0,0 +1,38 @@
+// Copyright 2020 IBM Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+package optimizer
```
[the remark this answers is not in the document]
zivnevo: We cannot place this code in `manager/controllers/app` because it will create a circular dependency: optimizer will have to import `app` package to use this code, and `app` package will have to import `optimizer` package to use the optimizer.
We can place this code in a package of its own. @shlomitk1 , what do you think?

pkg/adminconfig/model.go
```diff
@@ -94,3 +100,82 @@ type EvaluationOutputStructure struct {
 	// +optional
 	Optimize []OptimizationStrategy `json:"optimize,omitempty"`
 }
+
+// Validation of an object with respect to the admin config restriction
```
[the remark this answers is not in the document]
zivnevo: This code is used by both the optimizer and the existing implementation. I didn't want to duplicate it. Do you have a suggestion for a better place to put it?

zivnevo:
> Please add the issue # that this PR addresses. If it only addresses part of the issue please call out what is and is not included in the PR.

Done

pkg/optimizer/datapath_csp.go
```diff
@@ -0,0 +1,700 @@
...
+	clusterVarname   = "moduleCluster"         // Var's value says which of the available clusters to use (on each Edge)
+	saVarname        = "storageAccount"        // Var's value says which storage account to use (0 means no sa)
+	srcIntfcVarname  = "moduleSourceInterface" // Var's value says which interface to use as source
+	sinkIntfcVarname = "moduleSinkInterface"   // Var's value says which interface to use as sink
+	actionVarname    = "action_%s"             // Vars for each required action, say whether the action was applied
+	jointGoalVarname = "jointGoal"             // Var's value indicates the quality of the data path w.r.t. optimization goals
+)
+
+// Couples together a module and one of its capabilities
+type moduleAndCapability struct {
```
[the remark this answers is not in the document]
zivnevo: Note that this struct only holds a single capability of a given module. It is mainly used to serialize all (allowed) module capabilities in one array. I'm not sure if the other flags it holds are meaningful somewhere else. Maybe @shlomitk1 can comment.

pkg/optimizer/datapath_csp.go
```diff
@@ -0,0 +1,700 @@
...
+	module        *appApi.FybrikModule
+	capability    *appApi.ModuleCapability
+	capabilityIdx int  // The index of capability in module's spec
```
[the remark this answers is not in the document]
zivnevo: This struct is recreated on every call to `Optimizer.Solve()`.

pkg/optimizer/datapath_csp.go
zivnevo: Do you mean "bandwidth-like" optimization goals?
Such goals will return an empty InstanceType when calling `AttributeManager.GetInstanceType(attr)` and will be ignored.

pkg/optimizer/problem_data.go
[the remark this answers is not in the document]
shlomitk1: Perhaps a package called `datapath` ?

Makefile
shlomitk1: @simanadler This target refers to testing manager when it runs locally. 
As for the deployment, I think we'll deploy with csp but run twice: with or without its usage (passing the env var in the command line).

zivnevo: Added `fzn-or-tools` as a dependency of `pre-test`

manager/controllers/app/solve.go
zivnevo: Added a new environment variable: `USE_CSP`. Should be set to true for CSP solver to be used.

pkg/optimizer/optimizer.go
zivnevo: Do we want to use the same env variable? Currently its default is 2. Is that enough?

pkg/optimizer/optimizer_test.go
zivnevo: Fixed

manager/controllers/utils/configuration.go
```diff
@@ -69,9 +69,9 @@ func GetDataPathMaxSize() (int, error) {
 	return limit, nil
 }
 
-// UseCSP returns a boolean indicator for using a CSP solver when generating a plotter
-func UseCSP() bool {
-	return os.Getenv(UseCSPKey) == "true"
+// GetCSPPath returns the path of the CSP solver to use when generating a plotter, or "" if no CSP solver should be used
+func GetCSPPath() string {
```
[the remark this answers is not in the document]
zivnevo: Done

pkg/optimizer/datapath_csp.go
```diff
@@ -0,0 +1,700 @@
...
+	"strings"
+
+	appApi "fybrik.io/fybrik/manager/apis/app/v1alpha1"
+	"fybrik.io/fybrik/pkg/adminconfig"
+	"fybrik.io/fybrik/pkg/model/datacatalog"
+	"fybrik.io/fybrik/pkg/model/taxonomy"
+	"fybrik.io/fybrik/pkg/multicluster"
+)
+
+// Names of the primary variables on which we need to take decisions
```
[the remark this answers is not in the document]
zivnevo: Fixed

pkg/optimizer/datapath_csp.go
```diff
@@ -0,0 +1,700 @@
...
+	capability    *appApi.ModuleCapability
+	capabilityIdx int  // The index of capability in module's spec
+	virtualSource bool // whether data is consumed via API
+	virtualSink   bool // whether data is transferred in memory via API
+	hasSource     bool // whether module-capability has source interfaces
+	hasSink       bool // whether module-capability has sink interfaces
+}
+
+// The main class for producing a CSP from data-path constraints and for decoding solver's solutions
+type DataPathCSP struct {
```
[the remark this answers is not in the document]
zivnevo: Done. A package header was added to `optimizer.go`

pkg/optimizer/optimizer.go
zivnevo: Done

pkg/optimizer/optimizer_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"testing"
+
+	"fybrik.io/fybrik/pkg/logging"
+)
+
+var testLog = logging.LogInit("Optimizer", "Test")
+
+func TestOptimizer(t *testing.T) {
+	env := getTestEnv()
+	opt := NewOptimizer(env, getDataInfo(env), os.Getenv("ABSTOOLBIN")+"/fzn-or-tools", &testLog)
```
shlomitk1, line 17: Better to use CSP_PATH.

manager/controllers/app/select_modules.go
```diff
@@ -40,7 +40,7 @@ func (p *PathBuilder) solve() (Solution, error) {
 		p.Log.Error().Str(logging.DATASETID, p.Asset.Context.DataSetID).Msg(msg)
```
shlomitk1, line 40: These log printings can be done inside some util function shared by this implementation and the solver.

shlomitk1 (commented):
Looks very good. I've left some minor comments.

shlomitk1 (approved):

pkg/optimizer/optimizer.go
shlomitk1: Let's leave it for now and come back to it later after you implement optimization on different path lengths.

pkg/optimizer/optimizer_test.go
zivnevo: Fixed (should have been `solutionLen < 2`)

manager/controllers/utils/configuration.go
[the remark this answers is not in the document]
zivnevo: done

manager/controllers/app/solve.go
zivnevo: Added an additional boolean env-var, `USE_CSP`.

pkg/optimizer/problem_data.go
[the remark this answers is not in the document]
zivnevo: Moved the code to a new package called `datapath`

pkg/optimizer/solution.go
```diff
@@ -0,0 +1,46 @@
+// Copyright 2020 IBM Corp.
+// SPDX-License-Identifier: Apache-2.0
+
+package optimizer
```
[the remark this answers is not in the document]
zivnevo: Moved the code to a new package called `datapath`

pkg/optimizer/datapath_csp.go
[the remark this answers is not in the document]
zivnevo: Added package header with the link in `optimizer.go`

pkg/optimizer/datapath_csp.go
zivnevo: Common code moved to `solution.go`

pkg/optimizer/optimizer.go
zivnevo: done

pkg/optimizer/optimizer_test.go
zivnevo: Done

simanadler (approved):

PR #1501, opened 2022-05-24
bump base image to ubi-minimal:8.6

roytman:
As one of the requirements from https://github.com/fybrik/fybrik/issues/1480, the base ubunty image was updated.


revit13:
@roytman should the tag be `8.6` instead of `8.6-751`? this way we do not have to bump for minor version releases.

roytman:
> @roytman should the tag be `8.6` instead of `8.6-751`? this way we do not have to bump for minor version releases.

good point, will fix

revit13 (approved):

PR #1508, opened 2022-05-29
Deploy solver

shlomitk1:
Fixes https://github.com/fybrik/fybrik/issues/1260
A new section was added to manager values yaml:
  ```
# CSP solver for data plane optimization
  solver:
    # image of the container with solver binary and libs
    # when specified, the solver will be deployed in the manager pod
    image: "optimizer"
    imagePullPolicy: "Always"
    # Set to true to enable the use of the solver by Fybrik
    enabled: false

```
If an image including solver binary and libs is specified, an init container runs that copies solver files from the image to the manager container.

### CODE REVIEW

File: charts/fybrik/values.yaml [hunk 1]
```diff
@@ -191,6 +191,15 @@ manager:
...
   prometheus: false
 
+  # CSP solver for data plane optimization
+  solver:
+    # image of the container with solver binary and libs
+    # when specified, the solver will be deployed in the manager pod
+    image: "optimizer"
+    imagePullPolicy: "Always"
+    # Set to true to enable the use of the solver by Fybrik
+    enabled: false
```
simanadler, line 201: The default is to deploy the solver but to have it turned off?  Shouldn't the default be to have it turned on?

pkg/optimizer/optimizer.go
```diff
@@ -45,18 +46,20 @@ func NewOptimizer(env *datapath.Environment, problemData *datapath.DataInfo, sol
 }
 
 func (opt *Optimizer) getSolution(pathLength int) (string, error) {
+	opt.log.Debug().Msgf("finding solution of length %d", pathLength)
```
simanadler, line 49: Why not use the standardized fybrik logging?  Same comment for other logging calls in this file.

simanadler (commented):
See comments

pkg/optimizer/optimizer.go
shlomitk1: It is standardized. "log" is a member of Optimizer which comes from "github.com/rs/zerolog"
This printout will look like
`{"level":"debug","Controller":"FybrikApplication","FybrikApplication":"default/test-app-prod","app.fybrik.io/app-uuid":"b848eb60-8d2a-4d58-89b3-9d3e8197978b","time":"2022-05-30T06:57:27Z","caller":"/data/ws/the-mesh-for-data/pkg/optimizer/optimizer.go:49","message":"finding solution of length 1"}`

charts/fybrik/values.yaml
shlomitk1: I still consider this as an experimental release. If it goes into 0.7 it should definitely be "false". Later we'll change it to "true", after more runs and features.

simanadler: OK.  Makes sense

pkg/optimizer/Makefile
```diff
@@ -0,0 +1,22 @@
...
+
+# Overwrite docker-build from docker.mk
+.PHONY: docker-build
+docker-build: 	
+	mkdir solver-tools
+	mkdir solver-tools/bin
+	mkdir solver-tools/lib
+	cp ${ABSTOOLBIN}/fzn-or-tools solver-tools/bin/solver
+	cp ${ABSTOOLBIN}/../lib/libfz.so solver-tools/lib/
+	cp ${ABSTOOLBIN}/../lib/libortools.so solver-tools/lib/
```
zivnevo, line 20:
```suggestion
	cp ${ABSTOOLBIN}/fzn-or-tools solver-tools/bin/solver
	cp -r ${ABSTOOLBIN}/../lib solver-tools
```
It is not guaranteed that in the coming versions of `fzn-or-tools` we'll have exactly these 2 libraries (in fact, the next release seems to have a different name for these libraries)

pkg/optimizer/Makefile
```diff
@@ -0,0 +1,22 @@
...
+KUBE_NAMESPACE ?= fybrik-system
+CONTROLLER_NAMESPACE ?= ${KUBE_NAMESPACE}
+
+include $(ROOT_DIR)/hack/make-rules/tools.mk
+include $(ROOT_DIR)/hack/make-rules/docker.mk
+include $(ROOT_DIR)/hack/make-rules/verify.mk
+
+# Overwrite docker-build from docker.mk
+.PHONY: docker-build
+docker-build:
```
zivnevo, line 14: Don't we also need a rule for `docker-push`?

charts/fybrik/templates/manager-deployment.yaml
```diff
@@ -34,6 +34,20 @@ spec:
       serviceAccountName: {{ .Values.manager.serviceAccount.name }}
       securityContext:
         {{- toYaml .Values.manager.podSecurityContext | nindent 8 }}
+      {{- if .Values.manager.solver.image }}
```
zivnevo, line 37:
```suggestion
      {{- if .Values.manager.solver.enabled}}
```
It's the `enabled` flag we want to depend on, isn't it?
Note that this `if` statement appears 3 more times

zivnevo (changes requested):
Very neat and :cool:
See some small comments below

charts/fybrik/templates/manager-deployment.yaml
shlomitk1: We want to deploy once and be able to switch usage on the fly by changing the configmap manually.

pkg/optimizer/Makefile
shlomitk1: ok, will change, thanks.

zivnevo (approved):

pkg/optimizer/optimizer.go
simanadler: OK

roytman:
converted to draft to prevent merging it into v0.7.

PR #1512, opened 2022-05-31
Use `/data` as data root directory instead of `/tmp`.

revit13:
This PR does the following:
- changes the data root directory to `/data`  instead of `/tmp`.
`taxonomy` and `adminconfig` are now sub directories in `/data`.
- Adds `/tmp` as mount path which allows changing containerSecurityContext .readOnlyRootFilesystem field in the manager deployment to true.


pkg/adminconfig/compile.go
```diff
@@ -9,13 +9,15 @@ import (
...
 	"github.com/open-policy-agent/opa/ast"
 	"github.com/open-policy-agent/opa/rego"
 	"github.com/pkg/errors"
 )
 
-// A directory containing rego files that define admin config policies
-const RegoPolicyDirectory string = "/tmp/adminconfig/"
+// RegoPolicyDirectory is a directory containing rego files that
+// define admin config policies
+var RegoPolicyDirectory = utils.DataRootDir + "/adminconfig"
```
shlomitk1, line 20: add "/" at the end

revit13: Fixed. Thanks

shlomitk1 (commented):
I suggest to use $HOME/data instead of /data.
A change to the manager Dockerfile should be made as well.

revit13:
> I suggest to use $HOME/data (or just $HOME) instead of /data.
> A change to the manager Dockerfile should be made as well.

Can you please explain? not sure I understand what should be the values of `$HOME` in Dockerfile.

shlomitk1:
> > I suggest to use $HOME/data (or just $HOME) instead of /data.
> > A change to the manager Dockerfile should be made as well.
> 
> Can you please explain? not sure I understand what should be the value of `$HOME` in Dockerfile.

1. It may be a problem to access /data when running locally, thus suggesting $HOME.
2. In manager dockerfile $HOME is set to /tmp. With allocating a volume /tmp, I suggest to rename HOME to /data.

revit13:
@shlomitk1 following your suggestion and discussion with @roytman I made the data directory to be in `$DATA` env var where it is set to `/data` when running in k8s and to `/tmp` when running locally.

charts/fybrik/templates/manager-deployment.yaml
```diff
@@ -135,9 +135,11 @@ spec:
...
-            - mountPath: /tmp/taxonomy
+            - name: data
+              mountPath: /data
+            - mountPath: {{ include "fybrik.printDataSubdir" ( tuple "taxonomy" ) }}
               name: fybrik-taxonomy
-            - mountPath: /tmp/adminconfig
+            - mountPath: {{ include "fybrik.printDataSubdir" ( tuple "adminconfig" ) }}
               name: fybrik-adminconfig
             {{- if .Values.manager.solver.image }}
             - mountPath: /tmp/solver
```
shlomitk1, line 145: also needs to be changed

revit13: Thanks, I changed the mount dir of the solver also.

charts/fybrik/templates/_helpers.tpl
```diff
@@ -147,3 +147,20 @@ Relevant only when .Values.chartsPersistentVolumeClaim is set.
 {{- define "fybrik.localChartsMountPath" }}
 {{- printf "/opt/fybrik" }}
 {{- end }}
+
+{{/*
+Print Data directory.
+*/}}
+{{- define "fybrik.printDataDir" -}}
```
roytman, line 154: maybe fybrik.getDataDir ?

charts/fybrik/templates/_helpers.tpl
```diff
@@ -147,3 +147,20 @@ Relevant only when .Values.chartsPersistentVolumeClaim is set.
...
+/data
+{{- end }}
+
+{{/*
+Print sub directory in /data directory. The sub directory is
+passed as parameter to the function.
+*/}}
+{{- define "fybrik.printDataSubdir" -}}
```
roytman, line 162: getDataSubdir ?

charts/fybrik/templates/manager-deployment.yaml
```diff
@@ -42,11 +42,11 @@ spec:
           command:
           - cp 
           - -rf
-          - "/data/tools"
-          - "/tmp/solver/"
+          - {{ include "fybrik.printDataDir" . }}/tools
+          - {{ include "fybrik.printDataSubdir" ( tuple "solver" ) }}/
```
roytman, line 46: so solver dir will be `/data/solver` instead of `/tmp/solver`. is it correct ?

manager/apis/app/v1alpha1/utils/configuration.go
```diff
@@ -0,0 +1,13 @@
...
+
+package utils
+
+import (
+	"os"
+)
+
+// GetDataDir returns the directory where the data resides.
+func GetDataDir() string {
+	return os.Getenv("DATA")
```
roytman, line 12: should we add here some default, e.g `/tmp` ?

manager/main.go
```diff
@@ -187,7 +187,7 @@ func run(namespace string, metricsAddr string, enableLeaderElection bool,
 			return 1
 		}
 		defer watcher.Close()
-		// watch /tmp/adminconfig directory for changes
+		// watch /data/adminconfig directory for changes
```
roytman, line 190: $DATA/adminconfig ? or GetDataDir() + "/adminconfig"

manager/apis/app/v1alpha1/utils/configuration.go
revit13: nor sure about that as if the manager runs inside Kubernetes the default value is /data while locally its /tmp.

charts/fybrik/templates/_helpers.tpl
revit13: I changed that.

charts/fybrik/templates/_helpers.tpl
revit13: I changed that.

charts/fybrik/templates/manager-deployment.yaml
revit13: yes its correct.

manager/main.go
revit13: I changed that.

Makefile
```diff
@@ -1,6 +1,7 @@
 include Makefile.env
 export DOCKER_TAGNAME ?= 0.0.0
 export KUBE_NAMESPACE ?= fybrik-system
+export DATA ?= /tmp
```
shlomitk1, line 4: maybe DATA_DIR?

charts/fybrik/templates/manager-deployment.yaml
```diff
@@ -42,11 +42,11 @@ spec:
           command:
           - cp 
           - -rf
-          - "/data/tools"
-          - "/tmp/solver/"
+          - {{ include "fybrik.getDataDir" . }}/tools
```
shlomitk1, line 45: Leave "/data/tools" - this is a directory inside the solver container

manager/Dockerfile
```diff
@@ -3,7 +3,8 @@
 
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
 ENV HOME=/tmp
-WORKDIR /tmp
+ENV DATA=/data
```
shlomitk1, line 6: I think it should be configured in the deployment yaml

manager/apis/app/v1alpha1/utils/configuration.go
revit13: OK, where this file should reside? does `pkg/environment/` the right place? Thanks

Makefile
revit13: makes sense. I will change it. Thanks

manager/apis/app/v1alpha1/utils/configuration.go
shlomitk1: Yes, as long as we don't have conflicts between manager env and env of other components.

charts/fybrik/templates/manager-deployment.yaml
revit13: I fixed that. Thanks

manager/apis/app/v1alpha1/utils/configuration.go
revit13: I moved [manager/controllers/utils/configuration.go → pkg/environment/environment.go](https://github.com/fybrik/fybrik/pull/1512/files#diff-616422b9b983db329c853b3e73274dfdb95b49add7e27bcfd47257923eead1e8)

pkg/environment/environmentVariable.go
```diff
@@ -1,7 +1,7 @@
 // Copyright 2021 IBM Corp.
```
shlomitk1, line 1: Move the functions to environment.go - we don't need another file

pkg/environment/environment.go
```diff
@@ -1,7 +1,7 @@
 // Copyright 2020 IBM Corp.
 // SPDX-License-Identifier: Apache-2.0
 
-package utils
+package environment
```
shlomitk1, line 5: Move the functions that do not return env value (like SetIfNotSet, etc.) to utils.go

shlomitk1 (approved):

PR #1532, opened 2022-06-08
Use a fixed fybrik/optimizer image from ghcr.io repository

shlomitk1:
Fixes https://github.com/fybrik/fybrik/issues/1531

This PR uses ghcr.io/fybrik/optimizer:0.0.1 image. The version is specific to the solver version and is not related to Fybrik release.


charts/fybrik/values.yaml
```diff
@@ -215,8 +215,7 @@ manager:
   solver:
     # image of the container with solver binary and libs
     # when specified, the solver will be deployed in the manager pod
-    image: "optimizer"
-    imagePullPolicy: "Always"
+    image: "ghcr.io/fybrik/optimizer:master"
```
revit13, line 218: when creating a new release for Fybrik the optimizer tag should use the release tag?

shlomitk1: We are not creating a new image for optimizer for each release. But maybe it should represent the tool and the version underneath.

revit13: It might be better to use a fixed version for the optimizer rather than master for debugging purposes.

shlomitk1: Ok, using optimizer:0.1.

charts/fybrik/values.yaml
```diff
@@ -215,8 +215,7 @@ manager:
   solver:
     # image of the container with solver binary and libs
     # when specified, the solver will be deployed in the manager pod
-    image: "optimizer"
-    imagePullPolicy: "Always"
+    image: "ghcr.io/fybrik/optimizer:0.1"
```
revit13, line 218: i think its better to use a major.minor.patch format for the tag following [semantic versioning](https://semver.org/)

shlomitk1: Changed to 0.0.1

revit13 (approved):

PR #1540, opened 2022-06-11
Add backward compatability test

roytman:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
 include Makefile.env
 export DOCKER_TAGNAME ?= 0.0.0
 export KUBE_NAMESPACE ?= fybrik-system
-export OLD_CRD_VERSION ?= 0.7.0
+# the latest backward compatible CRD version
+export LATEST_BACKWARD_SUPPORTED_CRD_VERSION ?= 0.7.0
 export FYBRIK_CHARTS ?= https://fybrik.github.io/charts
 
 .PHONY: all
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -44,6 +46,18 @@ deploy: $(TOOLBIN)/kubectl $(TOOLBIN)/helm
...
 
+.PHONY: deploy_old_CRD_version
+deploy_old_CRD_version: export VALUES_FILE?=charts/fybrik/values.yaml
+deploy_old_CRD_version: $(TOOLBIN)/kubectl $(TOOLBIN)/helm
+	$(TOOLBIN)/kubectl create namespace $(KUBE_NAMESPACE) || true
+
+	$(TOOLBIN)/helm repo add fybrik-charts $(FYBRIK_CHARTS)
+	$(TOOLBIN)/helm repo update
+	$(TOOLBIN)/helm install fybrik-crd charts/fybrik-crd  \
+               --namespace $(KUBE_NAMESPACE) --version $(OLD_CRD_VERSION) --wait --timeout 120s
```
shlomitk1, line 57: I don't understand this command. If charts/fybrik-crd is used then we are installing the latest version, not the old one. For the old version the packaged charts should be used, shouldn't they?

roytman: Correct, the old release version should be used.

manager/testdata/notebook/read-flow/setup.sh
```diff
@@ -11,7 +12,12 @@ kubectl -n fybrik-notebook-sample apply -f s3credentials.yaml
...
 # Avoid using webhooks in tests
 kubectl delete validatingwebhookconfiguration fybrik-system-validating-webhook
-# Use master version of arrow-flight-module according to https://github.com/fybrik/arrow-flight-module#version-compatbility-matrix
-kubectl apply -f https://raw.githubusercontent.com/fybrik/arrow-flight-module/master/module.yaml -n fybrik-system
+
+if [[ -z "${OLD_AFM_VERSION}" ]]; then
+  # Use master version of arrow-flight-module according to https://github.com/fybrik/arrow-flight-module#version-compatbility-matrix
+  kubectl apply -f https://raw.githubusercontent.com/fybrik/arrow-flight-module/master/module.yaml -n fybrik-system
+else
+    kubectl apply -f https://github.com/fybrik/arrow-flight-module/releases/download/${OLD_AFM_VERSION}/module.yaml -n fybrik-system
```
revit13, line 20: indentation is slightly different than the upper line.

Makefile
```diff
@@ -1,6 +1,8 @@
 include Makefile.env
 export DOCKER_TAGNAME ?= 0.0.0
 export KUBE_NAMESPACE ?= fybrik-system
+export OLD_CRD_VERSION ?= 0.7.0
```
revit13, line 4: Consider using `LATEST_BACKWARD_SUPPORTED_VERSION` here as in the workflow above instead of `OLD_CRD_VERSION` to avoid future mismatch.

Makefile.env
```diff
@@ -9,6 +9,9 @@ export PATH:=$(ABSTOOLBIN):${PATH}
 export LOGGING_VERBOSITY = -1 
 export PRETTY_LOGGING = true
 
+# export DOCKER_HOSTNAME?=localhost:5000
+# export DOCKER_NAMESPACE?=fybrik-system
+
```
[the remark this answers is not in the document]
roytman: yes, there are a lot of repeptitions of these lines in our Makefiles, originally I replaced all of them by these 2 line, but after that decided to do it in a separate PR (different functionality) 
Thanks

manager/testdata/notebook/read-flow/setup.sh
roytman: fixed

Makefile
roytman: I'm not sure that I undustand your point, where do you see future mismatch. I do want to have an option to deploy different versions of CRDs, Fybrik and the test module. (we can update Fybrik without updating the module version)

roytman: P.S. there is difference in the Fybrik and AFM versions, the first is `0.7.0` and the second is `v0.7.0`

Makefile.env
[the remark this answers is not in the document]
revit13: I see the export is removed in several places for example in `run-integration-tests` - I assume you want to put them back.

revit13 (approved):

PR #1543, opened 2022-06-14
Update kubectl version in site/docs/get-started/quickstart.md

revit13:


simanadler:
@revit13 @roytman Is this something that should be flagged for backward compatibility?

roytman:
> @revit13 @roytman Is this something that should be flagged for backward compatibility?

No, we don't use any `kubectl` advanced features, and in general kubectl 1.18 works pretty good with Fybrik, here we just want to be inline with the latest K8s versions.

roytman (approved):

PR #1558, opened 2022-06-15
Bandwidth-like optimiztion goals

zivnevo:
This handles region->region optimization goals.

It also fixes an issue with restrictions that are assigned to `transform` capability.


pkg/optimizer/datapath_csp.go
```diff
@@ -170,7 +188,9 @@ func (dpc *DataPathCSP) BuildFzModel(pathLength int) (string, error) {
 // enforce restrictions from admin configuration decisions:
 // a. cluster satisfies restrictions for the selected capability
 // b. storage account satisfies restrictions for the selected capability
-// c. Ensure capabilities that must be deployed are indeed deployed
+// c. cluster satisfies "transform" restrictions if a governance action is selected
+// d. storage account satisfies "transform" restrictions if a governance action is selected
+// e. capabilities that must be deployed are indeed deployed
```
shlomitk1, line 193: Why is a module missing, i.e. module satisfies "transform" restrictions if a governance action is selected?

shlomitk1 (approved):

pkg/optimizer/datapath_csp.go
zivnevo: fixed

simanadler (approved):

PR #1561, opened 2022-06-18
Remove redundant code form helmer

roytman:
- Remove repetitive calls of `getConfig`
- Remove some redundant code
- Several addition helm related calls optimizations.  

### DIFF
--- a/manager/controllers/app/blueprint_controller.go
+++ b/manager/controllers/app/blueprint_controller.go
@@ -200,7 +200,7 @@ func (r *BlueprintReconciler) obtainSecrets(ctx context.Context, log *zerolog.Lo
 }
 
 func (r *BlueprintReconciler) applyChartResource(ctx context.Context, cfg *action.Configuration, chartSpec fapp.ChartSpec,
-	args map[string]interface{}, releaseNameSpace, releaseName string, log *zerolog.Logger) (ctrl.Result, error) {
+	args map[string]interface{}, releaseNamespace, releaseName string, log *zerolog.Logger) (*release.Release, error) {
 	log.Trace().Str(logging.ACTION, logging.CREATE).Msg("--- Chart Ref ---\n\n" + chartSpec.Name + "\n\n")
 
 	args = CopyMap(args)
@@ -214,12 +214,12 @@ func (r *BlueprintReconciler) applyChartResource(ctx context.Context, cfg *actio
 	var err error
 	registrySuccessfulLogin, err = r.obtainSecrets(ctx, log, chartSpec)
 	if err != nil {
-		return ctrl.Result{}, err
+		return nil, err
 	}
 
 	tmpDir, err := ioutil.TempDir("", "fybrik-helm-")
 	if err != nil {
-		return ctrl.Result{}, errors.WithMessage(err, chartSpec.Name+": failed to create temporary directory for chart pull")
+		return nil, errors.WithMessage(err, chartSpec.Name+": failed to create temporary directory for chart pull")
 	}
 	defer func(log *zerolog.Logger) {
 		if err = os.RemoveAll(tmpDir); err != nil {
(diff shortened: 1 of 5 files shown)


roytman:
@shlomitk1 @revit13 , I removed some redundant helm related code. 
Could you please take a look that I don't miss something.

pkg/helm/helm.go
```diff
@@ -154,22 +143,13 @@ type Impl struct {
...
-	}
+func (r *Impl) Uninstall(cfg *action.Configuration, releaseName string) (*release.UninstallReleaseResponse, error) {
 	uninstall := action.NewUninstall(cfg)
 	return uninstall.Run(releaseName)
 }
 
 // Load helm chart
 func (r *Impl) Load(ref, chartPath string) (*chart.Chart, error) {
-	// check for chart mounted in container
-	chrt, err := loader.Load(chartsMountPath + ref)
```
revit13, line 169: why this code is removed?

manager/controllers/app/blueprint_controller.go
```diff
@@ -193,10 +199,9 @@ func (r *BlueprintReconciler) obtainSecrets(ctx context.Context, log *zerolog.Lo
 	return registrySuccessfulLogin, nil
 }
 
-func (r *BlueprintReconciler) applyChartResource(ctx context.Context, log *zerolog.Logger, chartSpec fapp.ChartSpec,
-	args map[string]interface{}, blueprint *fapp.Blueprint, releaseName string) (ctrl.Result, error) {
+func (r *BlueprintReconciler) applyChartResource(ctx context.Context, cfg *action.Configuration, chartSpec fapp.ChartSpec,
+	args map[string]interface{}, releaseNameSpace, releaseName string, log *zerolog.Logger) (ctrl.Result, error) {
```
shlomitk1, line 203: releaseNameSpace -> releaseNamespace
Can't it be extracted from the config?

shlomitk1 (commented):
A couple of comments:
1. In the manager deployment we have an option to deploy the manager pod with a pre-configured chart. See https://github.com/fybrik/fybrik/blob/1c1529b4c4bfda73e90be6abd57b130e1a32ba57/charts/fybrik/templates/manager-deployment.yaml#L142-L145 and https://github.com/fybrik/fybrik/blob/1c1529b4c4bfda73e90be6abd57b130e1a32ba57/charts/fybrik/values.yaml#L222-L227
Either we disable this option, and the deployment yaml is updated accordingly, or we keep it. In the latter case, perhaps it will make more sense to define an env var that will prevent manager from trying to load a local chart if it is not there.

2. What is the reason for applying both Install and Upgrade instead of only Upgrade with Install option set to true ?

3. Keeping HELM_DEBUG env variable may be useful for debugging purposes.
4. Did WAIT option reduce the number of idle status checks? Please check that the timeout has been indeed propagated to the helmer inner functions (I had once problems with that, 0 was used instead of the passed timeout value, so wait basically did nothing...)

manager/controllers/app/blueprint_controller.go
roytman: - releaseNameSpace -> releaseNamespace 
OK
- Can't it be extracted from the config? 
yes and no. It is stored in one of implementations of `kube.Interface` when it is Kubernetes.Client, it's available there, but there are 2 other implementations...

roytman:
> A couple of comments:
> 
> 1. In the manager deployment we have an option to deploy the manager pod with a pre-configured chart. See https://github.com/fybrik/fybrik/blob/1c1529b4c4bfda73e90be6abd57b130e1a32ba57/charts/fybrik/templates/manager-deployment.yaml#L142-L145
>     and https://github.com/fybrik/fybrik/blob/1c1529b4c4bfda73e90be6abd57b130e1a32ba57/charts/fybrik/values.yaml#L222-L227
>    
>    Either we disable this option, and the deployment yaml is updated accordingly, or we keep it. In the latter case, perhaps it will make more sense to define an env var that will prevent manager from trying to load a local chart if it is not there.
> 2. Keeping HELM_DEBUG env variable may be useful for debugging purposes.
> 3. Did WAIT option reduce the number of idle status checks? Please check that the timeout has been indeed propagated to the helmer inner functions (I had once problems with that, 0 was used instead of the passed timeout value, so wait basically did nothing...)
> 4. Please set DisableOpenAPIValidation to the correct value.

1. I'll return this option.
2. HELM_DEBUG was used by the  helmer proprietary logger, now helmer uses the fybrik logger. So it prints log messages according to the fybrik log level.
3. If you specify WAIT and timeout 0, the install/upgrade operations fails with timeout error. You have to specify wait and timeout. And yes, it reduces number of status checks.
4. The correct value of DisableOpenAPIValidation is false and it is the default value. It was set to true as a workaround for the memory leak.

pkg/helm/helm.go
roytman: returned it back

shlomitk1 (approved):

PR #1562, opened 2022-06-19
⚠️ Add metrics to infrastructure.json

shlomitk1:
Fixes https://github.com/fybrik/fybrik/issues/1536
Fixes https://github.com/fybrik/fybrik/issues/1550


charts/fybrik/files/taxonomy/taxonomy.json
```diff
@@ -26,6 +26,42 @@
     "Attribute": {
```
simanadler, line 26: Please rename to InfraAttributeName.  However, do we really need this definition?  Can't there just be a string in InfraAttributes?

samples/adminconfig/infrastructure.json
```diff
@@ -1,45 +1,47 @@
 {
+    "definitions": [
+        {
+            "attribute": "storage-cost",
+            "type": "numeric",
+            "units": "US Dollar per TB per month",
+            "object": "fybrikstorageaccount"
```
simanadler, line 7: I thought the idea was to separate out the units and scales so they wouldn't have to be repeated, but the instance specific information would remain in the InfraAttribute.  Thus, wouldn't it make more sense to have the object in InfraAttribute?

samples/adminconfig/infrastructure.json
```diff
@@ -1,45 +1,47 @@
 {
+    "definitions": [
+        {
+            "attribute": "storage-cost",
```
simanadler, line 4: Do we need to map to the particular attribute type?  If for example, I have a price that is in us dollars with a scale of 0..100, couldn't I in theory use it with different attribute types?

samples/adminconfig/infrastructure.json
simanadler, line 7: Where is the scale?

charts/fybrik/files/taxonomy/taxonomy.json
```diff
@@ -26,6 +26,42 @@
...
+          "description": "A resource defined by the attribute (\"fybrikstorageaccount\",\"fybrikmodule\",\"cluster\")"
+        },
+        "objects": {
+          "description": "Argument types for cross-object attributes",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/InstanceType"
+          }
+        },
+        "scale": {
```
simanadler, line 51: Shouldn't scale, type, and unit be in a separate structure that can be reused by multiple InfraAttribute instances?

simanadler (changes requested):
See my comments.  If necessary please schedule a meeting to discuss.

charts/fybrik/files/adminconfig/infrastructure.json
```diff
@@ -1,3 +1,4 @@
 {
+    "definitions": [],
```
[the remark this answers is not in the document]
shlomitk1: It's not only units. I'll call it InfraAttributeMetrics.

samples/adminconfig/infrastructure.json
shlomitk1: The idea was to separate the instance-specific details from common attribute details. Let's take a storage cost as an example. It is a numeric value, measured in dollars per GB, varied from 0 to 100, defined in a storage account resource. 
Now, the specifics: storage cost associated with accountX costs Y. 
This way, it is easier to understand what is affected by an attribute - e.g. storage cost affects the choice of storage accounts. Otherwise, we would end up with a single attribute "cost" that is defined for storage accounts, clusters, networks, etc. For the current solver implementation, the assumption was that an attribute is defined in a unique way, and storage-cost and cluster-cost are different attributes even if they match units and scale.

shlomitk1: Added the scale.

charts/fybrik/files/taxonomy/taxonomy.json
shlomitk1: Created a separate InfrastructureMetrics structure.

samples/adminconfig/infrastructure.json
shlomitk1: Moved the attribute out. The connection to the metrics is done using metrics name.

samples/adminconfig/infrastructure.json
shlomitk1: The original idea was to have an attribute per object, i.e. storage-cost and cluster-cost are different entities. However, on the second thought, we might want to have the same "cost" attribute applied to different objects instead of separating the attributes. I've made the changes in the solver interfaces to support this. @zivnevo - please take a look (GetInstanceType has been replaced by GetInstanceTypes).

charts/fybrik/files/taxonomy/taxonomy.json
```diff
@@ -26,6 +26,42 @@
     "Attribute": {
       "type": "string"
     },
+    "AttributeDefinition": {
```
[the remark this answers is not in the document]
shlomitk1: Renamed to InfrastructureMetrics.

charts/fybrik/files/adminconfig/infrastructure.json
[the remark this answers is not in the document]
shlomitk1: Replaced "definitions" with "metrics".

samples/adminconfig/infrastructure.json
```diff
@@ -1,45 +1,65 @@
...
+                "min": 0,
+                "max": 20000
+            }
+        }
+    ],
     "infrastructure": [
         {
             "attribute": "storage-cost",
             "description": "neverland object store",
+            "metricsName": "cost",
```
zivnevo, line 28:
```suggestion
            "metricName": "cost",
```
Should be singular, because it's only one metric we are using (but verify with Sima)

samples/adminconfig/infrastructure.json
```diff
@@ -1,45 +1,65 @@
...
-            "attribute": "bandwidth",
-            "description": "bandwidth between neverland and theshire",
-            "value": "5",
-            "type": "numeric",
-            "units": "Mbps",
+            "attribute": "distance",
+            "description": "distance between neverland and theshire",
+            "metricsName": "distance",
+            "value": "5000",
+            "argObjects": ["fybrikstorageaccount","fybrikstorageaccount"],
```
zivnevo, line 46: This attribute refers to regions, so why do we need the argObjects and why are they set to `fybrikstorageaccount`?
Will we also have values for storage-account->cluster and cluster->cluster ?

pkg/optimizer/datapath_csp.go
```diff
@@ -539,48 +539,49 @@ func (dpc *DataPathCSP) addAnOptimizationGoal(goal adminconfig.AttributeOptimiza
...
 	resArray := []string{}
 	varName := ""
-	switch *instanceType {
-	case taxonomy.Cluster:
-		varName = clusterVarname
-		for _, cluster := range dpc.env.Clusters {
-			infraElement := dpc.env.AttributeManager.GetAttribute(attr, cluster.Name)
-			if infraElement == nil {
-				return "", "", fmt.Errorf("attribute %s is not defined for cluster %s", attr, cluster.Name)
+	for _, instanceType := range instanceTypes {
```
zivnevo, line 549: This is probably not how we want to handle bandwidth-like attributes. We have to check `len(instanceTypes)`. If it's 1, then it's a simple optimization goal and the current code is ok. If it's 2, then it's bandwidth-like optimization goal and we have to `return "", "", nil`

zivnevo (changes requested):

charts/fybrik/files/taxonomy/taxonomy.json
shlomitk1: Done

samples/adminconfig/infrastructure.json
shlomitk1: Changed to metricName

samples/adminconfig/infrastructure.json
shlomitk1: Right, reverting this change.

pkg/optimizer/datapath_csp.go
shlomitk1: This is not related to bandwidth-like attributes. The use-case is a single "cost" attribute defined for both clusters and storage-accounts. When this cost should be minimized it means minimizing everything on the data path that has "cost".
I saw that the code fills an array, so it should not be a problem, or should it?

samples/adminconfig/infrastructure.json
simanadler: Right.  I don't think we should have separate storage-cost and cluster-cost, just cost that can be used by multiple types of attributes.

samples/adminconfig/infrastructure.json
simanadler: Agree. singular

pkg/optimizer/datapath_csp.go
simanadler: Policies could be about storage cost, cluster cost, or all costs. Minimization should be done on the relevant costs indicated in the policy.  Right?

zivnevo (changes requested):
I'm afraid the suggested change to `setSimpleGoalVarArray()` is not going to work.
We really need a **different** paramArray for each instance type, because the i-th entry in the paramArray means the attribute-value for module/cluster/storage number i. So we cannot just put them all into one param array, bur rather need to call `getAttributeMapping()` for each `instanceType`.

In fact, we need to call `setSimpleGoalVarArray()` for each `instanceType`. Each such call will result in a variable-array holding the costs of the selected clusters/modules/storage-accounts along the data-path. We then need to sum each such array, and finally sum the sums :cry:.

I suggest to delay this "total cost" feature for a later PR. For now, forbid this feature and open a new issue.

zivnevo (approved):

shlomitk1:
@roytman This PR changes infrastructure.json format as well as the corresponding validation schema and taxonomy json. 
Need to be reflected in the release notes as changing user interfaces.

PR #1567, opened 2022-06-21
Define local charts mount path as env var.

### LINKED ISSUE #1563: Add tests to deploy pre loaded modules charts
Fybrik supports working with locally mounted charts, instead of a remote repository.
See https://github.com/fybrik/fybrik/blob/1c1529b4c4bfda73e90be6abd57b130e1a32ba57/charts/fybrik/templates/manager-deployment.yaml#L142,

We need to add tests for this feature.

revit13:
closes #1563

This PR changes local charts mount path from being defined as constant in the code to defined as env var.
It also contains a test for the case where the charts are located in a local directory and thus should not be retrieved from the registry.


pkg/helm/helm_test.go
```diff
@@ -60,7 +61,7 @@ var (
 	password      = os.Getenv("DOCKER_PASSWORD")
 	insecure, _   = strconv.ParseBool(os.Getenv("DOCKER_INSECURE"))
 	chartRef      = ChartRef(hostname, namespace, chartName, tagName)
-	impl          = NewHelmerImpl(true)
+	impl          = NewHelmerImpl("")
```
shlomitk1, line 64: What is the purpose of defining TestLocalChartsMount if the path is empty?

revit13: for this test the path is not empty: please see https://github.com/fybrik/fybrik/blob/995bd0d08d4f1490b6c96524fa1613f98ae66974/pkg/helm/helm_test.go#L142

shlomitk1: It would be cleaner to remove the global definition and use the right definition for each test.

revit13: Sure, will do that. thanks

pkg/helm/helm_test.go
```diff
@@ -93,6 +92,7 @@ func TestHelmRegistry(t *testing.T) {
 	if err != nil {
 		t.Errorf("Unable to setup test temp charts directory: %s", err)
 	}
+	var impl = NewHelmerImpl("")
```
shlomitk1, line 95: impl := NewHelmerImpl("")

pkg/helm/helm_test.go
```diff
@@ -139,7 +139,7 @@ func TestLocalChartsMount(t *testing.T) {
 
 	rootPath := filepath.Dir(tmpChart)
 	// remove the "charts" suffix from the file path
-	impl = NewHelmerImpl(filepath.Dir(rootPath))
+	var impl = NewHelmerImpl(filepath.Dir(rootPath))
```
shlomitk1, line 142: No need to add var, replace = with :=

pkg/helm/helm_test.go
```diff
@@ -181,7 +181,7 @@ func TestHelmRelease(t *testing.T) {
 	}
 	var err error
 	origChart := buildTestChart()
-
+	var impl = NewHelmerImpl("")
```
shlomitk1, line 184: same comment

pkg/helm/helm_test.go
revit13: Fixed Thanks

pkg/helm/helm_test.go
revit13: Fixed Thanks

pkg/helm/helm_test.go
revit13: Fixed. Thanks

shlomitk1 (approved):

roytman (approved):

PR #1576, opened 2022-06-23
issue #1574 - normalize attribute values to scale

zivnevo:

Making sure that values are always between 0 (scale.Min) and 100 (scale.Max).
That is `(value - scale.Min) * 100 / (scale.Max - scale.Min)`

### DIFF
--- a/pkg/infrastructure/attribute_manager.go
+++ b/pkg/infrastructure/attribute_manager.go
@@ -126,23 +126,37 @@ func validateStructure(bytes []byte) error {
 }
 
 // GetAttributeValue returns the value of an infrastructure attribute based on the attribute and instance names
-func (m *AttributeManager) GetAttributeValue(name, instance string) (string, bool) {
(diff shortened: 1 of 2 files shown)


pkg/infrastructure/attribute_manager.go
```diff
@@ -132,6 +136,19 @@ func (m *AttributeManager) GetAttributeValue(name, instance string) (string, boo
 	return "", false
 }
 
+// returns the normalized-to-scale value of an infrastructure attribute based on the attribute and instance names
+func (m *AttributeManager) GetNormalizedAttributeValue(name, instance string, scale *taxonomy.RangeType) (string, error) {
```
shlomitk1, line 140: I don't think scale should be passed to any of the methods in AttributeManager. Given name+instance, or any other key, scale can be calculated directly from the attribute structure.

zivnevo: Done

shlomitk1 (approved):

PR #1584, opened 2022-06-28
Bump github.com/distribution/distribution to fix security vulnerabilities

ityuhui:
Address GHSA-qq97-vm5h-rrhg

Signed-off-by: Hui Yu <ityuhui@gmail.com>


ityuhui:
@roytman 

I change the code as I find https://pkg.go.dev/github.com/distribution/distribution is usually not imported by other projects.

It seems developers prefer to import https://pkg.go.dev/github.com/docker/distribution

roytman (commented):
Hi @ityuhui , thanks for the PR.

Please return the import to `github.com/distribution/distribution/reference`. There is no repository with name `github.com/docker/distribution`. The requests to this URL are redirected to `github.com/distribution/distribution`.

Probably  the repository was moved to the new location,  and the references that you mentioned were not updated.

ityuhui:
Got it. Thanks for pointing it out. I have rolled back the change.

And I found the reason that why `distribution/distribution` imports `docker/distribution`:

https://github.com/distribution/distribution/blob/b5ca020cfbe998e5af3457fda087444cf5116496/reference/normalize.go#L8

PR #1590, opened 2022-07-07
Add a call to delete asset from the catalog in the manager code.

revit13:
This PR adds a call in the manager code to delete asset from the catalog.

TODO: add test to the [delete module](https://github.com/fybrik/delete-module) to check that the asset is deleted.

### DIFF
--- a/manager/controllers/app/data_catalog_interface.go
+++ b/manager/controllers/app/data_catalog_interface.go
@@ -7,10 +7,35 @@ import (
 	app "fybrik.io/fybrik/manager/apis/app/v1alpha1"
 	"fybrik.io/fybrik/manager/controllers/utils"
 	"fybrik.io/fybrik/pkg/model/datacatalog"
+	"fybrik.io/fybrik/pkg/model/taxonomy"
 	"fybrik.io/fybrik/pkg/vault"
 
 	"github.com/rs/zerolog/log"
 )
+
+// DeleteAsset deletes an asset from a catalog
+// Input arguments:
+// - assetID: DataSetID as it appears in fybrik-application
+// - input: fybrik application
+// Returns:
+// - an error if happened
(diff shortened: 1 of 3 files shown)


manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -202,6 +202,15 @@ func (r *FybrikApplicationReconciler) checkReadiness(applicationContext Applicat
 			continue
 		}
 
+		if !isAssetReady(applicationContext, assetID) && dataCtx.Flow == taxonomy.DeleteFlow {
```
shlomitk1, line 205: Why !isAssetReady?

revit13: To avoid re-deleting the asset.

manager/controllers/app/data_catalog_interface.go
```diff
@@ -7,11 +7,36 @@ import (
...
+	}
+	credentialPath := ""
+	if utils.IsVaultEnabled() {
+		credentialPath = utils.GetVaultAddress() + vault.PathForReadingKubeSecret(input.Namespace, input.Spec.SecretRef)
+	}
+	// ??? should response also be checked
+	if _, err := r.DataCatalog.DeleteAsset(&request, credentialPath); err != nil {
+		log.Error().Err(err).Msg("failed to receive the catalog connector response in DeleteAsset")
+		return err
+	}
```
simanadler, line 35: Should you perhaps delete the credentials as well?

shlomitk1: There may be multiple datasets sharing the same credentials. If so, deleting the credentials automatically would be a problem. Also, the data owner may want to reuse the credentials. 
If the datasets are created and deleted using Fybrik, the credentials are stored in a secret pointed to by a storage account resource. You'll have to delete both, and at some point there will be no storage accounts left.

PR #1597, opened 2022-07-10
Remove extra reconciles of FybrikApplication

shlomitk1:
Fixes https://github.com/fybrik/fybrik/issues/1565

This PR does not perform unnecessary reconciles resulting in redundant plotter generation.
- Finalizers are patched instead of updating an entire object, thus the generation version does not change.
- Events coming from plotter updates have a "plotter_" prefix prepended to the app name, thus we can know clearly that the reconcile is caused by the plotter update.
- Since the plotter is generated before updating the app status, we get a "wrong" object when trying to obtain the FybrikApplication from the server. This PR checks that the status is consistent, i.e. a plotter update is received for an application that specified the plotter generation in the status.

### DIFF
--- a/manager/controllers/app/blueprint_controller.go
+++ b/manager/controllers/app/blueprint_controller.go
@@ -116,7 +116,7 @@ func (r *BlueprintReconciler) reconcileFinalizers(ctx context.Context, cfg *acti
 			// remove the finalizer from the list and update it, because it needs to be deleted together with the object
 			ctrlutil.RemoveFinalizer(blueprint, BlueprintFinalizerName)
 
-			if err := utils.UpdateFinalizers(ctx, r.Client, blueprint); err != nil {
+			if err := r.Client.Update(ctx, blueprint); err != nil {
 				return ctrl.Result{}, err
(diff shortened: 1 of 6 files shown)


roytman:
Why cannot use K8s `metadata.ownerReferences` field at least for `blueprints` and `plotters` which are in the same namespace instead of the finalizers mechanism ? That will rid off setting and removing of Finalizers.

shlomitk1:
> Why cannot use K8s `metadata.ownerReferences` field at least for `blueprints` and `plotters` which are in the same namespace instead of the finalizers mechanism ? That will rid off setting and removing of Finalizers.

Blueprints and Plotters are in the same namespace but not (necessarily) in the same cluster. Besides, the cost of setting Finalizers is zero since it is done during the object creation, and removal is done when the object is deleted anyway.
What we **could** do to impact the performance is to avoid polling if there is a single blueprint created in the same cluster and rather watch for the update event in the same way we do with plotters. Not for this PR, though.

roytman (approved):

PR #1616, opened 2022-07-18
Optimizer documentation

shlomitk1:
Related to https://github.com/fybrik/fybrik/issues/1589

- add soft policies to Configuration Policies section
- add infrastructure attributes to tasks


simanadler (commented):
See comments

site/docs/concepts/config-policies.md
```diff
@@ -269,3 +278,44 @@ config[{"capability": "copy", "decision": decision}] {
 }
 ```
 
+## Optimization goals
+
+In a typical Fybrik deployment there may be several possibilities to create a data plane that satisfies the user requirements, governance and configuration policies. Based on the enterprise policy, an IT administrator may affect the choice of the data plane by defining a policy with optimization goals.
```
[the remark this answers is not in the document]
shlomitk1: Configuration policies are mentioned in the previous sentence. This one refers specifically to optimization goals.

site/docs/concepts/config-policies.md
```diff
@@ -221,18 +221,28 @@ When writing configuration policies, infrastructure metrics and costs may also b
 For example, selection of a storage account may be based on a storage cost, selection of a cluster may provide a restriction on cluster capacity, and so on. 
 Infrastructure attributes and metrics are stored in the `/tmp/adminconfig/infrastructure.json` directory of the manager pod. Collection of the metrics and their dynamic update is beyond the scope of Fybrik. One may develop or use 3rd party solutions for monitoring and updating these infrastructure metrics.
 
+### Infrastructure metrics
```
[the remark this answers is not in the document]
shlomitk1: Moved to tasks.

site/docs/concepts/config-policies.md
```diff
@@ -269,3 +278,44 @@ config[{"capability": "copy", "decision": decision}] {
...
+    input.request.usage == "copy"
+    policy := {"ID": "save-cost", "description":"Save storage costs", "version": "0.1"}
+    decision := {"policy": policy, "strategy": [{"attribute": "storage-cost", "directive": "min"}]}
+}
+```
+
+### Weights
+
+If more than one goal is provided, they can have a different weight. By default, all weights are equal to 1.  
+For example, the rule below defines two goals with weights in the 4:1 ratio.
```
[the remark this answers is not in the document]
shlomitk1: Added.

site/docs/concepts/config-policies.md
```diff
@@ -215,57 +215,45 @@ config[{"capability": "copy", "decision": decision}] {
 ```
 Note that an empty ConfigDecisions map will be returned if the expiration date is exceeded by the time when the policy is applied. 
 
-## Taking infrastructure metrics into consideration
 
-When writing configuration policies, infrastructure metrics and costs may also be taken into account in order to optimize the generated data plane. 
-For example, selection of a storage account may be based on a storage cost, selection of a cluster may provide a restriction on cluster capacity, and so on. 
-Infrastructure attributes and metrics are stored in the `/tmp/adminconfig/infrastructure.json` directory of the manager pod. Collection of the metrics and their dynamic update is beyond the scope of Fybrik. One may develop or use 3rd party solutions for monitoring and updating these infrastructure metrics.
+## Optimization goals
```
simanadler, line 219: Not related to the current change but ...
I will continue reviewing later, but a couple of general comments:
* Should the title of this page be IT config policies, rather than config policies?
* Regarding the policySetID, I wasn't aware that we had a way to link a policy set to a particular FybrikApplication.  Why?  When does this make sense?  
* There are instructions for providing custom policies when deploying fybrik, but no mention of how to update them after it's deployed

simanadler, line 219: We need to decide on terminology and be consistent throughout: are we calling these soft policies or optimization goals?  I'm fine with either as long as we're consistent in the code and in the documentation.

site/docs/concepts/config-policies.md
```diff
@@ -215,57 +215,45 @@ config[{"capability": "copy", "decision": decision}] {
...
-    policy := {"ID": "copy-restrict-storage", "description":"Use cheaper storage", "version": "0.1"}
-    decision := {"policy": policy, "restrictions": {"storageaccounts": [account_restrict]}}
+# minimize distance, minimize storage cost for read scenarios
+optimize[decision] {
+    input.request.usage == "read"
+    policy := {"ID": "general-strategy", "description":"focus on higher performance while saving storage costs", "version": "0.1"}
+    optimize_distance := {"attribute": "distance", "directive": "min", "weight": "0.8"}
+    optimize_storage := {"attribute": "storage-cost", "directive": "min", "weight": "0.2"}
+    decision := {"policy": policy, "strategy": [optimize_distance,optimize_storage]}
 }
```
simanadler, line 257: Should we mention that optimization is turned off by default and link to how to turn it on?

site/docs/concepts/config-policies.md
shlomitk1: I'll change to IT config policies and will add instructions how to update the policies.
Regarding policySetID, when we presented policies to get feedback, one of the feedbacks was that different organizations may have different policies while all the policies are kept in the same repo, and this is the way to differentiate. This approach is half baked: there is a possibility to write policies with different ids but we don't populate this information in FybrikApplication, so for now all policies are global for all fybrik applications.

shlomitk1: I vote for optimization goals. I think that soft policies can be misleading inferring that any type of policy can be set to either, while we always refer to `config` as hard and `optimize` as soft. You can't, for example, prefer deployment of read in the workload cluster, it is either restricted or not.

simanadler: OK.  I didn't recall that feedback.  In any case, I don't think it makes sense to expect the policySetID to be passed in the FybrikApplication.  I think this is a larger issue.  Perhaps just remove this section about policySetID and open a separate issue or discussion to decide how to address correctly?

simanadler: I'm fine with that as long as we're consistent.

shlomitk1: Done.

simanadler (dismissed):
Approved once link added to Ziv's page

site/docs/concepts/config-policies.md
shlomitk1: Added a link to the optimizer page and instructions how to enable the optimization.

simanadler (approved):

PR #1623, opened 2022-07-19
Change webhook cert dir.

revit13:
This PR changes the webhook cert dir to be `/data` instead of `/tmp` similar to where the other data resides.

Reported by @roytman

### DIFF
--- a/manager/main.go
+++ b/manager/main.go
@@ -42,6 +42,8 @@ import (
 	"fybrik.io/fybrik/pkg/storage"
 )
 
+const certSubDir = "/k8s-webhook-server"
+
 var (
 	gitCommit string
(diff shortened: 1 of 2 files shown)


manager/main.go
```diff
@@ -54,6 +54,8 @@ func init() {
 	_ = coordinationv1.AddToScheme(scheme)
 }
 
+var CertDir = environment.GetDataDir() + "/k8s-webhook-server"
```
roytman, line 57: do we need a separate variable CertDir ?
maybe to define a const `/k8s-webhook-server` and set the value in the code
```
CertDir:     environment.GetDataDir() + "/k8s-webhook-server"

roytman (approved):

PR #1639, opened 2022-07-24
fix print the ref of the last commit

### LINKED ISSUE #1638: Fix the print of the ltest commit reference in manager, add to conenctors.
When manager starts, it should print in the last git commit reference that was included in the current build.
Currently, it prints empty string.

After that, add the same mechanism to connectors.

roytman:
Fixes print the reference of the last commit in manager, add it to OPA and Katalog connectors. 

/closes #1638 

### DIFF
--- a/connectors/katalog/Makefile
+++ b/connectors/katalog/Makefile
@@ -4,8 +4,7 @@ DOCKER_NAME = katalog-connector
 include $(ROOT_DIR)/Makefile.env
 include $(ROOT_DIR)/hack/make-rules/docker.mk
 include $(ROOT_DIR)/hack/make-rules/tools.mk
-
-GIT_COMMIT := $(shell git rev-list -1 HEAD)
+include $(ROOT_DIR)/hack/make-rules/version.mk
 
 .PHONY: all
 all: docker-build docker-push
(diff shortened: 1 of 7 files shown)


revit13 (dismissed):

revit13:
btw there seems to be an alternative approach: https://levelup.gitconnected.com/a-better-way-than-ldflags-to-add-a-build-version-to-your-go-binaries-2258ce419d2d

roytman:
> btw there seems to be an alternative approach to avoid forgetting adding the flag: https://levelup.gitconnected.com/a-better-way-than-ldflags-to-add-a-build-version-to-your-go-binaries-2258ce419d2d

Cool, that was happened with the print in manager. Will update accordingly, thanks.

roytman:
After checking the suggested option, I don't think that it is better. 

1. We have to create a bash script files to generate the git info
2. The solution generates version.txt (or some other predefined name files), there is an option that an old file will be used, at least in a local run. 
3. anyway, we have to update the build commands to include `go generate` before. If we don't run `go generate` and the version.txt doesn't exist, the build will fail, but if the file does exist, the build will be OK with an old version.
4. if you work with IDE (goLand or VC) you have to run `go generate` each time before using the project. (I assume that version.txt, doesn't exist).

@revit13 what do you think ?

revit13:
The fact that the old version.txt can be used seems dangerous (although if the code is always built from the Makefile the chances for that are low right?). it might be safer to have a scenario where the version is empty than incorrect.

revit13 (approved):

PR #1641, opened 2022-07-25
Specify emptyDirs size limits

### LINKED ISSUE #1615: Specify emptyDirs size limits
Kubernetes best practices suggest defining size limits for volumes with type emptyDir

Please  specify it for `solver-volume` and `data`

revit13:
Closes #1615


roytman (commented):
I prefer to specify explicit size for each emptyDir, which will allow user 1. to overwrite the value and 2. to specify different values. 
We have seen that the solver is much bigger than `/data`

roytman (approved):

PR #1644, opened 2022-07-26
Update manager/testdata/blueprint-read.yaml

revit13:


shlomitk1:
There is also blueprint.yaml in testdata/ which should be either updated or removed if not used.
The uuid field in the arguments is missing.

shlomitk1 (approved):

PR #1647, opened 2022-07-28
Remove the tag of image and use the latest tag by default

ityuhui:
Without the tag, docker will pull the latest image of `ubi8/ubi-minimal`

We don't need manully bump the version later.

Signed-off-by: Hui Yu <ityuhui@gmail.com>

### DIFF
--- a/connectors/katalog/Dockerfile
+++ b/connectors/katalog/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/ubi-minimal:8.6
+FROM registry.access.redhat.com/ubi8/ubi-minimal
 
 ENV HOME=/tmp
 WORKDIR /tmp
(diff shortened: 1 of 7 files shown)


roytman:
I;m not sure that it is a good idea to pull the latest image. As one of the best practices, goLang recommendation is to specify the exact version of a package that you program works with. I think it is better to have a specific image version, so we can reproduce the running env, if we need it.

ityuhui:
> I;m not sure that it is a good idea to pull the latest image. As one of the best practices, goLang recommendation is to specify the exact version of a package that you program works with. I think it is better to have a specific image version, so we can reproduce the running env, if we need it.

OK. We can bump the tag manally later.

Maybe the base image for optimizer needs to bump to `8.6` to keep consistency with other images ?
https://github.com/fybrik/fybrik/blob/cb69f4dee2463f9cf9614791d13e3eb9bf61318c/pkg/optimizer/Dockerfile#L4

PR #1653, opened 2022-07-31
Code refactoring: apis cleanup

shlomitk1:
This PR replaces definitions used in manager/apis that are not a part of CRDs. Examples:
- Error messages
- Connection types/formats used in tests/mockups
- Names of labels


roytman (approved):

simanadler:
I'm struggling a bit to understand (1) what problem this is solving, (2) why put all the labels in utils, rather than with their related CRDs?   Utils were meant for things used by multiple components.

roytman:
> I'm struggling a bit to understand (1) what problem this is solving, (2) why put all the labels in utils, rather than with their related CRDs? Utils were meant for things used by multiple components.

When we create a new CRD version, we copy the old one and add required changes. When all constants and helmValues are there, we will need or to remove them from old version or have double values.

shlomitk1:
FybrikAppUUID was defined in utils, other labels in manager/apis - not very consistent.

PR #1666, opened 2022-08-04
Replace the wait flag in the helm operations by a configurable polling interval

### LINKED ISSUE #1661: Replace the wait flag in the helm operations by a configurable polling interval
We moved from poling of resources in Fybrik to `helm --wait`, but actually it uses 2 secs polling interval as we used in Fybrik. 
We have to provide an option to use a configurable polling interval, either by fybrik or by helm. 

CC @shlomitk1

revit13:
Closes #1661 
Closes https://github.com/fybrik/fybrik/issues/1622

This PR adds a configuration flag for the time interval to check the status of the resources deployed by the manager.
It is used by the blueprint-controller and in helm uninstall operation.


charts/fybrik/templates/manager-deployment.yaml
```diff
@@ -106,6 +106,8 @@ spec:
               value: {{ .Values.manager.tls.certs.certSecretName }}
             - name: CERT_SECRET_NAMESPACE
               value: {{ .Values.manager.tls.certs.certSecretNamespace }}
+            - name: RESOURCE_POLLING_INTERVAL
```
shlomitk1, line 109: Why here and not in the config map used by manager?

revit13: OK I can change that. Thanks

manager/controllers/app/blueprint_controller.go
```diff
@@ -376,7 +375,11 @@ func (r *BlueprintReconciler) reconcile(ctx context.Context, cfg *action.Configu
 	// the status is unknown yet - continue polling
 	if blueprint.Status.ObservedState.Error == "" {
 		log.Trace().Msg("blueprint.Status.ObservedState is not ready, will try again")
-		return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
+		interval, err := environment.GetResourcesPollingInterval()
+		if err != nil {
+			log.Error().Err(err).Msg("error getting resource polling interval :" + err.Error())
```
shlomitk1, line 380: The error here is meaningless since the operation will go through. It is better to print the error when there is a printout of all env vars used by manager. Something in light of "error getting the value. Setting the default to ..."

shlomitk1 (commented):
As a general comment, I don't think that the same number can be used for both. Checking every two seconds for a resource status makes sense, whereas install timeout is much longer. Not sure what happens if the install timeout is not sufficient - will the operation fail?

revit13:
> As a general comment, I don't think that the same number can be used for both. Checking every two seconds for a resource status makes sense, whereas install timeout is much longer. Not sure what happens if the install timeout is not sufficient - will the operation fail?

I think it will fail but I'll check. I will add an additional env var for operation timeout.

shlomitk1:
Also, please look at https://github.com/fybrik/fybrik/issues/1622 - another polling interval, the same configuration can be used for  polling blueprints and blueprint resources. Helm timeouts are different, I would not mix them all.

pkg/environment/environment.go
```diff
@@ -182,10 +225,20 @@ func LogEnvVariables(log *zerolog.Logger) {
...
 	}
+	interval, err := GetResourcesPollingInterval()
+	if err != nil {
+		log.Error().Msg("error getting " + ResourcesPollingInterval + ". Setting the default to " +
+			interval.String())
+	}
+	timeout, err := GetHelmWaitTimeout()
+	if err != nil {
+		log.Error().Msg("error getting" + HelmWaitTimeout + " Setting the default to " + timeout.String())
+	}
```
shlomitk1, line 243: get the new variables out of the loop and handle them separately. Add "else" to if err != nil to print the values.

revit13: @shlomitk1 I find that this can be a little confusing - if the env value is empty then we will print the default value but there is no way to tell that after we end the call to `GetResourcesPollingInterval` or `GetHelmWaitTimeout`. also, the printed value is not the value of the env var but rather of the returned value of  `GetResourcesPollingInterval` or `GetHelmWaitTimeout`.
another option is to initialize a new var based on the env var , log it then, and return that value afterward.  what do you think?

shlomitk1: Don't use these variables in the logEnvVar loop. Add else statement to print the value, like:
        timeout, err := GetHelmWaitTimeout()
	if err != nil {
		log.Warning().Msg("error getting" + HelmWaitTimeout + " Setting the default to " + timeout.String())
	} else {
	        log.Info().Msgf("%s set to \"%s\"", HelmWaitTimeout, timeout.String())
	}

revit13: OK, I change the code. Thanks

pkg/helm/helm.go
```diff
@@ -156,7 +154,11 @@ func NewHelmerImpl(chartsPath string) *Impl {
 func (r *Impl) Uninstall(cfg *action.Configuration, releaseName string) (*release.UninstallReleaseResponse, error) {
 	uninstall := action.NewUninstall(cfg)
 	uninstall.Wait = true
-	uninstall.Timeout = TIMEOUT
+	interval, err := environment.GetHelmWaitTimeout()
+	if err != nil {
```
shlomitk1, line 158: The idea is to use the default, not return the error.

revit13: Fixed. Thanks

shlomitk1 (dismissed):

pkg/environment/environment.go
```diff
@@ -49,6 +52,15 @@ const DefaultModulesNamespace = "fybrik-blueprints"
...
 const DefaultControllerNamespace = "fybrik-system"
 
+// defaultPollingInterval defines the default time interval to check the status of the resources
+// deployed by the manager. The interval is specified in milliseconds.
+const defaultPollingInterval = 2000 * time.Millisecond
+
+// defaultHelmWaitTimeout defines the default time to wait for helm operation
+// on the resources deployed by the manager to complete.
+// The timeout is specified in seconds.
+const defaultHelmWaitTimeout = 200 * time.Second
```
roytman, line 62: I'd set here 300 (5 mins), but it is not crucial

pkg/environment/environment.go
```diff
@@ -134,6 +146,37 @@ func GetModulesRole() string {
...
+// status of the resources deployed by the manager. The interval is specified
+// in milliseconds.
+func GetResourcesPollingInterval() (time.Duration, error) {
+	intervalStr := os.Getenv(ResourcesPollingInterval)
+	if intervalStr == "" {
+		return defaultPollingInterval, nil
+	}
+	interval, err := strconv.Atoi(intervalStr)
+	if err != nil {
+		return defaultPollingInterval, err
```
roytman, line 159: very strange, you return default value and error, it should be one of them

pkg/environment/environment.go
```diff
@@ -134,6 +146,37 @@ func GetModulesRole() string {
...
+// Currently used only in uninstall operation.
+// The timeout is specified in seconds.
+func GetHelmWaitTimeout() (time.Duration, error) {
+	timeoutStr := os.Getenv(ResourcesPollingInterval)
+	if timeoutStr == "" {
+		return defaultHelmWaitTimeout, nil
+	}
+	timeout, err := strconv.Atoi(timeoutStr)
+	if err != nil {
+		return defaultHelmWaitTimeout, err
```
roytman, line 175: like above

pkg/environment/environment.go
```diff
@@ -169,15 +212,28 @@ func GetDataCatalogServiceAddress() string {
 	return os.Getenv(CatalogConnectorServiceAddressKey)
 }
 
+const printValueStr = "%s set to \"%s\""
```
roytman, line 215: pls put constant at the start of the file

pkg/environment/environment.go
```diff
@@ -188,4 +244,9 @@ func LogEnvVariables(log *zerolog.Logger) {
 	for _, envVar := range envVarArray {
 		logEnvVariable(log, envVar)
 	}
+
```
roytman, line 247: GetDataPathMaxSize ?

revit13: the error of GetDataPathMaxSize if such exists is printed in FindPaths().

pkg/environment/environment.go
revit13: Fixed. thanks

pkg/environment/environment.go
revit13: done. also for  `GetDataPathMaxSize` function which uses the same pattern.

pkg/environment/environment.go
shlomitk1: Yes.

shlomitk1 (approved):

roytman (approved):

PR #1676, opened 2022-08-08
Fix a misleading statement in the manager-deployment.yaml` file

### LINKED ISSUE #1665: Fix a misleading statement in the manager-deployment.yaml` file
The `manager-deployment.yaml` file contains the following statement 
```yaml
metadata:
      {{- with .Values.manager.podAnnotations }}
      annotations:
        sidecar.istio.io/inject: "true"
        {{- toYaml . | nindent 8 }}
      {{- end }}
``` 
It's not clear what is a relationship between  `podAnnotations` and istio sidecar injection.  Why if we want to use any pod annotations, the istio sidecar injection will be set to `true`

In addition, we have to review istio usage, but it is another task.

revit13:
closes #1665 

This PR adds  annotations if such exists in `.Values.manager.podAnnotations`. 

The  `sidecar.istio.io/inject: "true"` exists as default annotation. it is need for sidecar injection for the manager if istio is needed.

### DIFF
--- a/charts/fybrik/values.yaml
+++ b/charts/fybrik/values.yaml
@@ -204,7 +204,8 @@ manager:
     # The name of the service account to use
     name: manager
 
-  podAnnotations: {}
+  podAnnotations:
+    sidecar.istio.io/inject: "true"
 
   # Pod Security Context. If set, the fields of podSecurityContext override
   # the equivalent fields of .Values.global.podSecurityContext.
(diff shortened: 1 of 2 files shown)


charts/fybrik/values.yaml
```diff
@@ -204,7 +204,8 @@ manager:
     # The name of the service account to use
     name: manager
 
-  podAnnotations: {}
+  podAnnotations:
+    sidecar.istio.io/inject: "true"
```
roytman, line 208: now by default you always add this annotation, I'm not sure that it is right

revit13: I put the label in a comment.

roytman (approved):

PR #1720, opened 2022-09-11
Add "Input to policies" section to using-opa.md 

revit13:


site/docs/tasks/using-opa.md
```diff
@@ -25,6 +26,14 @@ You can also add conditions like
 
 The verdict `allow` will be reached only if the conditions hold, and no other rule has been triggered, e.g. a rule requiring column redaction.
 
+## Input to policies
+
+The input object includes the application properties and the requested action as well as dataset details (id, metadata).
+
+- `properties`: application/workload properties defined in FybrikApplication, e.g. `properties.intent`
```
shlomitk1, line 33: I think it was `context` instead of `properties`

site/docs/tasks/using-opa.md
```diff
@@ -15,7 +15,8 @@ package dataapi.authz
 rule [{}] { true }
 ```
 
-You can also add conditions like
+To add a condition a new rule named `rule` should be added as shown below.
```
shlomitk1, line 18: This statement is misleading. It appears in the Default policies section. Should be in some other "general' section - an explanation about syntax - only `rule` is allowed, the output is either an empty object (representing allow action) or an object including action structure defined in the taxonomy as well as the policy description.

site/docs/tasks/using-opa.md
revit13: Fixed. Thanks

site/docs/tasks/using-opa.md
revit13: OK thanks will add a new section.

site/docs/tasks/using-opa.md
```diff
@@ -5,6 +5,44 @@ When OPA is used for data governance, it is deployed as a stand-alone service.
 
 For more details on OPA policies please refer to [OPA documentation](https://www.openpolicyagent.org/docs/latest/policy-language/) in particulate to [the basics](https://www.openpolicyagent.org/docs/latest/policy-language/#the-basics) section which explains how a policy is evaluated.
 
+## Fybrik Policies Syntax
+
+Policies are written in rego files.
+
+Fybrik policies are written in the following syntax: `rule[{"action": <action>, "policy": <policy>}]` where `policy` is a string describing the action and `action` is a rego object with the follwing form:
```
shlomitk1, line 12: Not necessarily in rego files, only for OPA. 
Maybe something like "OPA policies for Fybrik are written in rego files and have the following syntax..."

revit13: Fixed. thanks

site/docs/tasks/using-opa.md
```diff
@@ -5,6 +5,42 @@ When OPA is used for data governance, it is deployed as a stand-alone service.
 
 For more details on OPA policies please refer to [OPA documentation](https://www.openpolicyagent.org/docs/latest/policy-language/) in particulate to [the basics](https://www.openpolicyagent.org/docs/latest/policy-language/#the-basics) section which explains how a policy is evaluated.
 
+## OPA Policies Syntax
+
+OPA policies for Fybrik are written in rego files and have the following syntax: `rule[{"action": <action>, "policy": <policy>}]` where `policy` is a string describing the action and `action` is a rego object with the follwing form:
```
shlomitk1, line 10: JSON object
*following

site/docs/tasks/using-opa.md
```diff
@@ -5,6 +5,42 @@ When OPA is used for data governance, it is deployed as a stand-alone service.
...
+  rule[{"action": {"name":"RedactAction","columns": column_names}, "policy": description}] {
+        description := "Redact written columns tagged as sensitive in datasets tagged with finance = true. The data should not be stored in `neverland` storage account"
+        input.action.actionType == "write"
+        input.resource.metadata.tags.finance
+        input.action.destination != "neverland"
+        column_names := [input.resource.metadata.columns[i].name | input.resource.metadata.columns[i].tags.sensitive]
+    }
+```
+
+The following is an example of a new value that the rule above can have if it is evaluated to true:
```
shlomitk1, line 35: An example of an object that may be returned by the rule above:

site/docs/tasks/using-opa.md
revit13: Fixed. Thanks

site/docs/tasks/using-opa.md
revit13: Fixed. Thanks

shlomitk1 (approved):

PR #1731, opened 2022-09-19
OpenMetadata install script

cdoron:


third_party/openmetadata/prepare_OM_for_fybrik.go
```diff
@@ -0,0 +1,17 @@
+// Copyright 2020 IBM Corp.
```
roytman, line 1: why do you need this start connector here ?

cdoron: The NewOpenMetadataAPIService() method runs a prepareOpenMetadataForFybrik() method which creates the custom properties that we need.

third_party/openmetadata/prepare_OM_for_fybrik.go
```diff
@@ -0,0 +1,17 @@
...
+package main
+
+import (
+	logging "fybrik.io/fybrik/pkg/logging"
+	openapi_connector_core "fybrik.io/openmetadata-connector/pkg/openmetadata-connector-core"
+)
+
+func main() {
+	conf := map[interface{}]interface{}{
+		"openmetadata_endpoint": "http://localhost:8585/api",
```
roytman, line 13: can it be an argument or env parameter

third_party/openmetadata/prepare_OM_for_fybrik.go
```diff
@@ -0,0 +1,17 @@
...
+	}
+	logger := logging.LogInit(logging.CONNECTOR, "OpenMetadata Prepare for Fybrik")
+	openapi_connector_core.NewOpenMetadataAPIService(conf, &logger)
```
roytman, line 16: can we separate the connection and tag creation from NewOpenMetadataAPI Service creation, and get the custom properties and tags from external files

third_party/openmetadata/prepare_OM_for_fybrik.go
cdoron: I removed the go code.
The preparation is now done using the same executable as that of the openmetadata-connector (with different parameters).

third_party/openmetadata/prepare_OM_for_fybrik.go
cdoron: The preparation code gets the endpoint, username, and password from an environment variable.

third_party/openmetadata/prepare_OM_for_fybrik.go
cdoron: done.

third_party/openmetadata/install_OM.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+OM_VERSION=0.12.1
+export OPENMETADATA_AIRFLOW_VERSION=0.12.1
+export OPENMETADATA_HELM_CHART_VERSION=0.0.39
+
+export export FYBRIK_BRANCH="${FYBRIK_BRANCH:-master}"
+export export FYBRIK_NAMESPACE="${FYBRIK_NAMESPACE:-fybrik}"
```
roytman, line 8: I think its name is organization (guthub organization) and not a namespace

third_party/openmetadata/install_OM.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+OM_VERSION=0.12.1
+export OPENMETADATA_AIRFLOW_VERSION=0.12.1
+export OPENMETADATA_HELM_CHART_VERSION=0.0.39
+
+export FYBRIK_BRANCH="${FYBRIK_BRANCH:-master}"
+export FYBRIK_GITHUB_ORGANIZATION="${FYBRIK_NAMESPACE:-fybrik}"
```
roytman, line 8: ${FYBRIK_NAMESPACE:-fybrik} -> ${FYBRIK_GITHUB_ORGANIZATION:-fybrik}

roytman (approved):

PR #1732, opened 2022-09-20
Add readiness and liveness probs to other pods

### LINKED ISSUE #1723: Add readiness and liveness probs to other prods
In https://github.com/fybrik/fybrik/pull/1703 we added readiness and liveness probs to the manager pod. We need to add similar probes to other pods.

revit13:
/closes #1723


charts/fybrik/values.yaml
```diff
@@ -341,6 +364,11 @@ opaServer:
   # ref: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#security-context-1
   containerSecurityContext: {}
 
+  enableReadinessProbe: true
```
roytman, line 367: why did you add `enableReadinessProbe` and `enableLivenessProbe` to OPA server only ?

roytman (commented):
Great work, thanks. 
Do we have any test cases to check this PR ?

charts/fybrik/values.yaml
revit13: OPA server implements the proves with http requests which require certificate verification when it uses tls as is used in notebook-test-readflow.tls. to avoid configuring kublet with the CA certificates this flag was added.
The rest of the deployments are not using probes with http requests so they can be executed anyway.

revit13:
@roytman I did not include tests. I can try to add them.

charts/fybrik/values.yaml
roytman: Can you add some explanation in the file ?

revit13: Sure, I mistakenly put it only in charts/fybrik/notebook-test-readflow.tls.values.yaml

roytman (approved):

PR #1738, opened 2022-09-23
Configure Vault with TLS

### LINKED ISSUE #1736: Configure Vault with TLS
This issue includes:

- Adding TLS support to arrow-flight-module which communicates which Vault
- Adding a test

revit13:
/closes #1736

This PR configures vault to enable tls.
As vault `dev` mode does [not support tls](https://discuss.hashicorp.com/t/tls-in-dev-mode/44565/2), vault `HA` mode is used.
`run-notebook-readflow-tls-tests` is changed to use vault with tls


cdoron (approved):
Looks good to me. **Should also be reviewed by someone with better understanding of TLS**

charts/fybrik/notebook-test-readflow.tls.values.yaml
```diff
@@ -3,7 +3,7 @@
 # It overrides values that are different from the default values.yaml
 # The values used for testing read flow where the manager and connectors
 # uses mutual tls connection. In addition, OPA server is configured to use
-# mutual tls.
+# mutual tls and Vault server is configured to use tls.
```
roytman, line 6: mutual TLS is the server setting, how and where do you configure Vault to require mTLS ?

charts/vault/env/standalone/vault-single-cluster-values-tls.yaml
```diff
@@ -0,0 +1,102 @@
...
+      config: |
+        plugin_directory = "/usr/local/libexec/vault"
+        ui = true
+        listener "tcp" {
+          address = "[::]:8200"
+          cluster_address = "[::]:8201"
+          tls_disable = false
+          tls_cert_file = "/vault/userconfig/vault-server-tls/tls.crt"
+          tls_key_file  = "/vault/userconfig/vault-server-tls/tls.key"
+          # TLS client verification is optional with default settings and is not enforced.
```
roytman, line 20: TLS client verification is the mutual TLS

manager/testdata/notebook/read-flow-tls/setup-certs.sh
```diff
@@ -2,12 +2,13 @@
 # Copyright 2021 IBM Corp.
 # SPDX-License-Identifier: Apache-2.0
 
+kubectl create ns fybrik-system
```
roytman, line 5: should we define `fybrik-system` as a variable ? it is returned to itself a lot of times

python/tls/README.md
```diff
@@ -0,0 +1,3 @@
+## Python TLS Package for Fybrik Components
+The versions of this library do not correspond to the
+versions of the Fybrik repository.
```
roytman, line 3: why do we need a separate version here ? 
If we use a separate versioning here, please add it into the backward compatibility doc.

python/tls/setup.py
```diff
@@ -0,0 +1,18 @@
+from setuptools import setup
+import os
+
+setup(name='fybrik_python_tls',
+      version=os.environ.get('VERSION', '0.1.0'),
```
roytman, line 5: if you need this env variable,  so maybe "Python-TLS-Version" or something similar ? We can face with a conflict.

charts/fybrik/notebook-test-readflow.tls.values.yaml
revit13: @roytman I changed the test to use Vault with mutual TLS. please also see manager/testdata/notebook/README.txt for test description. Thanks

manager/testdata/notebook/read-flow-tls/setup-certs.sh
revit13: I fixed that Thanks

python/tls/README.md
revit13: I think we do. Fybrik has its own considerations for bumping versions that are unrelated to the python library (it relates to the module code if any). On the other hand, the go library  depends on fybrik. The logging mentioned in https://github.com/fybrik/fybrik/blob/master/docs/BackwardCompatibility.md#format-of-fybrik-log-files seems to refer to the go logging library as for python we have this [statement](https://github.com/fybrik/fybrik/tree/master/python/logging) in the README.d file. Let talk to clarify this.

python/tls/setup.py
revit13: Changed Thanks

roytman (approved):

PR #1741, opened 2022-09-28
add configurable discoveryBurst and discoveryQPS

roytman:
Allows to configure discovery K8s client parameters for helm operations.
/closes 1740

### DIFF
--- a/pkg/environment/environment.go
+++ b/pkg/environment/environment.go
@@ -188,7 +188,14 @@ func GetDiscoveryBurst() (int, error) {
 	if burstStr == "" {
 		return -1, nil
 	}
-	return strconv.Atoi(burstStr)
+	burst, err := strconv.Atoi(burstStr)
+	if err != nil {
+		return -1, err
+	}
+	if burst <= 0 {
+		return -1, fmt.Errorf("discovery burst should be positive, got %d", burst)
+	}
+	return burst, err
 }
 
 // GetDiscoveryQPS returns the K8s discovery QPS value if it is set, otherwise it returns -1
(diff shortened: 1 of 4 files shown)


pkg/environment/environment.go
```diff
@@ -179,6 +182,29 @@ func GetHelmWaitTimeout() (time.Duration, error) {
 	return time.Duration(timeout) * time.Second, nil
 }
 
+// GetDiscoveryBurst returns the K8s discovery burst value if it is set, otherwise it returns -1
+func GetDiscoveryBurst() (int, error) {
+	burstStr := os.Getenv(DiscoveryBurst)
+	if burstStr == "" {
+		return -1, nil
+	}
+	return strconv.Atoi(burstStr)
```
shlomitk1, line 191: In case of an error, 0 is returned as a value.

revit13 (approved):

pkg/helm/helm.go
```diff
@@ -299,6 +308,13 @@ func (r *Impl) GetConfig(kubeNamespace string, log action.DebugLog) (*action.Con
 	config := &genericclioptions.ConfigFlags{
 		Namespace: &kubeNamespace,
 	}
+	if r.discoveryBurst != -1 {
```
shlomitk1, line 311: I think it's better to refer to positive values only: **> 0**

pkg/helm/helm.go
```diff
@@ -299,6 +308,13 @@ func (r *Impl) GetConfig(kubeNamespace string, log action.DebugLog) (*action.Con
...
+		config.WithDiscoveryBurst(r.discoveryBurst)
+	}
+	if r.discoveryQPS != -1 {
```
shlomitk1, line 314: I think it's better to refer to positive values only: **> 0**

pkg/environment/environment.go
roytman: thanks, fixed

pkg/helm/helm.go
roytman: the `GetDiscoveryBurst` and `GetDiscoveryQPS` are defined that they return `-1` if something wrong, so I think checking `-1` is appropriate here.

pkg/helm/helm.go
roytman: as above

shlomitk1: Still, it is a floating point variable here, using ">" is safer.

roytman: I don't agree with you, there is the positivity check in the function.

shlomitk1: Never mind, you are not doing any calculations to be worried about precision.

shlomitk1 (approved):

PR #1757, opened 2022-10-03
add support for deploying vault with the openmetadata plugin

cdoron:
add support for deploying vault with the openmetadata plugin


hack/tools/configure_vault_kind_multi.sh
```diff
@@ -26,6 +26,7 @@ source ./vault_utils.sh
 # use the vault-plugin-secrets-kubernetes-reader plugin enabled in Vault
 # path kubernetes-secrets.
 : ${PLUGIN_PATH:=kubernetes-secrets}
+: ${OMD_PLUGIN_PATH:=kubernetes-secrets}
```
revit13, line 29: kubernetes-secrets -> omd-secrets

charts/vault/env/dev/plugin-secrets-omd-reader-values.yaml
```diff
@@ -0,0 +1,36 @@
+vault:
```
revit13, line 1: it might be better to merge plugin-secrets-omd-reader-values.yaml and plugin-secrets-kubernetes-reader-values.yaml to a single file that will have both of the plugins. similar to how vault-single-cluster-values.yaml looks like.

revit13 (approved):

charts/vault/templates/post-pod-creation-cm.yaml
```diff
@@ -16,13 +16,19 @@ data:
...
+      vault secrets enable -path=omd-secrets vault-plugin-secrets-omd-reader
       # create policy to access secrets
       # NOTE: fybrik/dataset-creds/ secret engine is enabled by the manager and is used temporarily
       # to store dataset credentials.
       vault policy write "allow-all-dataset-creds" - <<EOF
       path "kubernetes-secrets/*" {
       capabilities = ["create", "read", "update", "delete", "list"]
       }
+      path "omd-secrets/*" {
+      capabilities = ["create", "read", "update", "delete", "list"]
```
roytman, line 30: do we need these rules? Especially "create", "delete" and "update"? We just read from OMD, don't we ?

hack/tools/vault_utils.sh
```diff
@@ -81,6 +81,9 @@ create_policy_with_plugin_path() {
         path "$2" {
         capabilities = ["create", "read", "update", "delete", "list"]
         }
+        path "$3" {
```
roytman, line 84: the function comments are not updated:

>  Create a policy to allow access to Vault plugin path as well as
# for the path where dataset credentials resides. This is temporary
# until dataset credentials path become obselete.
# $1 - policy name
# $2 - plugin path

obselete - > obsolete 

@revit13 do you know something about "dataset credentials path become obsolete "?

roytman (approved):

PR #1779, opened 2022-10-25
Change cache key in build.yml

revit13:
This PR changes the [cache key](https://github.com/actions/cache) in build.yml workflow, which is an explicit key for restoring and saving the cache, to depend also on the versions of the tools.


.github/workflows/build.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           ~/go/pkg/mod
           ${{ github.workspace }}/hack/tools/bin
           ${{ github.workspace }}/hack/tools/lib
-        key: ${{ runner.os }}-c-${{ hashFiles('hack/make-rules/tools.mk') }}-go-${{ hashFiles('**/go.sum') }}
+        key: ${{ runner.os }}-c-${{ hashFiles('hack/make-rules/tools.mk') }}-${{ hashFiles('hack/tools/requirements.sh') }}-go-${{ hashFiles('**/go.sum') }}
```
roytman, line 42: the key was defined 9 times in the file, can we replace this long line by a single env variable ?

revit13: I think not, please see https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability for information where `hashFiles` and `runner.os` can be used in the workflow.

roytman: You are right.

roytman (approved):

PR #1781, opened 2022-10-25
Documentation explaining how to discover an S3 asset in OpenMetadata, and how to enrich it.

cdoron:


revit13:
@cdoron consider moving the page to tasks instead of sample. 
Also, as most of the details are not related to Fybrik, consider replacing them with specific links to openmetada data site so the burden of updating the instructions upon a change will be avoided.

cdoron:
> @cdoron consider moving the page to tasks instead of sample. Also, as most of the details are not related to Fybrik, consider replacing them with specific links to openmetada data site so the burden of updating the instructions upon a change will be avoided.

Regarding your second comment, I think links to an OpenMetadata site would not be enough, because:

1. the creation of `s3` assets is not straightforward
2. some information here, such as information related to the custom properties, is Fybrik-specific

site/docs/samples/omd-discover-s3-asset.md
```diff
@@ -0,0 +1,61 @@
+# Discovering an S3 asset through the OpenMetadata UI
+
+This page explains how to discover an existing S3 asset through the OpenMetadata UI. The screenshots refer to `localstack` cloud storage, but the explanations also apply to other S3 services.
+
+Begin by opening your browser to the OpenMetadata UI. If you installed OpenMetadata in your kubernetes cluster in the `open-metadata` namespace, go to http://localhost:8585 after running:
+```
```
aradhalevy, line 6: in order to make http://localhost:8585 a link when running in the site,  wrap it in [ ] and add ( link )

site/docs/samples/omd-discover-s3-asset.md
```diff
@@ -0,0 +1,61 @@
+# Discovering an S3 asset through the OpenMetadata UI
+
+This page explains how to discover an existing S3 asset through the OpenMetadata UI. The screenshots refer to `localstack` cloud storage, but the explanations also apply to other S3 services.
```
aradhalevy, line 3: maybe a link to the read sample where the there is an option to make a S3 object can be helpful here.
something like : "existing S3 asset (to add a s3 assest refer to the read...) through the OpenMetadata UI"

aradhalevy (commented):
I did the flow and it works great, also looks good in the site.

right now it is not in the site, to add this as a sample, add it to https://github.com/fybrik/fybrik/blob/master/site/mkdocs.yml under Notebook samples

aradhalevy (approved):

roytman (approved):

PR #1782, opened 2022-10-26
Replace datashim with minio sdk

shlomitk1:
- This PR uses Minio SDK to allocate/delete buckets and removes all references to datashim and Dataset custom resources.
- Changes to FybrikApplication status:
    - no need for DatasetRef since the resource does not exist
    - added an indication whether the storage should be freed upon deletion of FybrikApplication, or it is a persistent storage. 
    Previously, a label inside Dataset resource has been used for this purpose.
- No "late" checks of status before plotter creation. In case of allocation errors, no data path is built.

**This PR does not include:**
- enhancing FybrikStorageAccount to support not only S3
- taxonomy fixes 

**Tested the following cases:**
* write a new dataset, delete the app - the bucket is not deleted
* temporary copy, delete the app   - the bucket is deleted (tested on empty + non-empty buckets)
* http vs. https
* copy (ingest) - opened a bug report for RegisterAsset fail.


revit13:
Minor comment - samples/multicluster-razee/README.md should be updated as well

charts/fybrik-crd/templates/app.fybrik.io_fybrikapplications.yaml
```diff
@@ -247,9 +247,6 @@ spec:
                   additionalProperties:
                     description: DatasetDetails holds details of the provisioned storage
                     properties:
-                      datasetRef:
```
roytman, line 250: you changed the CRD `Status`. if there are CRD instances with this field, users will get an error.

manager/apis/app/v1beta1/fybrikapplication_types.go
```diff
@@ -116,10 +116,6 @@ type SecretRef struct {
 
 // DatasetDetails holds details of the provisioned storage
 type DatasetDetails struct {
-	// Reference to a Dataset resource containing the request to provision storage
```
roytman, line 119: like the comment above, we have a runtime backward compatibility issue.

manager/controllers/app/plotter_generator.go
```diff
@@ -47,44 +46,21 @@ type PlotterGenerator struct {
 	ProvisionedStorage map[string]NewAssetInfo
 }
 
-// AllocateStorage creates a Dataset for bucket allocation
+// AllocateStorage creates a new connection based on the selected account
 func (p *PlotterGenerator) AllocateStorage(item *datapath.DataInfo, destinationInterface *taxonomy.Interface,
 	account *fapp.FybrikStorageAccountSpec) (*fapp.DataStore, error) {
 	// provisioned storage
-	var genBucketName, genObjectKeyName string
+	var genName string
```
roytman, line 53: pls add a comment that it's a `datasetName`.
BTW, it can be confused, in the `CreateConnection` method signature, this argument is called `datasetName`, but it is stored into `object_key`. Maybe better to call it `object_key`?

charts/fybrik-crd/templates/app.fybrik.io_fybrikapplications.yaml
shlomitk1: Restored the field.

manager/apis/app/v1beta1/fybrikapplication_types.go
shlomitk1: Restored the field

manager/controllers/app/plotter_generator.go
shlomitk1: changed to genName - it is a generated name representing a copied dataset, not necessarily an object key.

shlomitk1:
> Minor comment - samples/multicluster-razee/README.md should be updated as well

Updated.

shlomitk1:
> Signed-off-by: Shlomit Koyfman [shlomitk@il.ibm.com](mailto:shlomitk@il.ibm.com)
> 
> * This PR uses Minio SDK to allocate/delete buckets and removes all references to datashim and Dataset custom resources.
> * Changes to FybrikApplication status:
>   
>   * no need for DatasetRef since the resource does not exist
>   * added an indication whether the storage should be freed upon deletion of FybrikApplication, or it is a persistent storage.
>     Previously, a label inside Dataset resource has been used for this purpose.
> * No "late" checks of status before plotter creation. In case of allocation errors, no data path is built.
> 
> **This PR does not include:**
> 
> * enhancing FybrikStorageAccount to support not only S3
> * taxonomy fixes

**Tested the following cases:**
* write a new dataset, delete the app - the bucket is not deleted
* temporary copy, delete the app   - the bucket is deleted (tested on empty + non-empty buckets)

go.mod
```diff
@@ -48,6 +48,13 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require (
```
roytman, line 51: why did you add this set as a separate require block?
`github.com/minio/minio-go/v7 v7.0.39` should be in the main (above) require block. All others will be automatically added in the second (// indirect) block.

shlomitk1: I did not make any manual changes, something went wrong during "go mod tidy" - I've removed downloaded packages and re-ran.

leseb:
Hello ✋🏻 , out of curiosity, why is datashim being removed in favor of minIO? Thanks!

shlomitk1:
> Hello ✋🏻 , out of curiosity, why is datashim being removed in favor of minIO? Thanks!

Hi @leseb, Fybrik aims to support other storage/connection types besides s3 - please see https://github.com/fybrik/fybrik/blob/master/docs/StorageManagement.md for more details.
In order to manage s3 storage (allocate a bucket, delete a bucket, delete all files in a bucket with a given prefix) Fybrik does not need datashim at its full capacity, the direct sdk is enough. Once there were thoughts of using datashim as an alternative to a copy module for giving access to remote S3 data inside a read server pod, but they were never carried out.

leseb:
> > Hello ✋🏻 , out of curiosity, why is datashim being removed in favor of minIO? Thanks!
> 
> Hi @leseb, Fybrik aims to support other storage/connection types besides s3 - please see https://github.com/fybrik/fybrik/blob/master/docs/StorageManagement.md for more details. In order to manage s3 storage (allocate a bucket, delete a bucket, delete all files in a bucket with a given prefix) Fybrik does not need datashim at its full capacity, the direct sdk is enough. Once there were thoughts of using datashim as an alternative to a copy module for giving access to remote S3 data inside a read server pod, but they were never carried out.

Thanks a lot for the prompt answer. I'll go and read that link as well. Thanks!

shlomitk1:
replaced by #1903

PR #1788, opened 2022-10-30
add OM_SERVER_USERNAME and OM_SERVER_PASSWORD environment variables in vault pod

cdoron:

### DIFF
--- a/charts/vault/env/dev/plugin-secrets-values.yaml
+++ b/charts/vault/env/dev/plugin-secrets-values.yaml
@@ -10,6 +10,8 @@ vault:
         readOnly: false
     extraEnvironmentVars:
       OM_SERVER_URL: http://openmetadata.open-metadata:8585/api
+      OM_SERVER_USERNAME: admin
+      OM_SERVER_PASSWORD: admin
     extraInitContainers:
       - name: kubesecrets-plugin
         image: ghcr.io/fybrik/vault-plugin-secrets-kubernetes-reader:0.5.0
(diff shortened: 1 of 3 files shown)


charts/vault/env/dev/plugin-secrets-values.yaml
```diff
@@ -10,6 +10,8 @@ vault:
         readOnly: false
     extraEnvironmentVars:
       OM_SERVER_URL: http://openmetadata.open-metadata:8585/api
+      OM_SERVER_USERNAME: admin
```
revit13, line 13: might be better to read this from a secret:
```
extraSecretEnvironmentVars: 
- envName: OM_SERVER_USERNAME
   secretName: ...
```

shlomitk1:
I don't like the idea of having OM specific data in a general installation of vault. What if I configure Fybrik to work with a different catalog, why should the OM data be there by default?
It would make more sense to leave chart/vault/env/dev as before, with no reference to OM or the related plugins, create a new directory e.g., chart/vault/env/openmetadata/ and copy files from there to the default folder before deploying Fybrik+vault with openmetadata.

revit13:
> I don't like the idea of having OM specific data in a general installation of vault. What if I configure Fybrik to work with a different catalog, why should the OM data be there by default? It would make more sense to leave chart/vault/env/dev as before, with no reference to OM or the related plugins, create a new directory e.g., chart/vault/env/openmetadata/ and copy files from there to the default folder before deploying Fybrik+vault with openmetadata.

 `chart/vault/env` are meant to contain directories for the different Vault modes, for example: `dev` and `ha` which I have in a [different pr](https://github.com/fybrik/fybrik/pull/1738).
The way to deploy vault with different values.yamj file is to redefine `VAULT_VALUES_FILE` in the [Makefile](https://github.com/fybrik/fybrik/blob/master/third_party/vault/Makefile) upon installation.
I am not sure I understand your suggestion. Let's talk.

PR #1796, opened 2022-11-08
Add notebook test with openmetadata catalog.

revit13:
This PR contains read notebook sample which uses openmetadata as the catalog.
It is triggered in github workflow but if decided it can be triggered only upon release as the runtime of deploying openmetadata is about 20 minutes.


Makefile
```diff
@@ -123,7 +123,7 @@ run-notebook-readflow-tests:
 run-notebook-readflow-tests-om: export HELM_SETTINGS=--set "coordinator.catalog=openmetadata"
 run-notebook-readflow-tests-om: export VALUES_FILE=charts/fybrik/notebook-test-readflow.values.yaml
 run-notebook-readflow-tests-om: export DEPLOY_OPENMETADATA=1
-run-notebook-readflow-tests-om: export CATALOGED_ASSET="openmetadata-s3.default.bucket1.\"data.csv\""
+run-notebook-readflow-tests-om: export CATALOGED_ASSET=openmetadata-s3.default.bucket1."data.csv"
```
shlomitk1, line 126: Why are there quotes inside a name?

revit13: @cdoron maybe you can answer that.. tnx (its also in https://fybrik.io/dev/samples/notebook-read/)

cdoron: @shlomitk1: Asset (table) names in OM look like this: A.B.C.D, where `A` is the databaseService, `B` is the database, `C` is the databaseSchema, and `D` is the table name. This gets confusing when there are `.` within the table name.
To avoid confusion, if the table name has `.` in it, then quotes are added

roytman (approved):

PR #1797, opened 2022-11-08
updated the architecture image on the site

aradhalevy:
updated the architecture image on the site with roles of different actors, and a few extra fixes

Signed-off-by: Arad Halevy <arad.halevy@ibm.com>


roytman (commented):
I think we need to add several words with explanation

aradhalevy:
> I think we need to add several words with explanation

I updated the image as we discussed.

Also, added in the text (not on the image) when it says:

data operator -> data operator (e.g. IT admin)

data user -> data user (e.g. data scientist)

data steward ->   data steward (e.g. data governance officer)

do you think it is okay?

roytman (approved):

roytman:
Thanks @aradhalevy

PR #1803, opened 2022-11-13
Minor changes to OpenMetadata docs

simanadler:
There were some mentions of katalog still, and some comments about storing dataset credentials in kubernetes secrets


roytman (approved):

site/docs/samples/pre-steps.md
```diff
@@ -1,7 +1,7 @@
 # Tools used by the actors
 
 - The data owner would typically register the dataset in a proprietary or open source catalog. We use [OpenMetadata](https://open-metadata.org/).
-- The data owner needs to provide credentials for accessing a dataset. One way to do it is by storing credentials in kubernetes secrets.
+- The data owner needs to provide credentials for accessing a dataset. This is usually done via the data catalog, but credentials could be stored in kubernetes secrets as an alternative.
```
cdoron, line 4: In most of the scenarios we describe, the credentials are stored in a kubernetes secret at first. Then an asset is created through a connector `createAsset` request. The connector extract the credentials from vault (through the k8s-secrets plugin), and then places those secrets in OpenMetadata.
The change you propose suggest that the credentials are stored in either the data catalog or in a kubernetes secret, when in fact they could be stored in both.

simanadler: In a flow where the data owner is contributing a dataset (the dataset is not being created via Fybrik) why would a data owner even know that there is such a thing as Fybrik or a Fybrik connector?  He would just register the asset and its credentials in the data catalog, wouldn't he?

cdoron: That's right. If the asset is created through the UI, then the credentials would appear only in OpenMetadata.
The scenarios for which the credentials may also appear in a k8s secrets are when either the user or Fybrik creates the asset through the connector.

PR #1826, opened 2022-11-23
Disable tests for draft mode.

revit13:
This PR disables some tests when the PR is in draft mode.


shlomitk1:
Can you be a little more specific about what tests are disabled and why.

revit13:
@shlomitk1 it disables tekton-pipelines and any test that is run in old k8s_version, for example, integration tests and notebook tests. those tests will only run once with new k8s_version.

roytman (approved):

PR #1829, opened 2022-11-24
preventing OR-Tools from writing logs to /tmp

### LINKED ISSUE #1828: Prevent OR-Tools CSP engine from filling /tmp with log files

zivnevo:
Fixes #1828 


pkg/optimizer/optimizer.go
```diff
@@ -55,9 +56,13 @@ func (opt *Optimizer) getSolution(pathLength int) (string, error) {
 		return "", errors.Wrap(err, "error building a model")
 	}
 
-	opt.log.Debug().Msgf("Executing %s %s", opt.solverPath, modelFile)
+	solverArgs := []string{modelFile}
+	if strings.Contains(opt.solverPath, "or-tools") { // OR-Tools-specific flags
+		solverArgs = append(solverArgs, "--logtostderr") // avoid filling /tmp with log files
+	}
```
shlomitk1, line 62: We want the code to be unaware of a specific setup. I think the best place to add the argument is in this section:
https://github.com/fybrik/fybrik/blob/5af2334371fb560fd3d4632cd8975cdba1a4fb9d/charts/fybrik/values.yaml#L276-L284
There are several possible approaches. One is to insert a new env var into fybrik_config configmap and use it to form a command in the code. Another is to replace CSP_PATH in manager deployment with CSP_CMD that will be a template of the command, like <csp_exe> %s <extra parms from the values yaml>. The advantage of the latter is that if we replace the solver with another one that has a different way to pass the input model (-i <model>, for example), we could easily do it.

shlomitk1:
Replacing by https://github.com/fybrik/fybrik/pull/1916

PR #1839, opened 2022-12-05
Use retryable http client in fybrikUtils.GetHTTPClient.

revit13:
This PR addresses https://github.com/fybrik/openmetadata-connector/issues/4 which asks to replace net/http with another http client with timeouts.


roytman (commented):
@revit13 the issue doesn't have a PR, so please add some explanation into the issue header. For example, what do you want to resolve, you don't change the default parameters of the retryablehttp client, so make sense to add some comments, maybe the https://github.com/hashicorp/go-retryablehttp/blob/493aa4cf372e47ec00228558d9d91a9517b045a5/client.go#L50 link. Should we allow to configure the parameters?
One of the parameters of retryablehttp client is [logger](https://github.com/hashicorp/go-retryablehttp/blob/493aa4cf372e47ec00228558d9d91a9517b045a5/client.go#L407), please try to set it.

revit13:
@roytman I added a description. Thanks

pkg/connectors/datacatalog/clients/datacatalog_openapi.go
```diff
@@ -45,7 +45,7 @@ func NewOpenAPIDataCatalog(name, connectionURL string) DataCatalog {
 			},
 		},
 		OperationServers: map[string]openapiclient.ServerConfigurations{},
-		HTTPClient:       utils.GetHTTPClient(&log),
+		HTTPClient:       utils.GetHTTPClient(&log).StandardClient(),
```
shlomitk1, line 48: I see that GetHTTPClient is not used without StandardClient(), so why not to change the utils.GetHTTPClient to return the final object?

revit13: In opametadata-connector code which uses this function not all calls will use StandardClient() (for example in Vault client). The above code needs to use StandardClient() because it uses autogenrated code.

shlomitk1: It is a problem. Whoever is looking at the code in fybrik repo has no idea that the same code is also used elsewhere in a different flavor. A cleaner way would be to create a separate package (with its own go.mod). It does not make sense for openmedata-connector to import the entire fybrik just to use a package or two.

revit13: @shlomitk1 OK, we can also create a separate package for vault and tls pkgs which the modules can also use. Discussing this with @roytman we might want to do that after the packages will be stabilized (there are functions in vault packages that need to be moved from openmetadata repo)

shlomitk1 (approved):

roytman (approved):

Issue #1525, opened 2022-06-08
Problems with the solver on Mac


roytman:
The #1497 added google `or-tools` (fzn-or-tools) to Fybrik.  There are several issues with it:

[ ] the `hack/tools/install_or_tools.sh` installs it into the hack/tools/bin based on the host OS (Linux or MacOS) but after that the binary are [copied](https://github.com/fybrik/fybrik/blob/a7fa6eee7c3e452ca5f681ad13cd08c9315f23ee/pkg/optimizer/Makefile#L17) into the RHEL Docker.  I doubt that Mac version will run there.
[ ] the  `hack/tools/install_or_tools.sh`  has a bug for MacOS - the downloaded file is or-tools`__`flatzinc_MacOsX-12.2.1_v9.2.9972.tar.gz, which is not exist.
[ ] Finally, meantime, `or-tools` doesn't support MacOS with apple CPU. it's targeted to the next release only, or there are attempts (not always successful) [to compile it]( https://github.com/google/or-tools/issues/2332.).  Therefore, I cannot run `make install-tools` on my laptop. 

Can it be part of a docker  only?

shlomitk1:
It can't be a part of a docker only, as it is tested mainly in the unit tests (running binary).
We still need to run the binary in CI.
I think we need:
1. not to create local Docker images for each test run but rather use the global image `ghcr.io/fybrik/optimizer:master` 
2. not to install or-tools for specific platforms
3. Empty CSP_PATH if the solver is not installed in (1)

roytman:
I know that it is part of the manual tests, but it can be integrated into a docker similar as we run openapi generator. https://github.com/fybrik/fybrik/blob/a7fa6eee7c3e452ca5f681ad13cd08c9315f23ee/site/Makefile#L25

shlomitk1:
> I know that it is part of the manual tests, but it can be integrated into a docker similar as we run openapi generator.
> 
> https://github.com/fybrik/fybrik/blob/a7fa6eee7c3e452ca5f681ad13cd08c9315f23ee/site/Makefile#L25

Not sure I understand. Do you suggest to run `go test` for the manager inside a container?

roytman:
if possible, yes

shlomitk1:
[...]

PR #1840, opened 2022-12-05
Mac support for or_tools

zivnevo:
Also upgrading to OR Tools version 9.5

Should solve some of the issues in #1525


shlomitk1 (commented):
Currently, the optimizer image is hard-coded to **image: "ghcr.io/fybrik/optimizer:0.0.1"** which was built quite a while ago. In order to proper test the Dockerfile, the following is required:
- replace solver.image with "optimizer" for all charts/fybrik/values*.yaml 
- add docker-build for optimizer in Makefile (where manager is built)
- add docker-push for optimizer in the git push action

zivnevo:
> Currently, the optimizer image is hard-coded to **image: "ghcr.io/fybrik/optimizer:0.0.1"** which was built quite a while ago. In order to proper test the Dockerfile, the following is required:
> 
>     * replace solver.image with "optimizer" for all charts/fybrik/values*.yaml
> 
>     * add docker-build for optimizer in Makefile (where manager is built)
> 
>     * add docker-push for optimizer in the git push action

I'm not sure I understand which tests take which image from which location. In practice, `Tekton Tests` passed only after I modified the Dockerfile to run the solver install script.
How about a new issue for the changes you proposed?

shlomitk1:
> 
> I'm not sure I understand which tests take which image from which location. In practice, `Tekton Tests` passed only after I modified the Dockerfile to run the solver install script. How about a new issue for the changes you proposed?

Tekton tests do not use the solver image, they use the local installation instead. I believe that they pass because you changed the install script. The changes to Dockerfile do not have any effect at all because "make docker-build docker-push" is never called for the optimizer. 
My concern is that your changes to Dockerfile have not been tested because of the hard-coded old image that was created in March. If you remove Dockerfile and Makefile from this PR, it's OK to have it merged.

shlomitk1 (approved):

PR #1904, opened 2023-01-17
Control testing verbosity option

shlomitk1:
Fixes https://github.com/fybrik/fybrik/issues/1861

Set TEST_VERBOSITY to 1 in order to get tests with full verbosity, by default, it's 0.
Details will be printed only for failing tests.


Makefile.env
```diff
@@ -16,3 +16,10 @@ GOBIN=$(shell go env GOBIN)
 endif
 
 GOPATH=$(shell go env GOPATH)
+
+# If true, run tests using -v option (verbosity)
+export TEST_VERBOSITY ?= 0
+
```
roytman, line 22: I suggest defining the `TEST_OPTIONS` as a separate variable, we even can get it for environment, so it will allow as to provide additional arguments to the tests. For example, `-timeout` or `-json`

manager/controllers/app/Makefile
```diff
@@ -3,7 +3,7 @@ include $(ROOT_DIR)/Makefile.env
 
 .PHONY: test
 test: *.go
-	go test -v
+	go test $(TEST_OPTIONS)
```
roytman, line 6: should we add `$(TEST_OPTIOS)` to the line 15 ?

Makefile.env
shlomitk1: ok

manager/controllers/app/Makefile
shlomitk1: added

revit13 (approved):

PR #1907, opened 2023-01-18
Remove hard-coded storage type selection

shlomitk1:
Related to https://github.com/fybrik/fybrik/issues/1864
This PR does not include solver changes.


pkg/optimizer/datapath_csp.go
```diff
@@ -999,7 +998,7 @@ func arrayOfSameInt(num, arrayLen int) []string {
 
 func getAssetInterface(connection *datacatalog.GetAssetResponse) taxonomy.Interface {
 	if connection == nil || connection.Details.Connection.Name == "" {
-		return taxonomy.Interface{Protocol: utils.GetDefaultConnectionType(), DataFormat: ""}
+		return taxonomy.Interface{Protocol: "", DataFormat: ""}
```
zivnevo, line 1001: This is not going to work, as the code currently does not handle `Protocol == ""` as a special case.
What should be the semantics? Should `""` match any protocol?

manager/controllers/app/solve_test.go
```diff
@@ -510,6 +510,70 @@ func TestWriteNewAsset(t *testing.T) {
 	g.Expect(solution.DataPath[0].Module.Name).To(gomega.Equal(writeModule.Name))
 }
 
+// This test checks the write scenario for a new asset.
+// Storage account is for MySQL, module supports S3 only.
+// Result: failure to create the data path
+func TestStorageAndModuleMismatch(t *testing.T) {
+	t.Parallel()
+	g := gomega.NewGomegaWithT(t)
```
zivnevo, line 518: The 2 new tests are not going to pass with `USE_CSP=True` until I push my changes.
Maybe leave this file as-is, and I will commit the modified file with my changes?

pkg/optimizer/datapath_csp.go
shlomitk1: Yes, the semantics is to skip the check. This is the use-case of writing a new asset when the destination is not known in advance but is rather dictated by the module capabilities and available storage.

manager/controllers/app/solve_test.go
shlomitk1: OK

pkg/optimizer/datapath_csp.go
zivnevo: Then this requires a few more check. For now, just replace `utils.GetDefaultConnectionType()` with `utils.S3`

manager/controllers/app/solve_test.go
shlomitk1: @zivnevo I've added the check for USE_CSP set to true, in which case the tests won't run.

pkg/optimizer/datapath_csp.go
shlomitk1: OK

zivnevo (approved):

PR #1911, opened 2023-01-19
Fixes install_or_tools.sh

### LINKED ISSUE #1909: The install_or_tools.sh script checks wrong file on MAC
When the `install_or_tools.sh` script runs, it checks that the `bin/fzn-or-tools` and `lib/libfz.so` files exist. However, on MAC there is no `lib/libfz.so` file, and as a result, the script always installs the tool.
And it doesn't check the installed version

roytman:
When the `install_or_tools.sh` script runs, it checks that the bin/fzn-or-tools and lib/libfz.so files exist. 
However, on MAC there is no lib/libfz.so file, and as a result, the script always installs the tool.

In addition, it doesn't check the installed version

/fixes #1909 


hack/tools/install_or_tools.sh
```diff
@@ -6,6 +6,8 @@
 cd "${0%/*}"
 source ./common.sh
 
+BUILD_FILE=lib/build
+VERSION_FILE=lib/version
```
zivnevo, line 10:
```suggestion
BUILD_FILE=lib/fzn-or-tools-build
VERSION_FILE=lib/fzn-or-tools-version
```
Maybe have more explicit file names, just in case another tool will also use the `lib` directory one day.

hack/tools/install_or_tools.sh
```diff
@@ -37,3 +39,5 @@ tar -zxvf ./${download_file} -C $tmp
 mv $tmp/*/bin/fzn-ortools ./bin/fzn-or-tools
 mv $tmp/*/lib*/lib*.${dyn_lib_ext}* ./lib
 chmod +x ./bin/fzn-or-tools
+echo ${OR_TOOLS_BUILD} > lib/build
+echo ${OR_TOOLS_VERSION} > lib/version
```
zivnevo, line 43:
```suggestion
echo ${OR_TOOLS_BUILD} > ${BUILD_FILE}
echo ${OR_TOOLS_VERSION} > ${VERSION_FILE}
```

zivnevo (dismissed):
Some minor comments. Feel free to merge without them.

shlomitk1 (dismissed):

zivnevo (approved):

hack/tools/install_or_tools.sh
```diff
@@ -6,6 +6,8 @@
 cd "${0%/*}"
 source ./common.sh
 
+BUILD_FILE=lib/fzn-or-tools-build
+VERSION_FILE=lib/fzn-or-tools-version
```
shlomitk1, line 10: Suggest to change to ./solver-build and ./solver-version

roytman:
TODO check with @ngoracke why tools are installed on Ubunty, but runs on Centos

hack/tools/install_or_tools.sh
roytman: I prefer to leave the tool related name (fzn-or-tools), so it will be easy to connect the installation script and the created file.

shlomitk1 (approved):

PR #1926, opened 2023-01-23
add explanations on working with katalog

cdoron:
quickstart + read sample

### DIFF
--- a/site/docs/samples/notebook-read.md
+++ b/site/docs/samples/notebook-read.md
@@ -61,20 +61,97 @@ Make a note of the service endpoint, bucket name, and access credentials. You wi
 
 In this step you are performing the role of the data owner, registering his data in the data catalog and registering the credentials for accessing the data in the credential manager.
 
-In this tutorial, we assume that OpenMetadata is used as the data catalog.
+=== "With OpenMetadata"
+    Datasets can be registered either directly, through the OpenMetadata UI, or indirectly, through the data-catalog connector:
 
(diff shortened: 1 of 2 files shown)


site/docs/samples/notebook-read.md
```diff
@@ -142,16 +219,16 @@ Datasets can be registered either directly, through the OpenMetadata UI, or indi
...
     ```bash
-    CATALOGED_ASSET="openmetadata-s3.default.demo.\"PS_20174392719_1491204439457_log.csv\""
+    CATALOGED_ASSET="fybrik-notebook-sample/paysim-csv-ab123"
     ```
 
-    If you look at the asset creation request above, you will notice that in the `resourceMetadata` field, we request that the asset should be tagged with the `Purpose.finance` tag, and that three of its columns should be tagged with the `PII.Sensitive` tag. Those tags will be referenced below in the access policy rules. Tags are important because they are used to determine whether an application would be allowed to access a dataset, and if so, which transformations should be applied to it.
+If you look at the asset creation request above, you will notice that in the `resourceMetadata` field, we request that the asset should be tagged with the `Purpose.finance` tag, and that three of its columns should be tagged with the `PII.Sensitive` tag. Those tags will be referenced below in the access policy rules. Tags are important because they are used to determine whether an application would be allowed to access a dataset, and if so, which transformations should be applied to it.
```
roytman, line 231: Looks like there are wrong instructions for Katalog 
comparing with https://fybrik.io/v1.1/samples/notebook-read/#register-the-dataset-in-a-data-catalog

roytman (approved):

Issue #1931, opened 2023-01-23
Release a new optimizer image, based on OR-tools v9.5


zivnevo:
- [x] Change the Dockerfile to run the `install_or_tools.sh` script during Docker build.
- [x] Properly handle image tag
- [x] Upload a new image to the Fybrik registry
- [x] Test the new image

PR #1932, opened 2023-01-23
Use install script inside Dockerfile

zivnevo:
This makes sure we take the correct CSP executable, which fits the image base OS.


zivnevo:
First task in #1931

PR #1934, opened 2023-01-23
update notebook write scenario with katalog instructions

cdoron:


roytman (commented):
can you change the line 460: 
"Execute all notebook cells and notice that the `oldbalanceOrg` column does not appear because it was redacted."
to someting like
Execute all notebook cells and notice that data in the `oldbalanceOrg` column does was redacted.

roytman (approved):

PR #1942, opened 2023-01-24
New optimizer docker image

shlomitk1:
Closes https://github.com/fybrik/fybrik/issues/1931


hack/tools/install_or_tools.sh
```diff
@@ -39,4 +39,5 @@ tar -zxvf ./${download_file} -C $tmp
 mv $tmp/*/bin/fzn-ortools ./bin/fzn-or-tools
 mv $tmp/*/lib*/lib*.${dyn_lib_ext}* ./lib
 chmod +x ./bin/fzn-or-tools
+cp ./bin/fzn-or-tools ./bin/solver
```
zivnevo, line 42: Why do we need two copies of the solver executable?

shlomitk1: Leaving only one.

zivnevo (approved):

PR #1944, opened 2023-01-24
Design document for the Network Policies based isolation

roytman:
Design document for the Network Policies based isolation 

/fixes https://github.com/fybrik/fybrik/issues/1962


docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
+# Isolation 
+
+_**Note:**_ below we start with Network Polices as an isolation method, lather we will add another methods.
```
shlomitk1, line 3: later

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+In the current Fybrik implementation the entry point of a data plane is published as a part of the FybrikApplication 
+status, it is completely non-protected, and as result everyone can connect to the data plane. 
+In “Read” flow cases it leads to governance/security violations, “Write” and “Delete” flows can create severe accidents, 
+such as unauthorized data changes or deletions. Given that we need to prevent unauthorized data plane access. 
+
+_Note:_ the same situation exists between chained modules.  
+
+## Challenge
+
+When we talk about access Fybrik data plane, we are talking about connectivity between a user workload, which is out of
```
shlomitk1, line 16: accessing

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+data plane isolation is shared between Fybrik core, Fybrik users, module developers and probably IT teams.
+
+- Fybrik users should provide all relevant information to Fybrik and correctly configure their workloads. For example, 
+to set labels on the workload and inform Fybrik about the labels.
+
+- Depends on the isolation method, Fybrik should propagate all relevant information to the deployed modules, and/or 
+correctly configure Kubernetes Network Policies.
+- Module developers should allow to protect the module instances. For example, to allow TLS support
+
+In some complicated cases, IT teams interfere can be required, e.g. to configure external access points.
```
shlomitk1, line 28: Why is IT team interference required? In which cases?

docs/IsolationNetworkPolicies.md
roytman: fixed

docs/IsolationNetworkPolicies.md
roytman: thanks, fixed

docs/IsolationNetworkPolicies.md
roytman: as I wrote to configure external access points

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+status, it is completely non-protected, and as result everyone can connect to the data plane. 
+In “Read” flow cases it leads to governance/security violations, “Write” and “Delete” flows can create severe accidents, 
+such as unauthorized data changes or deletions. Given that we need to prevent unauthorized data plane access. 
+
+_Note:_ the same situation exists between chained modules.  
+
+## Challenge
+
+When we talk about accessing Fybrik data plane, we are talking about connectivity between a user workload, which is out of 
+Fybrik scope, and Fybrik module, which can be developed by a third party. Therefore, in this document (and in its
```
simanadler, line 17: I agree that the highest priority is to provide isolation between the workload and the module(s) with which it interfaces.  However, I believe we also need to provide isolation between modules - even those that do not interact directly with the workload.  Right?

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+In some complicated cases, IT teams interfere can be required, e.g. to configure external access points.
+
+## Requirements
+
+- Only predefined users/workloads should be able to access the data plane.
+- We do not assume that client workloads will be collocated on the same Kubernetes cluster with the data module, 
+furthermore they can run out of Kubernetes at all. 
+  - However, we do check the collocation of a Fybrik module and a relevant workload in the same cluster as a specific use case. 
+- We should allow workload client connections from different IP addresses. 
+
```
simanadler, line 37: Isolation of modules in the data plane.  I.e. Modules should not be able to interface "illegally" with modules outside of the defined data plane(s)

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+## Network Policies
+
+Fybrik is a Kubernetes application, therefore, the simplest isolation method can be based on 
+[Kubernetes Network Policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) (NP).
+NP allow to control traffic flow at the IP address or port level (OSI layer 3 or 4). If a user workload or a Fybrik 
+previous chained module collocated with the next module in the same Kubernetes cluster, NPs can help to restrict the Fybrik 
+module ingress access. However, when a workload or a previous chained module runs out of the cluster, the NP based 
+isolation is going to be more complicated. The collocated deployment can be based on Kubernetes labels or namespaces, but 
+cross-clusters connections can be based on requester IP addresses, which is not always easy to discover. 
+Therefore, later we will provide other isolation methods as an extension to NP.
```
simanadler, line 48: What about egress?

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Advantages:
+- does not require any changes in modules implementations, can be implemented in the Fybrik control plane only.
+- can be combined with other isolation methods
+
+### Disadvantages:
+- complicated in cross clusters deployments.
+- NP are implemented by the network plugin. Creating a NetworkPolicy resource without a controller - a Container Network
```
simanadler, line 56: What is "the network plugin"?

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+- Only predefined users/workloads should be able to access the data plane.
+- We do not assume that client workloads will be collocated on the same Kubernetes cluster with the data module, 
+furthermore they can run out of Kubernetes at all. 
+  - However, we do check the collocation of a Fybrik module and a relevant workload in the same cluster as a specific use case. 
+- We should allow workload client connections from different IP addresses. 
+
+
+## Network Policies
+
+Fybrik is a Kubernetes application, therefore, the simplest isolation method can be based on
```
simanadler, line 41: What are the other options, other than network policies?  Why were they ruled out?

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+- complicated in cross clusters deployments.
+- NP are implemented by the network plugin. Creating a NetworkPolicy resource without a controller - a Container Network 
+Interface (CNI) that implements it will have no effect. Default Kind deployment which by default uses "kindnetd” doesn’t 
+support NP. It requires installation another CNI, e.g. [Calico](https://github.com/projectcalico/calico). Here is the 
+[instructions](https://alexbrand.dev/post/creating-a-kind-cluster-with-calico-networking/). All K8s clusters in 
+production deployments use CNIs which support NP.
+- requires information about workload deployment, and/or setting labels to the workload
+
+## Implementation
+
```
simanadler, line 64: This implementation is for isolating the workload from the module with which it interfaces.  Doesn't address isolating modules in the data plane I assume

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+The “from” element of ingress 
+[NetworkPolicySpec](https://kubernetes.io/docs/reference/kubernetes-api/policy-resources/network-policy-v1/#NetworkPolicySpec) 
+allows to restrict incoming network traffic based on “podSelectors” (labels based), and/or 
+“namespaceSelectors” (labels based) and/or “ipBlocks”. The last one permits incoming network traffic from predefined IP 
+address blocks.
+
+Current “FybrikApplicationSpec” has the “[selector](https://fybrik.io/v1.2/reference/crds/#fybrikapplicationspecselector)” 
+element, which is a combination of “clusterName” and “workloadSelector”. Unfortunately, based on this information we can 
+implement only one label based “podSelector”. Furthermore, for the backward compatibility of the current implementation
```
simanadler, line 73: Why is one label a problem?

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+The “from” element of ingress 
+[NetworkPolicySpec](https://kubernetes.io/docs/reference/kubernetes-api/policy-resources/network-policy-v1/#NetworkPolicySpec) 
+allows to restrict incoming network traffic based on “podSelectors” (labels based), and/or 
+“namespaceSelectors” (labels based) and/or “ipBlocks”. The last one permits incoming network traffic from predefined IP 
+address blocks.
+
+Current “FybrikApplicationSpec” has the “[selector](https://fybrik.io/v1.2/reference/crds/#fybrikapplicationspecselector)” 
+element, which is a combination of “clusterName” and “workloadSelector”. Unfortunately, based on this information we can 
+implement only one label based “podSelector”. Furthermore, for the backward compatibility of the current implementation 
+assumes that a Read scenario should have at least one label, which is not always true.
```
simanadler, line 74: In what read scenario do we not have a label for the workload?  I believe that ingest and maybe delete are the only scenarios where there is no workload label.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+element, which is a combination of “clusterName” and “workloadSelector”. Unfortunately, based on this information we can 
+implement only one label based “podSelector”. Furthermore, for the backward compatibility of the current implementation 
+assumes that a Read scenario should have at least one label, which is not always true. 
+
+### Suggetions:
+
+- extend the “FybrikApplicationSpec.selector” element with an array of 
+[NetworkPolicyPeers](https://github.com/kubernetes/api/blob/59fcd23597fd090dba6b7e903eb0a8c9e8efb0a6/networking/v1/types.go#L183)
+this “NetworkPoliciesPeers” array will be an input element for the NP `from`element. This will restrict possible incoming
+connections to the relevant Fybrik module.
```
simanadler, line 81: Isn't this in essence putting the responsibility of the isolation implementation on the data user / FybrikApplication creator?  Why should they have to know anything about network policies?

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+- a blueprint will provide the ingress input data for chain connected modules.
```
simanadler, line 82: Why only chaining modules?

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+
+- deprecate the “workloadSelector” element, but continue to support it for backward compatibility. Process it as an 
+additional “podSelector” entry in the “NetworkPoliciesPeers” array
```
simanadler, line 85: See previous question about workload label

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+
+- terminate the backward compatibility assumption that all Read flow should have at least a single workload label.
```
simanadler, line 87: See previous question about workload label

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+connections to the relevant Fybrik module.
+- a blueprint will provide the ingress input data for chain connected modules.
+
+- deprecate the “workloadSelector” element, but continue to support it for backward compatibility. Process it as an 
+additional “podSelector” entry in the “NetworkPoliciesPeers” array
+
+- terminate the backward compatibility assumption that all Read flow should have at least a single workload label. 
+It was deprecated several Fybrik versions ago.
+
+- change the Flow field to be required instead of optional.
```
simanadler, line 90: I agree that this makes more sense, but this will have backward compatibility implications

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+
+- if the NetworkPolicyPeers array and “workloadSelector” are empty, no Network Policies will be created, which is 
+similar to current implementation.
```
simanadler, line 93: So no isolation in non-read/write flows?

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+
+- extend blueprint’s module install/uninstall process by installation of a relevant instance of network policies.
```
simanadler, line 95: So network policies will be defined by the plotter controller or by the blueprint controller?

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,105 @@
...
+
+  - add the relevant ports to the NetworkPolicyIngressRules.
+  - in future Fybrik releases, if the same module serves several workloads or other modules, we will create an 
+independent NP instance per source. That will help us to manage the NP instances.
```
simanadler, line 99: Can you elaborate?  Who/what will manage the NPs across workloads?  Who will decide when to delete them, etc?

simanadler (commented):
See my questions / comments

docs/IsolationNetworkPolicies.md
roytman: Absolutely, and it is explained in the document

docs/IsolationNetworkPolicies.md
roytman: Container Network Interface - CNI, when you run Kubernetes you can specify which CNI will be used

docs/IsolationNetworkPolicies.md
roytman: I plan to work incrementally, they will be added later in parallel with NP implementation

docs/IsolationNetworkPolicies.md
roytman: For chaining modules we don't need this, blueprint has all required information

docs/IsolationNetworkPolicies.md
roytman: it is not a one label, but a single label-based selector. (Will connect label and based). In theory you can create a fybrik data plane, and allow different workloads with different labels to connect to the data plane. In addition you cannot restrict access based on namespaces or IP addresses.

docs/IsolationNetworkPolicies.md
roytman: If a user doesn't want to set labels on workloads, but prefer to restrict access based on namespaces. 
In our examples we defined that user always set labels to workloads but it is only our assumption.

docs/IsolationNetworkPolicies.md
roytman: They don't have to know about network policies, but they do know where and how they will run workloads. It is similar to the label setting requirement.  All entries are optional, so users do not have to set all options. 
In the current implementation, the `workloadSelector` is much more complicated than just a list of labels.

docs/IsolationNetworkPolicies.md
roytman: I thought about egress, but don't remember why I did not mention it here, will add it.

docs/IsolationNetworkPolicies.md
roytman: What do you mean? It is exactly answering to your question about input to the NP creation.
> This implementation is for isolating the workload from the module with which it interfaces. Doesn't address isolating modules in the data plane I assume

docs/IsolationNetworkPolicies.md
roytman: can you elaborate more ?

docs/IsolationNetworkPolicies.md
roytman: The Flow were implemented an year ago.

docs/IsolationNetworkPolicies.md
roytman: Examples ? Delete module doesn't have a server, no one can connect to it. Data movers are deployed by Fybrik, without input from a FybrikApplication, and blueprint knows all required information about theire sources and destinations.

docs/IsolationNetworkPolicies.md
roytman: the blueprint controller

docs/IsolationNetworkPolicies.md
roytman: if we will support a shared modules, when a single module can be used by different Plotters / blueprints, each blueprint will deploy and manage its independent NP instance. 
For example:
 a fybrik application A created a Plotter/blueprint and they deployed a module M. together with the module the blueprint deployed a NP instance NP1.
 now, another fybrik application B created another Plotter/blueprint but the new blueprint discovered that the required module M was been deployed, so it will reuse the module and instead of deployment a new instance of module, it will create only new instance of NP - NP2. 
 
 I know that this scenario is not supported, i Just mentioned it here to show that the NP mechanism is very flexible, and allows to do it.

roytman: Actually, you gave me another scenario. 
A user created a fybrik application, there is a data plane, but not the user wants to change the isolation, adds or remove workload sources, without any changes in the data plane. We have to check if it possible to update fybrik application spec (the selector only) without redeployment the data plan - only update the NP instance.

docs/IsolationNetworkPolicies.md
roytman: added it

docs/IsolationNetworkPolicies.md
simanadler: Wouldn't the plotter define them, and the blueprint deploy them?

roytman: The plotter defines the data plane so yes it defines who will talk with whom, but it is semantic.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,113 @@
...
+
+- Depends on the isolation method, Fybrik should propagate all relevant information to the deployed modules, and/or 
+correctly configure Kubernetes Network Policies.
+- Module developers should allow to protect the module instances. For example, to allow TLS support
+
+In some complicated cases, IT teams interfere can be required, e.g. to configure external access points.
+
+## Requirements
+
+- Only predefined users/workloads should be able to access the data plane.
```
revit13, line 32: The default behavior is to deny access, right?

docs/IsolationNetworkPolicies.md
revit13: I think that ingress controller is also needed https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/ which is different from CNI

docs/IsolationNetworkPolicies.md
simanadler: Shouldn't it be included here in the challenge part?  Here you only mention isolating the connection between the workload and the module(s) it interacts with

docs/IsolationNetworkPolicies.md
simanadler: If there is another design doc mentioning all the approaches please link to it.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,113 @@
...
+- We do not assume that client workloads will be collocated on the same Kubernetes cluster with the data module,
```
simanadler, line 33: Currently this is the assumption.  Fybrik does not currently work if the workload is in a different cluster than the control plane.  It is something we want to change in the future.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,113 @@
...
+
+
+## Network Policies
+
+Fybrik is a Kubernetes application, therefore, the simplest isolation method can be based on 
+[Kubernetes Network Policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) (NP).
+NP allow to control traffic flow at the IP address or port level (OSI layer 3 or 4). If a user workload or a Fybrik 
+previous chained module collocated with the next module in the same Kubernetes cluster, NPs can help to restrict the Fybrik 
+module ingress access. However, when a workload or a previous chained module runs out of the cluster, the NP based 
+isolation is going to be more complicated. The collocated deployment can be based on Kubernetes labels or namespaces, but
```
simanadler, line 48: co-located (spelling)
I do not believe that Fybrik currently supports namesapace only (without a label) to identify a workload.  Namespace wouldn't be sufficient since there could be multiple workloads in one namespace.

docs/IsolationNetworkPolicies.md
simanadler: See comment above

docs/IsolationNetworkPolicies.md
simanadler: Yes, the flows were implemented but if no flow is indicated in the spec then it is assumed to be "read".  If you make it required there are customers who will have to change their implementations

docs/IsolationNetworkPolicies.md
simanadler: The modules deployed even in ingest/copy/delete flows (which have no workload) should have isolation shouldn't they?  I.e. A copy, quality, delete, etc module shouldn't be able to access data stores, other modules, etc except what is described in the plotter flow.

docs/IsolationNetworkPolicies.md
simanadler: So the plotter defines, and the blueprint deploys.  Right?

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,113 @@
...
+
+Current “FybrikApplicationSpec” has the “[selector](https://fybrik.io/v1.2/reference/crds/#fybrikapplicationspecselector)” 
+element, which is a combination of “clusterName” and “workloadSelector”. Unfortunately, based on this information we can 
+implement only one label-based “podSelector”. Furthermore, for the backward compatibility of the current implementation 
+assumes that a Read scenario should have at least one label, which is not always true. 
+
+Similarly, the "egress.to" element restricts teh outgoing connections. In this way we can restrict destinations where a  
+module can connect.
+
+### Suggetions:
```
simanadler, line 81: Suggestions (spelling)

docs/IsolationNetworkPolicies.md
simanadler: This section is unclear to me.  Can you give some examples?  It's really important that we provide something as simple as possible for the user who is creating the FybrikApplication to indicate the workload.  Let Fybrik discover as much as possible on its own.

simanadler (commented):
See comments

docs/IsolationNetworkPolicies.md
roytman: No, the default behavior is "Allow", if no Network Policies is defined.

docs/IsolationNetworkPolicies.md
roytman: OK, added it

docs/IsolationNetworkPolicies.md
roytman: It is in a box under development now. After that it will be a separate document or extension of this one.

docs/IsolationNetworkPolicies.md
roytman: For example, a user might run a Jupiter notebook on his laptop, outside the Fybrik cluster. Therefore, in order to protect the data plane and allow connectivity only from his laptop, the user can provide the IP address as the input information for the NP creation. 

```yaml
apiVersion: app.fybrik.io/v1beta1
kind: FybrikApplication
metadata:
  name: my-notebook
  labels:
    app: my-notebook
spec:
  selector:
    networkPeers:
      - ipBlock:
            cidr: 172.17.23.5/32
  appInfo:
    intent: Fraud Detection
  data:
    - dataSetID: ${CATALOGED_ASSET}
      requirements:
        interface: 
          protocol: fybrik-arrow-flight
```

docs/IsolationNetworkPolicies.md
roytman: My plan to validate it with our customers, this bullet is not crucial, bit NP.

docs/IsolationNetworkPolicies.md
roytman: Yes, it is described in the Plotter/blueprint flow, user should not and cannot provide this information. The bullets talks about the Workload<->data Plane connection (the first module) only.

docs/IsolationNetworkPolicies.md
roytman: Yes.

docs/IsolationNetworkPolicies.md
roytman: Except checking label existence for deprecated Flow definition, Fybrik do nothing with the labels.  (It's just copied into Plotter)

roytman: I replaced collocated to co-located, but I've seen both forms as acceptable.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+unauthorized data plane access.
+
+## Challenge
+
+When we talk about Fybrik data plane isolation, we are talking about connectivity between a user workload, which is out of
+Fybrik scope, and Fybrik modules, which can be developed by third parties. Therefore, in this document (and in its
+extensions) we provide our recommendations how to protect data plane access and possible solutions. Responsibility of the
+data plane isolation is shared between Fybrik core, Fybrik users, module developers and probably IT teams.
+
+- Fybrik users should provide all relevant information about possible workload locations to Fybrik, correctly
```
shlomitk1, line 23: Fybrik users should provide Fybrik with the information about possible workload locations and configure their workloads. For example, they need to ...

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+
+## Challenge
+
+When we talk about Fybrik data plane isolation, we are talking about connectivity between a user workload, which is out of
+Fybrik scope, and Fybrik modules, which can be developed by third parties. Therefore, in this document (and in its
+extensions) we provide our recommendations how to protect data plane access and possible solutions. Responsibility of the
+data plane isolation is shared between Fybrik core, Fybrik users, module developers and probably IT teams.
+
+- Fybrik users should provide all relevant information about possible workload locations to Fybrik, correctly 
+configure their workloads. For example, to set labels on the workload and inform Fybrik about the labels and namespace
```
shlomitk1, line 24: set labels on the workload pods?

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+## Challenge
+
+When we talk about Fybrik data plane isolation, we are talking about connectivity between a user workload, which is out of
+Fybrik scope, and Fybrik modules, which can be developed by third parties. Therefore, in this document (and in its
+extensions) we provide our recommendations how to protect data plane access and possible solutions. Responsibility of the
+data plane isolation is shared between Fybrik core, Fybrik users, module developers and probably IT teams.
+
+- Fybrik users should provide all relevant information about possible workload locations to Fybrik, correctly 
+configure their workloads. For example, to set labels on the workload and inform Fybrik about the labels and namespace 
+where the workload runs.
```
shlomitk1, line 25: runs -> is deployed

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+
+- Depends on the isolation method, Fybrik should propagate all relevant information to the deployed modules, and/or
```
shlomitk1, line 27: Depends -> Depending

shlomitk1: all relevant -> all the relevant

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+correctly configure a chosen protection mechanism, e.g. Network Policies..
+
+- Module developers should allow to protect the module instances. For example, to allow TLS support
```
shlomitk1, line 30: should allow protection of

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+- Fybrik users should provide all relevant information about possible workload locations to Fybrik, correctly 
+configure their workloads. For example, to set labels on the workload and inform Fybrik about the labels and namespace 
+where the workload runs.
+
+- Depends on the isolation method, Fybrik should propagate all relevant information to the deployed modules, and/or
+correctly configure a chosen protection mechanism, e.g. Network Policies..
+
+- Module developers should allow to protect the module instances. For example, to allow TLS support 
+ 
+- In some complicated cases, IT teams interfere can be required, e.g. to configure external access points, firewalls and so on.
```
shlomitk1, line 32: interference

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+
+## Requirements
+
+- Only predefined users/workloads should be able to access the data plane.
+- If a data plane consist of a module chain we should prevent authorized access to the intermediate modules too.
```
shlomitk1, line 37: If a chain of modules is deployed in the data plane, we should ...
authorized -> unauthorized

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+- Modules should not be able to "illegally" communicate with modules outside of the defined data plane(s) and should 
+connect to only predefined data sets.
```
shlomitk1, line 39: I don't see how network policies prevent the module from connecting a different dataset. Is perhaps giving them credentials to specific datasets only enough?

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+- Currently, Fybrik assumes co-location of client workloads with data plane entry point modules. Future implementations
```
shlomitk1, line 40: co-location in terms of cluster. Different namespaces are ok.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+might support deployment workloads on different clusters or run them out of Kubernetes as a standalone applications.
```
shlomitk1, line 41: deployment of workloads

shlomitk1, line 41: as a standalone -> as standalone

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+  - We check the co-location of a Fybrik module and a relevant workload in the same cluster as a specific use case.
```
shlomitk1, line 42: How do we check the co-location? We may enforce it using IT config policies.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+- We should allow workload client connections isolation based on IP addresses. It can be useful if a workload is not
```
shlomitk1, line 43: workload client connections isolation -> workload-client isolation

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+
+- Only predefined users/workloads should be able to access the data plane.
+- If a data plane consist of a module chain we should prevent authorized access to the intermediate modules too.
+- Modules should not be able to "illegally" communicate with modules outside of the defined data plane(s) and should 
+connect to only predefined data sets.
+- Currently, Fybrik assumes co-location of client workloads with data plane entry point modules. Future implementations
+might support deployment workloads on different clusters or run them out of Kubernetes as a standalone applications.
+  - We check the co-location of a Fybrik module and a relevant workload in the same cluster as a specific use case.
+- We should allow workload client connections isolation based on IP addresses. It can be useful if a workload is not 
+co-locate with a Fybrik data plane.
```
shlomitk1, line 44: I don't understand this sentence. By default, Fybrik generates the data plane, and the workload is deployed somewhere else.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+- We should allow workload client connections isolation based on IP addresses. It can be useful if a workload is not 
+co-locate with a Fybrik data plane.  
+
+## Network Policies
+
+Fybrik is a Kubernetes application, therefore, the simplest isolation method can be based on 
+[Kubernetes Network Policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) (NP).
+NP allow to control traffic flow at the IP address or port level (OSI layer 3 or 4). NP can restrict incoming (ingress) 
+and/or outgoing (egress) communications. 
+The restrictions are based on combinations of Pod labels, Namespaces labels and IP blocks.
```
shlomitk1, line 52: Namespaces -> Namespace

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Advantages:
+- does not require any changes in modules implementations, can be implemented in the Fybrik control plane only.
+- can be combined with other isolation methods
+
+### Disadvantages:
+- a single Kubernetes cluster oriented, cross clusters restrictions can be based on IP blocks only, which significantly 
+complicates the configuration. Therefore, later we will provide other isolation methods as an extension to NP.
```
shlomitk1, line 60: Will we be able to support module chaining when modules are deployed on different clusters? Or does this mean that isolation will be limited in this case?

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+- NP are implemented by the network plugin. Creating a NetworkPolicy resource without a controller - a Container Network 
+Interface (CNI) that implements it will have no effect. Kind deployments which by default use "kindnetd" do not
+support NP. Support NP requires installation another CNI, e.g. [Calico](https://github.com/projectcalico/calico). Here is the
```
shlomitk1, line 63: NP support
installation of

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+allows to restrict incoming network traffic based on `podSelectors` (labels based), and/or 
+`namespaceSelectors` (labels based) and/or `ipBlocks`. The last one permits incoming network traffic from predefined IP 
+address blocks. 
+The `to` element of `egress` restricts outgoing connections and has similar structure. 
+
+Current `FybrikApplicationSpec` has the “[selector](https://fybrik.io/v1.2/reference/crds/#fybrikapplicationspecselector)” 
+element, which is a combination of `clusterName` and `workloadSelector`. Unfortunately, based on this information we can 
+implement only one label-based `podSelector`, which will allow incoming connections from pods in the namespace where the NP 
+instance is deployed and have the defined label. In order to restrict incoming connections to Fybrik modules, the NP
+instances should be deployed in teh same namespace where the modules are running. However, Fybrik separates user's
```
shlomitk1, line 80: *the

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+element, which is a combination of `clusterName` and `workloadSelector`. Unfortunately, based on this information we can 
+implement only one label-based `podSelector`, which will allow incoming connections from pods in the namespace where the NP 
+instance is deployed and have the defined label. In order to restrict incoming connections to Fybrik modules, the NP
+instances should be deployed in teh same namespace where the modules are running. However, Fybrik separates user's 
+objects, e.g. workloads and Fybrik objects, e.g. modules. Therefore, the existing mechanism should be extended.
+
+### Suggestions:
+
+- Extend the `FybrikApplicationSpec.selector` element with an array of `WorkloadLocations`, this array will be an input 
+element for the Network Policy `from`element of the data plane entry module.
```
shlomitk1, line 86: Why only from? Sometimes 'to' as well, correct? Depending on the flow - read vs. write

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+- The `WorkloadLocation` wil have the following structure: 
+
+```go
+type WorkloadLocation struct {
+    // This is a workload's pod label selector which selects user workload pods. This field follows standard label
+    // selector semantics; if present but empty, it selects all pods.
+    //
+    // If NamespaceSelector is also set, then the WorkloadLocation as a whole selects
+    // the workload pods matching WorkloadPodSelector in the namespaces selected by NamespaceSelector.
+    // Otherwise, it selects the pods matching WorkloadPodSelector in the Fybrik data plane entry point namespace.
```
shlomitk1, line 96: According to this semantics, I would assume that if NamespaceSelector is not set, all namespaces are selected.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+### Suggestions:
+
+- Extend the `FybrikApplicationSpec.selector` element with an array of `WorkloadLocations`, this array will be an input 
+element for the Network Policy `from`element of the data plane entry module. 
+- The `WorkloadLocation` wil have the following structure: 
+
+```go
+type WorkloadLocation struct {
+    // This is a workload's pod label selector which selects user workload pods. This field follows standard label
+    // selector semantics; if present but empty, it selects all pods.
```
shlomitk1, line 92: What happens if not present?

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+    // +optional
+    NamespaceSelector *metav1.LabelSelector `json:"namespaceSelector,omitempty" protobuf:"bytes,2,opt,name=namespaceSelector"`
+
+    // IPBlock defines policy on a particular IPBlock. If this field is set then neither of the other fields can be.
+	// the structure of the IPBlock is defined at https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#ipblock-v1-networking-k8s-io
+    // +optional
+    IPBlock *IPBlock `json:"ipBlock,omitempty" protobuf:"bytes,3,rep,name=ipBlock"`
+} 
+```
+- A Plotter and relevant blueprints will provide the ingress/egress data for chain connected modules.
```
shlomitk1, line 118: What does it mean? What extra information is required in the plotter/blueprints?

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+`FybrikApplication` is deployed. For example, if a `FybrikApplication` is deployed in the `fybrik-notebook-sample` 
+namespace and a `workloadSelector` defines that a workload should be labeled as `app: my-notebook`, this will be equivalent to  
+```yaml
+workloadLocation:
+  workloadPodSelector:
+    matchLabels:
+      app: my-notebook
+  namespaceSelector:
+    matchLabels:
+      kubernetes.io/metadata.name: fybrik-notebook-sample
```
shlomitk1, line 131: I suggest to deprecate `workloadSelector` without converting it to this structure, as it may break backwards compatibility - suddenly we start creating network policies for existing applications...

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+          matchLabels:
+            kubernetes.io/metadata.name: fybrik-notebook-sample 
+...
+```
+- If a user wants to run his workloads in different location, he can define several `WorkloadLocation`, they all will be
+processed as logical OR, similar to NP.
+- Terminate the backward compatibility assumption that all `Read` flows should have at least a single workload label. 
+It was deprecated several Fybrik versions ago.
+
+_Note_:  should be validated with our customers.
```
shlomitk1, line 173: Alternative suggestion: if a flow is empty, treat it as Read without referring to workload label selector.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+
+- Change the Flow field to be required instead of optional.
+
+_Note_:  should be validated with our customers.
+
+- If a `WorkloadLocation` array and a `workloadSelector` are empty, no Network Policies will be created, which is 
+similar to current implementation. Probably other isolation methods are used.
+
+- Allow to a user by modifying the `FybrikApplication.spec.selector` in the instance of its `FybrikApplication` to modify
```
shlomitk1, line 182: By modifying the `FybrikApplication.spec.selector`, allow the user to modify the NP...

docs/IsolationNetworkPolicies.md
roytman: done

docs/IsolationNetworkPolicies.md
roytman: added pods

docs/IsolationNetworkPolicies.md
roytman: ok

docs/IsolationNetworkPolicies.md
roytman: done

docs/IsolationNetworkPolicies.md
roytman: Changed to intervention

docs/IsolationNetworkPolicies.md
roytman: of course, thanks

docs/IsolationNetworkPolicies.md
roytman: We can prevent connectivity to a different destination. (Egress restriction)

docs/IsolationNetworkPolicies.md
roytman: yeap

roytman: added "in the same cluster"

docs/IsolationNetworkPolicies.md
roytman: fixed

roytman: fixed

docs/IsolationNetworkPolicies.md
roytman: yes, plus it can be seen from the input provided by users. in the FybrikApplication.spec.selector

docs/IsolationNetworkPolicies.md
roytman: ok

docs/IsolationNetworkPolicies.md
shlomitk1: co-locate -> co-located
add "in the same cluster"

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+## Network Policies
+
+Fybrik is a Kubernetes application, therefore, the simplest isolation method can be based on 
+[Kubernetes Network Policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) (NP).
+NP allow to control traffic flow at the IP address or port level (OSI layer 3 or 4). NP can restrict incoming (ingress) 
+and/or outgoing (egress) communications. 
+The restrictions are based on combinations of Pod labels, Namespaces labels and IP blocks.
+
+### Advantages:
+- does not require any changes in modules implementations, can be implemented in the Fybrik control plane only.
```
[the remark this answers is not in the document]
roytman: fixed

docs/IsolationNetworkPolicies.md
roytman: I am not sure that I understand your question. We are going to present cross cluster connectivity, which can be done with different methods. In some cases, NP based isolation can be limited to IP blocks, which is not very convenient.

docs/IsolationNetworkPolicies.md
roytman: Changed to "Support of NP requirements"...
added of

docs/IsolationNetworkPolicies.md
roytman: fixed

docs/IsolationNetworkPolicies.md
roytman: no, in the FybrikApplication  a user hints where his Workload will run, and it is the incoming flow to a Fybrik modules.
For `to` (egress) configuration, the data will be provided by Plotter (a modules chain) and probably the data set info.

docs/IsolationNetworkPolicies.md
roytman: see below

docs/IsolationNetworkPolicies.md
roytman: I did not say extra information, a Plotter/blueprints deploy modules, and they know which modules should connect to another module

docs/IsolationNetworkPolicies.md
roytman: Agree

docs/IsolationNetworkPolicies.md
roytman: OK

docs/IsolationNetworkPolicies.md
shlomitk1: OK. As for the data set info, I don't see how you get this information. At most, you can get a host url - not the exact IP.

docs/IsolationNetworkPolicies.md
shlomitk1: Why will provide? They provide already.

roytman: it's semantic. They **will** provide to the NP creation mechanism.

docs/IsolationNetworkPolicies.md
roytman: OK

docs/IsolationNetworkPolicies.md
roytman: I can translate DNS host name to IP

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+[NetworkPolicySpec](https://kubernetes.io/docs/reference/kubernetes-api/policy-resources/network-policy-v1/#NetworkPolicySpec) 
+allows to restrict incoming network traffic based on `podSelectors` (labels based), and/or 
+`namespaceSelectors` (labels based) and/or `ipBlocks`. The last one permits incoming network traffic from predefined IP 
+address blocks. 
+The `to` element of `egress` restricts outgoing connections and has similar structure. 
+
+Current `FybrikApplicationSpec` has the “[selector](https://fybrik.io/v1.2/reference/crds/#fybrikapplicationspecselector)” 
+element, which is a combination of `clusterName` and `workloadSelector`. Unfortunately, based on this information we can 
+implement only one label-based `podSelector`, which will allow incoming connections from pods in the namespace where the NP 
+instance is deployed and have the defined label. In order to restrict incoming connections to Fybrik modules, the NP
```
simanadler, line 79: There is no mention of who deploys the NP nor how it is decided where to deploy them, so this problem is unclear to me.

docs/IsolationNetworkPolicies.md
simanadler: I don't understand the need for a list of WorkloadLocations.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+objects, e.g. workloads and Fybrik objects, e.g. modules. Therefore, the existing mechanism should be extended.
+
+### Suggestions:
+
+- Extend the `FybrikApplicationSpec.selector` element with an array of `WorkloadLocations`, this array will be an input 
+element for the Network Policy `from`element of the data plane entry module. 
+- The `WorkloadLocation` wil have the following structure: 
+
+```go
+type WorkloadLocation struct {
```
simanadler, line 90: I would suggest calling it Workload - _without the location_

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+	// IF you want to select a specific namespace based on its name, it can be done by selecting an immutable label 
+	// `kubernetes.io/metadata.name` which is automatically set by the Kubernetes control plane. The value of the label 
+	// is the namespace name.
+    // +optional
+    NamespaceSelector *metav1.LabelSelector `json:"namespaceSelector,omitempty" protobuf:"bytes,2,opt,name=namespaceSelector"`
+
+    // IPBlock defines policy on a particular IPBlock. If this field is set then neither of the other fields can be.
+	// the structure of the IPBlock is defined at https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#ipblock-v1-networking-k8s-io
+    // +optional
+    IPBlock *IPBlock `json:"ipBlock,omitempty" protobuf:"bytes,3,rep,name=ipBlock"`
```
simanadler, line 115: See previous comments about IP based isolation.  We either need to be clear about when to use it or leave it out for now, and wait until we talk about addressing the relevant problem.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+} 
+```
+- A Plotter and relevant blueprints will provide the ingress/egress data for chain connected modules.
+
+- Deprecate the `workloadSelector` element, but continue to support it for backward compatibility. Process it as an
```
simanadler, line 120: Not sure I'm convinced that we need to make changes for the initial version of isolation.  Current structure supports workload label and namespace doesn't it?

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+spec:
+  selector:
+    workloadLocations:
+      - workloadPodSelector:
+          matchLabels:
+            app: my-notebook
+        namespaceSelector:
+          matchLabels:
+            kubernetes.io/metadata.name: fybrik-notebook-sample 
+...
```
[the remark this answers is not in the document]
simanadler: What have we gained by moving to this new structure that we can't support today?  (Putting aside IP based NP)

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+```
+- If a user wants to run his workloads in different location, he can define several `WorkloadLocation`, they all will be
```
simanadler, line 168: Why does he need to define different locations?  If he provides a label won't a k8s query get him all of the workloads with that label?  For multi-cluster, isn't it the role of the multi-cluster control plane to provide all of them across the workloads?  Not sure why we are asking the FybrikApplication user to provide this info, rather than relying on k8s to do it.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+processed as logical OR, similar to NP.
+- Terminate the backward compatibility assumption that all `Read` flows should have at least a single workload label. 
+It was deprecated several Fybrik versions ago.
+
+_Note_:  should be validated with our customers.
+
+- Change the Flow field to be required instead of optional.
```
simanadler, line 175: What does this have to do with isolation?  Should be a separate issue in my opinion

docs/IsolationNetworkPolicies.md
simanadler, line 182: I don't understand this statement.  The user shouldn't know anything about the NPs.  He provides the info about his workload and the data he wants to access.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+- Allow to a user by modifying the `FybrikApplication.spec.selector` in the instance of its `FybrikApplication` to modify 
+the NP without modification/redeployment of the data plane.
+
+_Note_: we have to check if it is possible.    
+
+- Extend blueprint’s module install/uninstall process by installation/removing of relevant NP instances.
+
+- Fybrik will not support different network restrictions to different module ports. (NP allow it)
+
+- If a user expects to run workload outside Fybrik clusters (it requires additional networking configuration) and the user
```
simanadler, line 191: Fybrik does not currently support this.  Do we really want to add this, especially with the user providing IP addresses?  Seems like potential to be a major security issue.  Workloads running inside managed clusters are under security controls of the enterprise.  Those running from outside are less so.

simanadler (commented):
See comments.  I thought the intent was to add isolation to the current Fybrik capabilities. Assuming that, I'm not clear on what limitations the current FybrikApplication.spec presents and why it needs to be extended.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+### Disadvantages:
+- a single Kubernetes cluster oriented, cross clusters restrictions can be based on IP blocks only, which significantly 
+complicates the configuration. Therefore, later we will provide other isolation methods as an extension to NP.
+- NP are implemented by the network plugin. Creating a NetworkPolicy resource without a controller - a Container Network 
+Interface (CNI) that implements it will have no effect. Kind deployments which by default use "kindnetd" do not
+support NP. Support NP requires installation another CNI, e.g. [Calico](https://github.com/projectcalico/calico). Here is the
+[instructions](https://alexbrand.dev/post/creating-a-kind-cluster-with-calico-networking/). On the other hand, all K8s 
+clusters in production deployments use CNIs which support NP.
+
+## Implementation
```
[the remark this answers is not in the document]
simanadler: What are the changes required to the Plotter and Blueprint?  What is the algorithm to be used to calculate the NPs required?

simanadler (commented):
Added 1 more comment

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+status, it is completely unprotected, and as result everyone can connect to the data plane. If a Fybrik data plane 
+consist of a module chain, the same vulnerability exists for the chained modules.  
+In `Read` flow cases it leads to governance/security violations, `Write` and `Delete` flows can create severe accidents, 
+such as unauthorized data changes or deletions. Given that we have to provide a mechanism, which will prevent 
+unauthorized data plane access.
+
+## Challenge
+
+When we talk about Fybrik data plane isolation, we are talking about connectivity between a user workload, which is out of
+Fybrik scope, and Fybrik modules, which can be developed by third parties. Therefore, in this document (and in its
```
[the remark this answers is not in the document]
roytman: OK, done

docs/IsolationNetworkPolicies.md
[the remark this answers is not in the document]
roytman: @shlomitk1 pointed this out, fixed.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+
+Fybrik is a Kubernetes application, therefore, the simplest isolation method can be based on 
+[Kubernetes Network Policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) (NP).
+NP allow to control traffic flow at the IP address or port level (OSI layer 3 or 4). NP can restrict incoming (ingress) 
+and/or outgoing (egress) communications. 
+The restrictions are based on combinations of Pod labels, Namespaces labels and IP blocks.
+
+### Advantages:
+- does not require any changes in modules implementations, can be implemented in the Fybrik control plane only.
+- can be combined with other isolation methods
```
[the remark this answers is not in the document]
roytman: OK

docs/IsolationNetworkPolicies.md
roytman: >Why does he need to define different locations? 

A user(s) can deploy workloads in namespaces. And or out of K8s cluster.

> If he provides a label won't a k8s query get him all of the workloads with that label?

By default, K8s returns pods from the current namespace. NP's `PodSelector` works for the NP's instance namespace.

>For multi-cluster, isn't it the role of the multi-cluster control plane to provide all of them across the workloads? 

depends ...

> Not sure why we are asking the FybrikApplication user to provide this info, rather than relying on k8s to do it.

But even today we require from the user to define labels, what is the difference.

docs/IsolationNetworkPolicies.md
roytman: Maybe, will remove it

docs/IsolationNetworkPolicies.md
roytman: User or an IT person, or a future REST/UI interface

roytman:
> What are the changes required to the Plotter and Blueprint? What is the algorithm to be used to calculate the NPs required?

In their algorithm - nothing. They do not calculate the NP requirements, rather creates NP instances, based on their information.

docs/IsolationNetworkPolicies.md
roytman: done

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Disadvantages:
+- a single Kubernetes cluster oriented, cross clusters restrictions can be based on IP blocks only, which significantly 
+complicates the configuration. Therefore, later we will provide other isolation methods as an extension to NP.
+- NP are implemented by the network plugin. Creating a NetworkPolicy resource without a controller - a Container Network 
+Interface (CNI) that implements it will have no effect. Kind deployments which by default use "kindnetd" do not
```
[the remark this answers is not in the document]
roytman: Network Policies is an optional feature of CNI. If you try to use NP on a Kubernetes cluster that runs with a CNI that doesn't support NP, you will not get any errors, but Network Policies restrictions will not work, will have no effect"

By default Kind uses a CNI that doesn't support NP. 
All production Kubernets deployments usually support Network Policies.

docs/IsolationNetworkPolicies.md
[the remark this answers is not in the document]
roytman: I don't expect any changes in the Plotter and Blueprint algorithm, Plotter defines the data plane as today, and provides this information to blueprints. 
The blueprints in addition to deployment modules will deploy NP instances based on the information from the Plotter.

docs/IsolationNetworkPolicies.md
roytman: See Line: 184 "Extend blueprint’s module install/uninstall process by installation/removing of relevant NP instances."

This problem is a consequent of the NP [spec](https://kubernetes.io/docs/concepts/services-networking/network-policies/), you are welcome to take a look at it, plus it was presented on one of our group meetings.

docs/IsolationNetworkPolicies.md
roytman: Why ?

docs/IsolationNetworkPolicies.md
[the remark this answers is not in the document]
roytman: correct.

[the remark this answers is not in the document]
roytman: L: 176, "If a `WorkloadLocation` array and a `workloadSelector` are empty, no Network Policies will be created, which is similar to current implementation. Probably other isolation methods are used."

`workloadSelector` can be discussed separately.

docs/IsolationNetworkPolicies.md
roytman: no, it doesn't support namespaces

docs/IsolationNetworkPolicies.md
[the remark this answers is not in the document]
roytman: first at all namespaces

docs/IsolationNetworkPolicies.md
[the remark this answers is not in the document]
shlomitk1: This is what I understood: no changes to Blueprint/Plotter CRDs are required. BlueprintController will create NP as well, based on the information from the Blueprint.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,207 @@
...
+- We should allow workload-client isolation based on IP addresses. It can be useful if a workload is not 
+co-located with a Fybrik data plane in the same cluster  
+
+## Network Policies
+
+Fybrik is a Kubernetes application, therefore, the simplest isolation method can be based on 
+[Kubernetes Network Policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) (NP).
+NP allow to control traffic flow at the IP address or port level (OSI layer 3 or 4). NP can restrict incoming (ingress) 
+and/or outgoing (egress) communications. 
+The restrictions are based on combinations of Pods and Namespaces labels and IP blocks.
```
shlomitk1, line 53: combinations of Pod labels, Namespace labels and IP blocks

docs/FybrikNetworkTopologiesAndRequirements.md
```diff
@@ -0,0 +1,78 @@
...
+
+User applications (workloads) achieve the Fybrik data plane via its entry point. Depends on the location of the
+application we can separate 3 different connection types:
+- `1a` - the application and the data plane entry module are collocated on the same Kubernetes cluster.
+- `1b` - the application is running on a Kubernetes cluster, but Fybrik cannot control the cluster.
+- `1c` - the application is not deployed on a kubernetes cluster.
+
+In all these connections, the Fybrik entry point is a server, and user applications are clients. Connections `1b` and
+`1c` require external access to the Kubernetes pods, a public entry point or other mechanisms.  
+We will start on isolation of the `1a` connections. Isolation and protection of other connections types will be investigated later.
```
shlomitk1, line 22: start with

docs/FybrikNetworkTopologiesAndRequirements.md
```diff
@@ -0,0 +1,78 @@
...
+
+In the connections `3`, `4`, and `5` Fybrik modules are clients that connects to services or data sources. From the
```
shlomitk1, line 24: clients that connect

docs/FybrikNetworkTopologiesAndRequirements.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Responsibilities
+
+Responsibility of the data plane isolation is shared between Fybrik core, Fybrik users, module developers and probably IT teams.
+
+- Fybrik users should inform Fybrik about possible workload locations and configure their workloads accordingly.
+  For example, to set labels on the workload pods and inform Fybrik about the labels and the namespace where the 
+  workload is deployed.
+
+- Modules might support certain protection mechanisms, such as TLS/mTLS, AIM or JWT tokens, and others. These modules
```
shlomitk1, line 61: IAM

docs/FybrikNetworkTopologiesAndRequirements.md
```diff
@@ -0,0 +1,78 @@
...
+## Responsibilities
+
+Responsibility of the data plane isolation is shared between Fybrik core, Fybrik users, module developers and probably IT teams.
+
+- Fybrik users should inform Fybrik about possible workload locations and configure their workloads accordingly.
+  For example, to set labels on the workload pods and inform Fybrik about the labels and the namespace where the 
+  workload is deployed.
+
+- Modules might support certain protection mechanisms, such as TLS/mTLS, AIM or JWT tokens, and others. These modules 
+  should declare these optional functionality in their manifests. Fybrik users may require that the entire data plain will
```
shlomitk1, line 62: should declare what mechanisms they support.

docs/FybrikNetworkTopologiesAndRequirements.md
```diff
@@ -0,0 +1,78 @@
...
+  be build with these features. Fybrik core should support it.
```
shlomitk1, line 63: be built

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,121 @@
...
+	Namespaces []string `json:"namespaces"`
+
+	// IPBlocks define policy on particular IPBlocks.
+	// the structure of the IPBlock is defined at https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#ipblock-v1-networking-k8s-io
+	// +optional
+	IPBlocks []*IPBlock `json:"ipBlocks,omitempty"`
+}
+```
+The mapping the `Selector` elements into the `from` NP ingress definitions will be done according the following algorithm:
+- if `Namespaces` is not empty, for each entry in the array will be created a
```
shlomitk1, line 86: If Namespaces is not empty but WorkloadSelector is empty, then it means any pods in the namespaces, correct?

docs/FybrikNetworkTopologiesAndRequirements.md
roytman: fixed

docs/FybrikNetworkTopologiesAndRequirements.md
roytman: thanks, fixed

docs/FybrikNetworkTopologiesAndRequirements.md
roytman: rephrased, thanks

docs/FybrikNetworkTopologiesAndRequirements.md
roytman: fixed

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,121 @@
...
+On the other hand, all K8s clusters in production deployments use CNIs which support NP.
+
+## Implementation
+
+The `from`/`to` element of `ingress`/`egress`  
+[NetworkPolicySpec](https://kubernetes.io/docs/reference/kubernetes-api/policy-resources/network-policy-v1/#NetworkPolicySpec) 
+allows to restrict incoming/outgoing network traffic based on `podSelectors` (labels based), and/or 
+`namespaceSelectors` (labels based) and/or `ipBlocks`. The last parameter can be used to restrict cross cluster connections 
+
+We suggest, after deployment a data plane module, to create an instance of a network policy that will restrict the
```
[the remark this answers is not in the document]
roytman: done

docs/IsolationNetworkPolicies.md
roytman: correct, added the explanation.

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Implementation
+
+The `from`/`to` element of `ingress`/`egress`  
+[NetworkPolicySpec](https://kubernetes.io/docs/reference/kubernetes-api/policy-resources/network-policy-v1/#NetworkPolicySpec) 
+allows to restrict incoming/outgoing network traffic based on `podSelectors` (labels based), and/or 
+`namespaceSelectors` (labels based) and/or `ipBlocks`. The last parameter can be used to restrict cross cluster connections 
+
+We suggest, after deployment a data plane module, to create an instance of a network policy that will restrict the 
+module connectivity.
```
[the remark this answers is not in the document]
roytman: > Who is responsible for the network policy creation - Blueprint controller?

Yes, there is an explanation about it:
"BlueprintController install and uninstall Fybrik modules. We suggest extending this functionality by creating an instance of NP for each installing module, and delete it when the module is uninstalled." 

> The module status will be marked as ready only after the network policy is deployed, correct?
Actually, we can deploy NP **before** module deployment, so the above statement will be correct. (updated the explanation)

> Does the deployment take effect immediately? 
Yes, I've never seen any comments about NP delay, however they do not influence on the existing connections, so deployment NP before the relevant module is better. 

> Should some resource status be checked?
Usually we don't have to. We can validate that NP ingress/egress definitions match the incoming/outgoing trafic. 
If somebody insist he can run netcat([nc](https://linux.die.net/man/1/nc)) on a pod. It can be part of our integration tests. 
NetworkPolicies are an optional CNI feature, and if a CNI doesn't support NP (e.g. kind by default), no error will be sent, but all connections will be allowed. 

> What happens in case of a failure (on the network plugin side) - will the error be propagated to the user?
I'm not sure that I understand your question, do you mean error in the CNI implementation? It depends on CNI, some of them can create Kubernetes event, most of them write an error/warning into **their** logs.

@zivnevo told me that NP-gurard could help to debug a system.

docs/FybrikNetworkTopologiesAndRequirements.md
```diff
@@ -0,0 +1,78 @@
...
+We have started from the data plane because it is dynamically created and might include third-part developed components.
+Furthermore, user data is accessed, modified and moved by the data plane only.
+Therefore, networking isolation and protection of data plane is very important and challenged task, and we start from it.
+
+### Types of Fybrik Network Connections
+Fybrik data plane can be spread across several Kubernetes clusters and theoretically include any number of modules. 
+The following picture demonstrates different Fybrik data plane connection types.
+![](./images/ConnectionTypes.png)
+
+User applications (workloads) achieve the Fybrik data plane via its entry point. Depends on the location of the
```
shlomitk1, line 14: Depending

shlomitk1: Better to rephrase: "There are 3 different connection types, based on the location of the application:

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,124 @@
...
+
+In order to create ingress policies for modules, the blueprint that creates NPs should know potential clients, therefore
+PlotterController who has the entire picture, should provide this information. 
+
+NetworkPolicies select pods according to labels. In order to separate policies for 2 different modules, each module should
+have a unique label, in additional to the common, fybrik application defined labels.
+
+### BlueprintController extensions
+
+BlueprintController install and uninstall Fybrik modules. We suggest extending this functionality by creating an instance
```
shlomitk1, line 114: installs and uninstalls

docs/FybrikNetworkTopologiesAndRequirements.md
roytman: OK

docs/IsolationNetworkPolicies.md
roytman: fixed

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,124 @@
...
+
+BlueprintController installs and uninstalls Fybrik modules. We suggest extending this functionality by creating an instance
+of NP for each installing module, and delete it when the module is uninstalled.
+
+For the data plane entry module, the ingress element will be created based on information from a 
+relevant `FybrikApplication`. For other modules this information will be provided by the `PlotterController`
+Fybrik will not support different network restrictions to different module ports. (NP allow it)
+
+The egress element will be a combination of destinations, such as, the next module (`2a`), data source (`5`), Fybrik (`3`) 
+or module required (`4`) services. The connection types are taken from the [network topologies](FybrikNetworkTopologiesAndRequirements.md)
```
shlomitk1, line 122: How are these destinations formed? Specifically:
2a - Fybrik knows where the module is deployed (namespace) but not the labels of the deployed pods which is internal to the module chart implementation. Unless I am missing some assumptions on module implementation.
3 - any change to the existing deployment of Vault and the relevant policies?
5 - Fybrik is not aware of the data store destination. The connection details are passed as-is to the access module. How do you see this policy implemented?

docs/IsolationNetworkPolicies.md
```diff
@@ -0,0 +1,124 @@
...
+We suggest adding an `externalServices` array into the `FybrikModule.Spec`. Each entry will be a DNS name or IP address of 
+the external service. During `egress` NetworkPolicyPeer creations, each service name will be translated to a single IPBlock entry.
+
+### PlotterController extensions
+
+In order to create ingress policies for modules, the blueprint that creates NPs should know potential clients, therefore
+PlotterController who has the entire picture, should provide this information. 
+
+NetworkPolicies select pods according to labels. In order to separate policies for 2 different modules, each module should
+have a unique label, in additional to the common, fybrik application defined labels.
```
shlomitk1, line 110: Fybrik passes a set of labels inside Helm values but does not enforce these labels on the deployed resources. No one guarantees that a 3rd party module will propagate the labels to its pods.

docs/IsolationNetworkPolicies.md
roytman: 2a, it is a regular chaining convection, fybrik adds some of the labels, and actually for NP I need only a single one `app.kubernetes.io/instance`, I can demonstrate how it works
3- I don't see any changes
5 - I plan to get the destination details from the connection details

PR #1955, opened 2023-01-29
Fixing illegal asset name in the Katalog

tomersolomon1:
Fixes https://github.com/fybrik/fybrik/issues/1789

Changed the function `ShortenedName` to validate now the name is proper DNS subdomain.

Signed-off-by: Tomer Solomon <tomer.solomon1@gmail.com>


connectors/katalog/pkg/connector/connector_test.go
```diff
@@ -119,86 +118,104 @@ func TestCreateAsset(t *testing.T) {
...
-				},
-			},
+	var tests = []struct {
+		csvFormat       taxonomy.DataFormat
+		sourceAssetName string
+		destAssetName   string
+		destCatalogID   string
+	}{
+		{
+			csvFormat:       "csv",
```
shlomitk1, line 128: Can csvFormat get a different value?

connectors/katalog/pkg/connector/connector_test.go
```diff
@@ -119,86 +118,104 @@ func TestCreateAsset(t *testing.T) {
...
+			destAssetName:   "new-paysim-csv",
+			destCatalogID:   "fybrik-system",
 		},
-		Details: datacatalog.ResourceDetails{
-			Connection: s3Connection,
-			DataFormat: csvFormat,
+		{
+			csvFormat:       "csv",
+			sourceAssetName: "a%ba/df12-csv",
+			destAssetName:   "a%ba/d//f=12-csv",
```
shlomitk1, line 136: This one should fail. You can't create Asset with name "d//f=12-csv"

connectors/katalog/pkg/connector/connector.go
```diff
@@ -121,6 +121,10 @@ func (r *Handler) createAsset(c *gin.Context) {
 		assetPrefix = request.DestinationAssetID + "-"
 	}
 
+	// Split the string after the first "/", if exists, and take what's after the "/"
+	slices := strings.SplitAfterN(assetPrefix, "/", 2)
+	assetPrefix = slices[len(slices)-1]
+
```
shlomitk1, line 127: I don't see how this turns an illegal asset name into a legal one for a general case.
`a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')" for field "metadata.generateName"`

tomersolomon1: The test doesn't fail b/c internally it makes it a valid name.

connectors/katalog/pkg/connector/connector_test.go
tomersolomon1: The test doesn't fail b/c internally it makes it a valid name.

connectors/katalog/pkg/connector/connector_test.go
tomersolomon1: See now.

pkg/utils/utils.go
```diff
@@ -77,6 +80,13 @@ func Hash(value string, hashLength int) string {
 // This function shortens a name to the maximum length given and uses rest of the string that is too long
 // as hash that gets added to the valid name.
 func ShortenedName(name string, maxLength, hashLength int) string {
+	actualBound := int(math.Min(float64(maxLength), float64(len(name))))
+	// First shorten so if there is a problem it's due to ilegal chars, rahter its legnth
```
shlomitk1, line 84: ilegal -> illegal (here and in other places)

pkg/utils/utils.go
```diff
@@ -77,6 +80,13 @@ func Hash(value string, hashLength int) string {
...
+	if ilegal := validation.IsDNS1123Subdomain(name[:actualBound]); ilegal != nil {
```
shlomitk1, line 85: Suppose you have an illegal name "xxx/yyy", maxLength=5, hashLength=2. In this case actualBound = 5, so you check "xxx/y" and end up in hashing all of it. In the original code, only "xxx" is taken as prefix and the rest is used as hashed suffix of length 2. So, you over-check.

connectors/katalog/pkg/connector/connector_test.go
```diff
@@ -114,91 +115,109 @@ func TestCreateAsset(t *testing.T) {
...
-				},
-			},
+	var tests = []struct {
+		csvFormat       taxonomy.DataFormat
+		sourceAssetName string
+		destAssetName   string
+		destCatalogID   string
+	}{
+		{
+			csvFormat:       "json",
```
shlomitk1, line 130: csvFormat is csv, not json. Did you mean dataFormat?

pkg/utils/utils.go
shlomitk1, line 85: I don't understand the check. The function returns a string array which is never nil, so you will always do hashing even if the name is ok.
`func IsDNS1123Subdomain(value string) []string {`

connectors/katalog/pkg/connector/connector.go
```diff
@@ -121,6 +130,7 @@ func (r *Handler) createAsset(c *gin.Context) {
 		assetPrefix = request.DestinationAssetID + "-"
```
shlomitk1, line 130: The validation should be done here (line 130) and not in the line 133

connectors/katalog/pkg/connector/connector.go
```diff
@@ -39,6 +41,13 @@ func NewHandler(client kclient.Client) *Handler {
 	return handler
 }
 
+func conformeName(name string) string {
```
shlomitk1, line 44: conformName?

connectors/katalog/pkg/connector/connector_test.go
```diff
@@ -277,7 +293,6 @@ func TestCreateAssetWthNoDestinationAssetID(t *testing.T) {
 
 		assetName := response.AssetID
 		namespace := "fybrik-system"
-		g.Expect(assetName).Should(HavePrefix(FybrikAssetPrefix))
```
shlomitk1, line 280: Either return this check or simplify the connector code not to use the prefix in the first place.

connectors/katalog/pkg/connector/connector.go
shlomitk1: Look at the following code:
	```
        assetPrefix := FybrikAssetPrefix
	if request.DestinationAssetID != "" {
		assetPrefix = request.DestinationAssetID + "-"
	}
	assetPrefix = r.conformeName(assetPrefix)
```
either request.DestinationAssetID exists or not, in both cases assetPrefix will end with "-" which fails the conform check. The check should be done on request.DestinationAssetID itself.

File: connectors/katalog/pkg/connector/connector_test.go (hunk 9 above)
Comment (tomersolomon1) [reply connectors/katalog/pkg/connector/connector_test.go:280]: I don't understand this test, let's discuss it.

File: connectors/katalog/pkg/connector/connector.go (hunk 7 above)
Comment (tomersolomon1) [reply connectors/katalog/pkg/connector/connector.go:134]: Ok I see, "-" is not allowed in the end. I don't understand then whey we added this "-" to begin with, let's discuss tomorrow.

File: connectors/katalog/pkg/connector/connector_test.go [hunk 10]
```diff
@@ -119,86 +121,119 @@ func TestCreateAsset(t *testing.T) {
...
-				},
-				{
-					Name: "nameOrig",
-					Tags: &taxonomy.Tags{Properties: serde.Properties{Items: map[string]interface{}{
-						"SPI": true,
-					}}},
-				},
-			},
+	var tests = []struct {
+		csvFormat       taxonomy.DataFormat
```
shlomitk1, line 125: rename to format

shlomitk1 (changes requested):
Discard go.mod and go.sum, resolve conflicts with master

shlomitk1 (approved):

pkg/utils/utils.go
```diff
@@ -94,7 +96,7 @@ func ShortenedName(name string, maxLength, hashLength int) string {
 // This method shortens the name keeping a prefix and using the last 5 characters of the
 // new name for the hash of the postfix.
 func K8sConformName(name string) string {
-	return ShortenedName(name, k8sMaxConformNameLength, hashPostfixLength)
+	return ShortenedName(name, K8sMaxConformNameLength, hashPostfixLength)
```
roytman, line 99: can you rename the method to K8sConformNameLenght

roytman (approved):

PR #1966, opened 2023-02-08
Propagate messages from the connectors

shlomitk1:
Closes https://github.com/fybrik/fybrik/issues/1858

This PR adds an optional Message field to GetAssetInfo and GetPolicyDecisions API responses, then propagates them back to the status of FybrikApplication.


manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -532,15 +540,16 @@ func (r *FybrikApplicationReconciler) constructDataInfo(req *datapath.DataInfo,
 
 		if response, err = r.DataCatalog.GetAssetInfo(&request, credentialPath); err != nil {
 			log.Error().Err(err).Msg("failed to receive the catalog connector response")
-			return err
+			return "", err
```
roytman, line 543: let me understand, the catalogMsg is relevant to the success response only ?

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -553,24 +562,27 @@ func (r *FybrikApplicationReconciler) constructDataInfo(req *datapath.DataInfo,
 	configEvaluatorInput.Request = CreateDataRequest(input, req.Context, &req.DataDetails.ResourceMetadata)
 
 	// Governance actions
-	err = r.checkGovernanceActions(configEvaluatorInput, req, appContext, env)
+	governanceMsg, err = r.checkGovernanceActions(configEvaluatorInput, req, appContext, env)
 	if err != nil {
-		return err
+		return "", errors.WithMessage(err, governanceMsg)
```
roytman, line 567: we include governaceMsg in the error response, but lose the catalogMsg. Is it correct?

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -553,24 +562,27 @@ func (r *FybrikApplicationReconciler) constructDataInfo(req *datapath.DataInfo,
...
 		appContext.Log.Error().Err(err).Msg("Error evaluating config policies")
-		return err
+		return "", err
 	}
 	logging.LogStructure("Config Policy Decisions", configDecisions, appContext.Log, zerolog.DebugLevel, false, false)
 	req.WorkloadCluster = configEvaluatorInput.Workload.Cluster
 	req.Configuration = configDecisions
-	return nil
+	// propagate messages from the catalog and policy manager
+	msg := strings.TrimPrefix(catalogMsg+Separator+governanceMsg, Separator)
```
roytman, line 578: Note:
What is the message's format? If it is a "closed" JSON, I mean 
```json
{
	"code": "FYBMG0001W",
	"level": "WARN / INFO / ERROR",
	"description": "some message description"
}
```
 we will get  something like:
 ```json
{
	"code": "FYBMG0001W",
	"level": "WARN / INFO / ERROR",
	"description": "some message description"
};
{
	"code": "FYBMG0001W",
	"level": "WARN / INFO / ERROR",
	"description": "some message description"
}
```
Which is not a valid JSON, on the other hand the message is defined as a String, so on the Fybrik side it will work.

File: manager/controllers/app/fybrikapplication_controller.go [hunk 4]
```diff
@@ -618,24 +630,24 @@ func (r *FybrikApplicationReconciler) checkGovernanceActions(configEvaluatorInpu
 			ProcessingLocation: geo,
 		}
 
-		actions, err := LookupPolicyDecisions(req.Context.DataSetID, resMetadata, r.PolicyManager, appContext, &reqAction)
+		actions, _, err := LookupPolicyDecisions(req.Context.DataSetID, resMetadata, r.PolicyManager, appContext, &reqAction)
```
roytman, line 633: why do you ignore message here and in the case of errors, here and below, drop previously return message ?

manager/controllers/app/policy_compiler_interface.go
```diff
@@ -83,13 +83,13 @@ func LookupPolicyDecisions(datasetID string, resourceMetadata *datacatalog.Resou
 	openapiResp, err := policyManager.GetPoliciesDecisions(openapiReq, creds)
 	var actions []taxonomy.Action
 	if err != nil {
-		return actions, err
+		return actions, "", err
```
roytman, line 86: in the case of error we don't return openapiResp.Message, but below at Line 108 we do return the message and a new error.
Later in FybricApplicationController, in the case of error, sometimes we add the message into the return error and sometimes do not.

Maybe the logic is correct but please add some comments with explanation.

pkg/model/datacatalog/api.go
```diff
@@ -21,6 +21,8 @@ type GetAssetResponse struct {
 	// Vault plugin path where the data credentials will be stored as kubernetes secrets
 	// This value is assumed to be known to the catalog connector.
 	Credentials string `json:"credentials"`
+	// Additional message to be reported to the user
```
roytman, line 24: Should we provide more info about the message format (the code, level, ....)

pkg/model/policymanager/api.go
```diff
@@ -14,6 +14,9 @@ type GetPolicyDecisionsRequest struct {
 }
 
 type GetPolicyDecisionsResponse struct {
-	DecisionID string       `json:"decision_id,omitempty"`
-	Result     []ResultItem `json:"result"`
+	DecisionID string `json:"decision_id,omitempty"`
+	// Additional message to be reported to the user
```
roytman, line 18: as in the datacatalog.

manager/controllers/app/fybrikapplication_controller.go
shlomitk1: Yes, all fields in the response are queried only if there is a successful response.

manager/controllers/app/fybrikapplication_controller.go
shlomitk1: The message from the policy manager might be relevant to understand the cause of the Deny action, for example.
Messages from catalog and policy manager are both included if there are no errors to construct the data plane. Otherwise, errors are not accumulated.

manager/controllers/app/fybrikapplication_controller.go
shlomitk1: For Fybrik, it is a generic string.

manager/controllers/app/fybrikapplication_controller.go
shlomitk1: These calls are done to evaluate the potential storage accounts that can be used in case of copy, what actions need to be taken should the copy be done. We refer only to the message for "real" requests not including copy.

manager/controllers/app/policy_compiler_interface.go
shlomitk1: If there is an error from the connector, the response may be nil, invalid, etc.
Added comments with explanation to all relevant functions.

pkg/model/datacatalog/api.go
shlomitk1: No, the message format is specific to the catalog in use.

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -511,12 +518,20 @@ func (r *FybrikApplicationReconciler) ValidateAssetResponse(response *datacatalo
 		datasetID, allErrs)
 }
 
+// constructDataInfo collects the following information about the asset:
+// - asset metadata
+// - governance actions to be performed on the data
+// - potential governance actions in case of caching the asset in a specific location
+// - decisions after evaluating config policies
+// The function returns an error received whilw communication with connectors or evaluating policies
```
roytman, line 526: whilw?
The function returns an error received (from, by?) communication with connectors or evaluating policies

manager/controllers/app/fybrikapplication_controller.go
roytman: Please add this explanation into the code, it can be relevant if somebody else would need to work with the code.

manager/controllers/app/fybrikapplication_controller.go
roytman: Again, can you add comments, pls

manager/controllers/app/policy_compiler_interface.go
```diff
@@ -64,9 +64,19 @@ func ValidatePolicyDecisionsResponse(response *policymanager.GetPolicyDecisionsR
...
+// - application info
+// - data flow and locations
+// Output:
+// - a list of governance actions (upon a successful response)
+// - a message from the connector (upon a successful response)
+// - an error from the connector or an error formulated by Fybrik in case of Deny
 func LookupPolicyDecisions(datasetID string, resourceMetadata *datacatalog.ResourceMetadata,
 	policyManager connectors.PolicyManager, appContext ApplicationContext,
-	op *policymanager.RequestAction) ([]taxonomy.Action, error) {
+	op *policymanager.RequestAction) ([]taxonomy.Action, string, error) {
```
roytman, line 79: L: 87 
managment -> management

shlomitk1: Fixed

manager/controllers/app/fybrikapplication_controller.go
shlomitk1: Fixed

manager/controllers/app/fybrikapplication_controller.go
shlomitk1: Done

manager/controllers/app/fybrikapplication_controller.go
shlomitk1: Added comments

roytman (approved):

Issue #1240, opened 2022-02-01
Storage allocation - support data stores other than COS which is currently supported


simanadler:

shlomitk1:
@simanadler datashim supports various S3 data stores, not only COS. "cos" is just a name they use for s3 inside a dataset object.

revit13:
the quick start uses localstack...

revit13:
what about allocation in other storage types like google-sheet etc? I think it was raised before to consider the option that an external component will handle the allocation.

PR #1973, opened 2023-02-15
Allocation of MySQL storage

### LINKED ISSUE #1928: Support allocation/de-allocation of MySQL storage

shlomitk1:
Fixes #1928, #1240
This PR allocates storage for MySQL based on MySQL storage account. It connects to the MySQL server and generates a new database. Upon deletion, the database is deleted.
To be done in the future PRs:
- consider "DeleteIfEmpty" flag passed in the request
- check that there are no other tables before database deletion
- extend https://github.com/fybrik/airbyte-module/blob/main/fybrik/README_mysql.md with the example of creating a new dataset
- allow user to specify the database and table names


roytman (commented):
@shlomitk1 , it is not this PR review, but please add "-" to https://github.com/fybrik/fybrik/blob/25bb9aeeaf228afcc66cc506fda2fda7a485ff5d/charts/fybrik/templates/manager-deployment.yaml#L71
{{**-** if .Values.storageManager.image }}

shlomitk1:
> {{**-** if .Values.storageManager.image }}

Done

pkg/storage/registrator/agent/agent.go
```diff
@@ -30,9 +31,9 @@ func GetProperty(props map[string]interface{}, t taxonomy.ConnectionType, key st
 	}
 	switch propertyMap := propertyMap.(type) {
 	case map[string]interface{}:
-		property := propertyMap[key]
-		if property != nil {
-			return property.(string), nil
+		property, found := propertyMap[key]
+		if found && (property != nil) {
```
roytman, line 35: why do you separate `nil` return and unfound, if you have the same processing ?
why not just `if property != nil ` ?

pkg/storage/impl/mysql/mysql_impl.go
```diff
@@ -21,7 +47,7 @@ type MySQLImpl struct {
 
 // implementation of AgentInterface for MySQL
 func NewMySQLImpl() *MySQLImpl {
-	return &MySQLImpl{Name: "mysql", Log: logging.LogInit(logging.CONNECTOR, "StorageManager")}
+	return &MySQLImpl{Name: mysqlAgent, Log: logging.LogInit(logging.CONNECTOR, "StorageManager")}
```
roytman, line 50: I'd suggest to add impl type into the LogInit, so later we will be able easily separate messages from different implementations.

pkg/storage/impl/mysql/mysql_impl.go
```diff
@@ -37,12 +63,103 @@ func (impl *MySQLImpl) GetConnectionType() taxonomy.ConnectionType {
 	return impl.Name
 }
 
-// storage allocation - placeholder
+func dsn(username, password, hostname, port, dbName string) string {
+	return fmt.Sprintf("%s:%s@tcp(%s:%s)/%s", username, password, hostname, port, dbName)
```
roytman, line 67: Printing password is problematic

pkg/storage/impl/mysql/mysql_impl.go
```diff
@@ -37,12 +63,103 @@ func (impl *MySQLImpl) GetConnectionType() taxonomy.ConnectionType {
...
+// storage allocation
+// host, port are taken from the storage account
+// database name is generated, table is taken from the DatasetProperties.Name
+// TODO: allow database and table specification inside FybrikApplication
 func (impl *MySQLImpl) AllocateStorage(request *storagemanager.AllocateStorageRequest, client kclient.Client) (taxonomy.Connection, error) {
-	return taxonomy.Connection{Name: impl.Name}, nil
+	var host, port string
+	var err error
+	if host, err = agent.GetProperty(request.AccountProperties.Items, impl.Name, hostKey); err != nil {
+		return taxonomy.Connection{}, err
```
roytman, line 96: we return the same error from different locations below without any logging. It will create an issue to identify which case threw the error.

pkg/storage/impl/mysql/mysql_impl.go
```diff
@@ -37,12 +63,103 @@ func (impl *MySQLImpl) GetConnectionType() taxonomy.ConnectionType {
...
+	if err != nil {
+		return taxonomy.Connection{}, err
+	}
+
+	// generate database name
+	suffix, _ := random.Hex(randomSuffixLength)
+	database := request.Opts.AppDetails.Name + "-" + request.Opts.AppDetails.Namespace + suffix
+
+	table := request.Opts.DatasetProperties.Name
+	// connect to the server and create the database if empty
```
roytman, line 111: here there is a contradiction, if database name is auto generated, it should not exist, if we want to reuse the data base, we should not generate the suffix.

pkg/storage/impl/mysql/mysql_impl.go
```diff
@@ -37,12 +63,103 @@ func (impl *MySQLImpl) GetConnectionType() taxonomy.ConnectionType {
...
+			},
+		},
+	}
+	return connection, nil
 }
 
-// storage deletion - placeholder
+// storage deletion
 func (impl *MySQLImpl) DeleteStorage(request *storagemanager.DeleteStorageRequest, client kclient.Client) error {
+	var host, port, database string
```
roytman, line 142: we should be caution with this, to prevent deletion of a user database.

pkg/storage/impl/mysql/mysql_impl.go
shlomitk1: Password is not printed. Here it's just formatting a string.

pkg/storage/impl/mysql/mysql_impl.go
shlomitk1: It's for the future when database reuse becomes supported.

pkg/storage/registrator/agent/agent.go
shlomitk1: changed

pkg/storage/impl/mysql/mysql_impl.go
shlomitk1: ok

pkg/storage/impl/mysql/mysql_impl.go
shlomitk1: added details

pkg/storage/impl/mysql/mysql_impl.go
shlomitk1: right, this is temporary as long as Fybrik auto-generates a database with a single table

pkg/storage/registrator/agent/agent.go
```diff
@@ -32,7 +33,7 @@ func GetProperty(props map[string]interface{}, t taxonomy.ConnectionType, key st
 	case map[string]interface{}:
 		property := propertyMap[key]
 		if property != nil {
-			return property.(string), nil
+			return fmt.Sprintf("%v", property), nil
```
roytman, line 36: if property == nil, you return `"", nil`, is it correct?
should it be `errors.New("undefined property " + key)` ?
or just 
```golang
func GetProperty(props map[string]interface{}, t taxonomy.ConnectionType, key string) (string, error) {
	propertyMap := props[t]
	if propertyMap == nil {
		return "", errors.New("undefined properties for " + string(t))
	}
	propMap, err := propertyMap.(map[string]interface{})
	if err != true {
		return "", errors.New("undefined property " + key)
	}
	property := propMap[key]
	if property != nil {
		return fmt.Sprintf("%v", property), nil
	}
	return "", errors.New("undefined property " + key)
}
```

pkg/storage/impl/mysql/mysql_impl.go
roytman: I meant, how can you separate errors from L97, L100, L115, and L123. `details` are the same, just according to the original error ?

pkg/storage/registrator/agent/agent.go
shlomitk1: modified

roytman (approved):

PR #1981, opened 2023-02-20
Validate create asset response.

revit13:
This PR adds a missing taxonomy verification of the create asset response.

### DIFF
--- a/manager/controllers/app/data_catalog_interface.go
+++ b/manager/controllers/app/data_catalog_interface.go
@@ -62,6 +62,12 @@ func (r *FybrikApplicationReconciler) RegisterAsset(assetID string, catalogID st
 		log.Error().Err(err).Msg("failed to receive the catalog connector response")
 		return "", err
 	}
+	err = r.ValidateAssetResponse(response, DataCatalogCreateAssetResponseTaxonomy, assetID)
+	if err != nil {
+		log.Error().Err(err).Msg("failed to validate the catalog connector response")
+		// return the error from the schema validator
+		return "", err
+	}
 
 	return response.AssetID, nil
 }
--- a/manager/controllers/app/fybrikapplication_controller.go
+++ b/manager/controllers/app/fybrikapplication_controller.go
@@ -70,7 +70,8 @@ type ApplicationContext struct {
 }
 
 var ApplicationTaxonomy = environment.GetDataDir() + "/taxonomy/fybrik_application.json"
-var DataCatalogTaxonomy = environment.GetDataDir() + "/taxonomy/datacatalog.json#/definitions/GetAssetResponse"
+var DataCatalogGetAssetResponseTaxonomy = environment.GetDataDir() + "/taxonomy/datacatalog.json#/definitions/GetAssetResponse"
+var DataCatalogCreateAssetResponseTaxonomy = environment.GetDataDir() + "/taxonomy/datacatalog.json#/definitions/CreateAssetResponse"
 
 const (
 	FybrikApplicationKind = "FybrikApplication"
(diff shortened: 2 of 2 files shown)


manager/controllers/app/data_catalog_interface.go
```diff
@@ -21,7 +21,7 @@ import (
 // - an error if happened
 // - the new asset identifier
 func (r *FybrikApplicationReconciler) RegisterAsset(assetID string, catalogID string,
-	info *fapp.DatasetDetails, input *fapp.FybrikApplication) (string, error) {
+	info *fapp.DatasetDetails, input *fapp.FybrikApplication) (*datacatalog.CreateAssetResponse, error) {
```
shlomitk1, line 24: Why are you changing the signature? The original behavior was to register the asset and return the registered asset id or an error. Why not do the validation here and return the same after validation.

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -242,8 +243,16 @@ func (r *FybrikApplicationReconciler) checkReadiness(applicationContext Applicat
...
 			// register the asset
-			if newAssetID, err := r.RegisterAsset(assetID, dataCtx.Requirements.FlowParams.Catalog,
+			if response, err := r.RegisterAsset(assetID, dataCtx.Requirements.FlowParams.Catalog,
 				&provisioned, applicationContext.Application); err == nil {
+				newAssetID := response.AssetID
+				err = r.ValidateCreateAssetResponse(response, DataCatalogCreateAssetResponseTaxonomy, assetID)
+				if err != nil {
+					applicationContext.Log.Error().Err(err).Msg("failed to validate the catalog connector " +
+						"response when creating new asset")
+					setReadyCondition(applicationContext, assetID)
```
shlomitk1, line 253: We could not properly register the asset. Why is it ready?

manager/controllers/app/fybrikapplication_controller.go
```diff
@@ -518,6 +522,26 @@ func (r *FybrikApplicationReconciler) ValidateAssetResponse(response *datacatalo
...
+	taxonomyFile, datasetID string) error {
+	// Convert GetAssetRequest Go struct to JSON
+	responseJSON, err := json.Marshal(response)
+	if err != nil {
+		return err
+	}
+	return r.validateAssetResponse(responseJSON, taxonomyFile, datasetID)
+}
+
+func (r *FybrikApplicationReconciler) ValidateCreateAssetResponse(response *datacatalog.CreateAssetResponse,
```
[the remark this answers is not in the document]
revit13: @shlomitk1 ok, should I use reflect to choose the right taxonomy file based on the response type? Thanks

manager/controllers/app/data_catalog_interface.go
revit13: I'll do that Thanks, I mistakenly thought the functions do not belong to the same class.

manager/controllers/app/fybrikapplication_controller.go
revit13: I'll fix that thanks

manager/controllers/app/fybrikapplication_controller.go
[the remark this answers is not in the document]
shlomitk1: No, the function will already get the right taxonomy, as you implemented. I suggest merely to change the function signature and then you'll be able to call the same function with different taxonomy file.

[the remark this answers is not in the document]
revit13: @shlomitk1 Thanks please take a look at the changes. (I overwrite the previous commits) Thanks

shlomitk1 (approved):

PR #2018, opened 2023-03-06
Remove option to overwrite base images.

### LINKED ISSUE #2013: Different base container mages used by Fybrik
Fybrik uses different base container images, e.g. ubi8/ubi:8.7 and ubi8/ubi-minimal:8.7.

 ubi8/ubi:8.7 doesn't include `microdnf` so the checks of PR #2008 fail. We have to check if we can install `microdnf` or use another image.

revit13:
/Closes #2013
Testing the option to remove option to overwrite base images.


roytman (commented):
A single comment, we have here 4 Docker files and in each one we define the image name, 
can it be a kind of share env variable ?

manager/Dockerfile
```diff
@@ -1,7 +1,6 @@
 # Copyright 2020 IBM Corp.
 # SPDX-License-Identifier: Apache-2.0
-ARG image=registry.access.redhat.com/ubi8/ubi-minimal:8.7
-FROM $image
+FROM registry.access.redhat.com/ubi8/ubi-minimal:8.7
```
roytman, line 3: In order to use the same image tag, we can use the following:
```
ARG tag=8.7
FROM registry.access.redhat.com/ubi8/ubi-minimal:$tag
```

manager/Makefile
```diff
@@ -27,13 +27,8 @@ source-build:
 # Overwrite docker-build from docker.mk
 .PHONY: docker-build
 docker-build: source-build
-ifdef FROM_IMAGE
-	docker build . -t ${IMG} -f Dockerfile --build-arg image=${FROM_IMAGE}
-	rm manager
-else
 	docker build . -t ${IMG} -f Dockerfile
```
roytman, line 30: docker build . -t ${IMG} -f Dockerfile --build-arg tag=${BASE_IMAGE_TAG}

and to add `BASE_IMAGE_TAG=8.7` into `requirements.env`

roytman (approved):

Issue #2000, opened 2023-02-28
Updating rego policies in TLS settings


tomersolomon1:
It seems there might be an issue with deploying new rego policies via a configmap in the TLS settings.
For instance, in the settings of `https://github.com/fybrik/fybrik/blob/master/charts/fybrik/notebook-test-readflow.tls.values.yaml` we tried to remove the bootstrap policies and create the policies via a configmap.

When trying to deploy the following yaml we get an error:
```
apiVersion: v1
data:
  policy-read.rego: |+
    package dataapi.authz

    rule[{}] {
        description := "allow read datasets with no tags"
        input.action.actionType == "read"
        not input.resource.metadata.tags
    }

    rule[{"action": {"name":"RedactAction", "columns": column_names}, "policy": description}] {
        description := "While reading redact columns tagged as PII in datasets tagged with finance = true"
        input.action.actionType == "read"
        input.resource.metadata.tags.finance
        column_names := [input.resource.metadata.columns[i].name | input.resource.metadata.columns[i].tags.PII]
    }

kind: ConfigMap
metadata:
  labels:
    openpolicyagent.org/policy: rego
  name: policy-read
  namespace: fybrik-system
```

The error:
```
ind: ConfigMap
metadata:
  annotations:
    [openpolicyagent.org/policy-status](http://openpolicyagent.org/policy-status): '{"status":"error","error":{"Op":"Put","URL":"http://localhost:8181/v1/policies/fybrik-system/policy-read/policy-read.rego","Err":{"Op":"dial","Net":"tcp","Source":null,"Addr":{"IP":"127.0.0.1","Port":8181,"Zone":""},"Err":{"Syscall":"connect","Err":111}}}}'
```

PR #2027, opened 2023-03-13
Change OPA server to use TLS instead of mutual TLS in notebook test

revit13:
Relates to #2000 

OPA's [kube-mgmt](https://github.com/open-policy-agent/kube-mgmt) sidecar is responsible for automatically discover policies and JSON data stored in ConfigMaps in Kubernetes and loads them into OPA.
When configuring OPA with mutual tls then the client (kube-mgmt) certificate should be provided as well. to avoid providing such a certificate this PR changes the OPA server to use TLS and not mutual TLS.


manager/testdata/notebook/README.md
```diff
@@ -10,7 +10,7 @@ are configured to use TLS:
 - Fybrik manager
 - Vault
 
-All servers above except Openmetadata server use mutual TLS.
+All servers above except Openmetadata and OPA server use mutual TLS. OPA server uses TLS. Openmetadata server not using TLS.
```
roytman, line 13: Openmetadata server does not use TLS. 
or
Meantime, Openmetadata server does not support TLS.  (is it correct?)

revit13: Fixed thanks. Last time I checked Openmetadata had some open issues related to their TLS support.

roytman (approved):

PR #2034, opened 2023-03-18
fixed the arch pptx file

aradhalevy:
fixed the arch pptx file, as it was of an older version of the arch png


roytman (commented):
Thanks @aradhalevy,

Could you please move the file to fybrik/docs. 
The file is not part of the Website, so probably will be better to store it there.

roytman (commented):
architcture -> architecture

roytman (approved):

PR #2017, opened 2023-03-06
A script to collect fybrik information using kubectl

Vicky-Goldin:
For the record: started to work on the bash script on 28/2.

Wrote a `bash` script, which saves in a local directory the logs, output of `get pods`, modules yamls, certain configmaps, and fybrikapplication yaml if a fybrikapplication uuid (or namespace and name) were specified.
The behavior is configured by flags, which are explained in `./hack/save-logs-files.sh -h`

An example for the output of running the script from root of `fybrik`:
`./hack/save-logs-files.sh -u 1d8c95ff-0eda-45b9-a38d-291b68170ad7 -d 10m`
[fybrik_logs_2023-03-16_15-25_last_10m.zip](https://github.com/fybrik/fybrik/files/10991286/fybrik_logs_2023-03-16_15-25_last_10m.zip)

Closes #568

PR #2039, opened 2023-03-22
A script to collect fybrik information using kubectl [new branch]

### LINKED ISSUE #568: A method to collect debugging information from a running system
A method to collect debugging information from a running system

reported by:  @ronenkat 

This can be implemented as a script which will dump all relevant data to help collect information on the system.

The relevant data could contain:
- last x lines from logs of the manager and connectors,
- status of m4dapplications that are running
- dump of the relevant config maps in m4d-system

Vicky-Goldin:
This PR continues PR #2017, but from a new branch, since the previous one didn't use `git` commands correctly when merging new commits from `master`.

Closes #568

### DIFF
--- a/hack/save-logs-files.sh
+++ b/hack/save-logs-files.sh
@@ -3,6 +3,11 @@
 # Copyright 2023 IBM Corp.
 # SPDX-License-Identifier: Apache-2.0
 
+if [[ ! "$OSTYPE" == "linux"* ]]; then
+    echo "OS '$OSTYPE' is not supported. Aborting." >&2
+    exit 1
+fi
+
 # error and exit configuration
 set -eu
 
--- a/hack/save-logs-files.sh
+++ b/hack/save-logs-files.sh
@@ -6,26 +6,18 @@
 # error and exit configuration
 set -eu
 
-FLAG_DURATION="-d|--duration"
-FLAG_NAMESPACES="-n|--namespaces"
-FLAG_PATH="-p|--path"
-FLAG_APP="-a|--application"
-FLAG_UUID="-u|--uuid"
-LONG_POS=3
+FLAG_DURATION="d"
+FLAG_NAMESPACES="n"
+FLAG_PATH="p"
+FLAG_APP="a"
+FLAG_UUID="u"
+FLAG_HELP="h"
 ARGS_SEPARATOR=","
-APP_CONFLICT_MESSAGE="Only one of --application / --uuid may be used"
+APP_CONFLICT_MESSAGE="Only one of -$FLAG_APP/-$FLAG_UUID may be used"
 
-# define the possible flag options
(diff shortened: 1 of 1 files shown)


shlomitk1 (approved):

hack/save-logs-files.sh
```diff
@@ -0,0 +1,236 @@
...
+    # if given fybrik application uuid, save its yaml and create another file which only contains logs with the uuid
+    if [[ ! -z fybrik_application_uuid ]] && [[ is_found_fybrik_application_yaml -eq 0 ]]; then
+        for fybrik_application in $(kubectl get fybrikapplications -n $ns -o=jsonpath='{.items[*].metadata.name}'); do
+            if [[ $(kubectl get fybrikapplication $fybrik_application -n $ns -o=jsonpath='{.metadata.uid}') == $fybrik_application_uuid ]]; then
+                kubectl get fybrikapplication $fybrik_application -n $ns -o=yaml &> $dirname_logs/fybrik_application_$ns--$fybrik_application.yaml
+                is_found_fybrik_application_yaml=1
+            fi
+        done
+    fi
+done
```
roytman, line 186: when I ran the script without any arguments I got the following message:
`./save-logs-files.sh: line 186: namespaces_list[@]: unbound variable`

hack/save-logs-files.sh
```diff
@@ -0,0 +1,236 @@
...
+            # check that the argument is valid
+            re='^[0-9]+[smh]$'
+            if [[ ! $duration =~ $re ]]; then
+                echo "error: invalid argument '$duration' for $FLAG_DURATION flag: should contain a number and a time unit (s/m/h), e.g. 1h or 30m"
+                print_usage
+                exit 1
+            fi
+            ;;
+        $FLAG_NAMESPACES)
+            readarray -d $ARGS_SEPARATOR -t namespaces_list < <(printf '%s' "$OPTARG")
```
roytman, line 58: ./save-logs-files.sh: line 58: readarray: command not found

roytman: Here are several possible solutions: https://stackoverflow.com/questions/23842261/alternative-to-readarray-because-it-does-not-work-on-mac-os-x

roytman:
@Vicky-Goldin, if you see that support MacOS is too complicated, please just add in the start check OS and in the case of MacOS print a message that it is not supported. 
I don't want to spend too much time for the script, it is just a tool for your main task. 

Something like:

```bash
if [[ "$OSTYPE" != "linux*" ]]; then
  echo "OS '$OSTYPE' is not supported. Aborting." >&2
  exit 1
fi
```

hack/save-logs-files.sh
Vicky-Goldin: This doesn't happen on linux, this is probably because of an old version of `bash` in macOS: https://stackoverflow.com/questions/7577052/bash-empty-array-expansion-with-set-u
For now it was decided to not support mac.

roytman (approved):

PR #2046, opened 2023-03-26
Print debugging information of the deployed module in the notebook readflow test.

revit13:
Relates to https://github.com/fybrik/fybrik/issues/1860

This PR prints information that might be useful for debugging the deployed modules before fybrikapplication is deleted at the end of the test:

- print the logs of the deployed module pod
- print the config-maps in the module namespace


shlomitk1:
Can this information be printed only in case of errors?

revit13:
@shlomitk1 I do not know will have to check... I hope to catch failures like in https://github.com/fybrik/fybrik/actions/runs/4483018011/jobs/7881785854

shlomitk1:
> @shlomitk1 I do not know will have to check... I hope to catch failures like in https://github.com/fybrik/fybrik/actions/runs/4483018011/jobs/7881785854

Create a boolean variable "success" set to false which will be set to true at the end of the test function (after all checks). In the function that prints logs + cms, if success is true, exit the function without printing anything.

revit13:
@shlomitk1 fixed thanks!

shlomitk1 (commented):
I am not sure that the functionality of getting pod logs belongs to pkg/utils and not to the test file specifically.

revit13:
Abandon this approach for debugging the test for now...

PR #2053, opened 2023-03-29
Blueprint changes for isolation

### LINKED ISSUE #2037: Extend Blueprint CRD with module inter-connection info
Add to the blueprint: 
- A list of connections to connect to by the module.
- A list of connections to the module.

### LINKED ISSUE #2048: Clear FybrikApplication selector for chained modules.
In chain deployments, the FybrikApplication selector is relevant to the module that is the data plane entry point. The selector is used for NetworkPolicies creations. 
In the current implementation, a workload is able to skip the starting data plane module and connect to the next one. 
For example, in our [FybrikModule chaining sample](https://fybrik.io/v1.3/samples/chaining-sample/), if the entry point of read module (airbite) is known, a workload can skip the transformation (PII data reduction) module (arrow-flight). 

The selector is relevant to only modules which can be accessed by user workloads.

shlomitk1:
Closes #2037
Closes #2048


manager/apis/app/v1beta1/blueprint_types.go
```diff
@@ -47,6 +47,29 @@ type ModuleArguments struct {
...
 }
 
+// ModuleDeployment specifies the deployment details of the module
+type ModuleDeployment struct {
+	// Cluster name
+	Cluster string `json:"cluster"`
+	// Release name
+	Release string `json:"release"`
+	// URLs the module needs access to
+	URLs []string `json:"urls"`
```
roytman, line 57: what will be format of the URLs? Is it net.URL,  just hostname, or any format defined in the taxonomy?

shlomitk1: For service it will probably be release.namespace:port. For datasets, either domain or host:port. Depends on whatever details we get from the catalog...

roytman (approved):

PR #2060, opened 2023-04-01
Replace reflect.DeepEqual by equality.Semantic.DeepEqual

shlomitk1:
Sometimes reflect.DeepEqual returned false when comparing two equal structures (blueprint specs). As a result, a ready status was not propagated into the plotter status, and application was not becoming ready.
Fixes https://github.com/fybrik/fybrik/issues/1860


roytman (commented):
This change is OK, but there are 2 others `reflect.DeepEqual` calls in `pkg/optimizer/flatzinc_model_test.go`.
Should we change them too? They are very test specific, and I don't expect empty maps, or arrays, but for symmetry.

shlomitk1:
> This change is OK, but there are 2 others `reflect.DeepEqual` calls in `pkg/optimizer/flatzinc_model_test.go`. Should we change them too? They are very test specific, and I don't expect empty maps, or arrays, but for symmetry.

No need. For these objects, it is better to compare values as strictly as possible.

roytman (approved):

PR #2061, opened 2023-04-02
Update setup clusters scripts.

revit13:
This PR updates hack/setup-local-single-cluster.sh and hack/setup-local-multi-cluster.sh as follows:

- it updates the ingress resource
- it changes the catalog used to katalog
- Removes NodePort type for Vault (we use ingress to Vault instead)
- Updates KUBE_VERSION


Mohammad-nassar10 (approved):
There is an error in `golint` action, you need to increase the timeout value.

Mohammad-nassar10 (dismissed):

hack/setup-local-multi-cluster.sh
```diff
@@ -2,27 +2,36 @@
...
+export DEPLOY_OPENMETADATA_SERVER=0
 
 make kind-setup-multi
 kubectl config use-context kind-control
 make -C third_party/razee all
 
+# setup control cluster
 kubectl config use-context kind-control
 make cluster-prepare
-make docker-minimal-it
```
roytman, line 16: if we remove docker-minimal-it from all scripts, why do we leave it in the Makefile?

charts/fybrik/kind-control.values.yaml
```diff
@@ -18,18 +18,21 @@ cluster:
   # Set to the name of the cluster.
   name: kind-control
   # Set to cluster zone.
-  zone: localhost
```
roytman, line 21: in general, why do we need all these yaml files used by  Razee samples and all notebook tests files in the fybrik chart?

roytman (commented):
I have several questions, let's talk

charts/fybrik/kind-control.values.yaml
revit13: @roytman as you suggested, the tests related values files are now under test/charts folder

Makefile
```diff
@@ -243,7 +243,7 @@ ifeq ($(DEPLOY_TLS_TEST_CERTS),1)
 	cd manager/testdata/notebook/read-flow-tls && ./setup-certs.sh
 endif
 ifeq ($(DEPLOY_OPENMETADATA_SERVER),1)
-	$(MAKE) -C third_party/openmetadata all
+#	$(MAKE) -C third_party/openmetadata all
```
roytman, line 246: why did you comment it out ?

revit13: Fixed. thanks for the catch!

roytman (approved):

PR #2069, opened 2023-04-11
Improve logs collector script 

### LINKED ISSUE #2049: Adapt logs collector script to macOS
[The script](https://github.com/fybrik/fybrik/blob/master/hack/save-logs-files.sh) currently aborts when it is not run on `linux`, because of different behavior of `bash` commands in `macOS`.
The problematic commands should be replaced so the script will be able to run on `macOS` as well.

Additionally, the script can be improved:
1. Saving the deployed modules in a dedicated subdirectory instead of the root of the output.
2. Filter the logs_combined file by log level with only warnings and errors, and save those logs into a separate additional file.
3. The script stops unexpectedly if it tries to load logs from a not accessible pod (e.g. due to `ImagePullBackOff`), instead it should ignore the error and continue to run, or at least complete `get pods` first.

Vicky-Goldin:
Made adaptations so the script will run on macOS (by changing several bash commands to make the script compatible across different shell versions). Tested the script on the mac with Apple chip.
Also made several improvements: 
* Check if pod is ready or not, and save 'describe' output if not.
* Create another logs file filtered by only log level of warn or error.
* Save all deployed modules in a dedicated subdirectory.
* Improve code readability with variables for files paths.
* Match terms spelling to the kubernetes way (the CRDs).

Closes #2049

### DIFF
--- a/hack/save-logs-files.sh
+++ b/hack/save-logs-files.sh
@@ -128,12 +128,16 @@ if [[ ${#namespaces_list[@]} -eq 0 ]]; then
     done
 fi
 
-# create a directory for saving output
+# create a directory and subdirectories for saving output
 dirname_logs=$output_path"fybrik_logs_$(date +%Y-%m-%d_%H-%M)_last_$duration"
 echo "Output will be saved in $dirname_logs"
 mkdir $dirname_logs
 dirname_logs_by_container=$dirname_logs/logs_by_container
 mkdir $dirname_logs_by_container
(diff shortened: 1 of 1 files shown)


roytman:
It works on my Mac, thanks. 
A couple of questions/comments:
1. What is the `s` character at the end of line 174?
2. I suggest changing the line 175 as :
```bash
  prefix=$ns/$cm
```
it is not crucial, but usually K8s objects are printed as nameSpace/objectName

roytman (approved):

Issue #2070, opened 2023-04-11
`errors.Wrap` calls shadow the source of the error


roytman:
When a Fybrik controller faces an error, sometimes its error is wrapped by the `error.Wrap` call and returned to the upper call.
As result, we can see the error in the fybrik application `status.errorMessage`, and we don't see the real point in the code where the error was accused. 
Before returning the wrapped error, we need to print an error message into the log.

roytman:
[manager.log](https://github.com/fybrik/fybrik/files/11310721/manager.log)

PR #2091, opened 2023-05-04
Log the caller of `setErrorCondition`

Vicky-Goldin:
Relates to #2070

Modify the log entry of `setErrorCondition` to show the caller of the function, to better capture where the error occurred.
Additionally fix some log calls so they will be executed properly.


roytman (commented):
good catch with `log.Error().Err(err).Send()`, but there are several more cases of `log.Error().Err(err)` without `Msg` or `Send`. 
Could you please fix them too
e.g. 
connectors/opa/connector.go:40
pkg/storage/impl/s3/s3_impl.go:49
pkg/storage/impl/mysql/mysql_impl.go:57
pkg/multicluster/razee/razee_cluster_manager.go:289
manager/controllers/app/npIsolation.go:126
manager/controllers/app/npIsolation.go:134

roytman (approved):

